PR #11, opened 2020-05-12
Extract the source type into a separate package.

immutableT:
Extract the source type into a separate package.

### DIFF
--- a/internal/levee.go
+++ b/internal/levee.go
@@ -26,10 +26,8 @@ import (
 	"golang.org/x/tools/go/analysis/passes/buildssa"
 	"golang.org/x/tools/go/ssa"
 
-	"github.com/eapache/queue"
-
 	"github.com/google/go-flow-levee/internal/pkg/config/regexp"
-	"github.com/google/go-flow-levee/internal/pkg/sanitizer"
+	"github.com/google/go-flow-levee/internal/pkg/source"
 	"github.com/google/go-flow-levee/internal/pkg/utils"
 )
 
@@ -88,7 +86,7 @@ func (c config) isSink(call *ssa.Call) bool {
 	return false
 }
 
-func (c config) isSanitizer(call *ssa.Call) bool {
+func (c config) IsSanitizer(call *ssa.Call) bool {
 	for _, p := range c.Sanitizers {
 		if p.matchMethodName(call) {
 			return true
@@ -112,7 +110,7 @@ func (c config) isSource(t types.Type) bool {
 	return false
 }
 
-func (c config) isSourceFieldAddr(fa *ssa.FieldAddr) bool {
+func (c config) IsSourceFieldAddr(fa *ssa.FieldAddr) bool {
 	// fa.Type() refers to the accessed field's type.
 	// fa.X.Type() refers to the surrounding struct's type.
 	deref := utils.Dereference(fa.X.Type())
@@ -127,7 +125,7 @@ func (c config) isSourceFieldAddr(fa *ssa.FieldAddr) bool {
 	return false
 }
 
-func (c config) isPropagator(call *ssa.Call) bool {
+func (c config) IsPropagator(call *ssa.Call) bool {
 	return c.isFieldPropagator(call) || c.isTransformingPropagator(call)
 }
 
@@ -306,81 +304,6 @@ var Analyzer = &analysis.Analyzer{
 	Requires: []*analysis.Analyzer{buildssa.Analyzer},
 }
 
-// source represents a source in an SSA call tree.
-// It is based on ssa.Node, with the added functionality of computing the recursive graph of
-// its referrers.
-// source.sanitized notes sanitizer calls that sanitize this source
-type source struct {
-	node       ssa.Node
-	marked     map[ssa.Node]bool
-	sanitizers []*sanitizer.Sanitizer
-}
-
-func newSource(in ssa.Node, config *config) *source {
-	a := &source{
-		node:   in,
-		marked: make(map[ssa.Node]bool),
-	}
-	a.bfs(config)
-	return a
-}
-
-// bfs performs Breadth-First-Search on the def-use graph of the input source.
-// While traversing the graph we also look for potential sanitizers of this source.
-// If the source passes through a sanitizer, bfs does not continue through that Node.
-func (a *source) bfs(config *config) {
-	q := queue.New()
-	q.Add(a.node)
-	a.marked[a.node] = true
-
-	for q.Length() > 0 {
-		e := q.Remove().(ssa.Node)
-
-		if c, ok := e.(*ssa.Call); ok && config.isSanitizer(c) {
-			a.sanitizers = append(a.sanitizers, &sanitizer.Sanitizer{Call: c})
-			continue
-		}
-
-		if e.Referrers() == nil {
-			continue
-		}
-
-		for _, r := range *e.Referrers() {
-			if _, ok := a.marked[r.(ssa.Node)]; ok {
-				continue
-			}
-			a.marked[r.(ssa.Node)] = true
-
-			// Need to stay within the scope of the function under analysis.
-			if call, ok := r.(*ssa.Call); ok && !config.isPropagator(call) {
-				continue
-			}
-
-			// Do not follow innocuous field access (e.g. Cluster.Zone)
-			if addr, ok := r.(*ssa.FieldAddr); ok && !config.isSourceFieldAddr(addr) {
-				continue
-			}
-
-			q.Add(r)
-		}
-	}
-}
-
-// hasPathTo returns true when a Node is part of declaration-use graph.
-func (a *source) hasPathTo(n ssa.Node) bool {
-	return a.marked[n]
-}
-
-func (a *source) isSanitizedAt(call ssa.Instruction) bool {
-	for _, s := range a.sanitizers {
-		if s.Dominates(call) {
-			return true
-		}
-	}
-
-	return false
-}
-
 // varargs represents a variable length argument.
 // Concretely, it abstract over the fact that the varargs internally are represented by an ssa.Slice
 // which contains the underlying values of for vararg members.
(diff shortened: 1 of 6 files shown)


internal/pkg/source/source_test.go
```diff
@@ -0,0 +1,166 @@
...
+	Doc:        "test harness for the logic related to sources",
+	Requires:   []*analysis.Analyzer{buildssa.Analyzer},
+	ResultType: reflect.TypeOf(analyzerResult{}),
+}
+
+func run(pass *analysis.Pass) (interface{}, error) {
+	in := pass.ResultOf[buildssa.Analyzer].(*buildssa.SSA)
+	var result analyzerResult
+	for _, fn := range in.SrcFuncs {
+		fmt.Printf("Func: %v\n", fn.Name())
```
PurelyApplied, line 69: I would prefer these to be `t.Logf` statements so they don't clutter up the command-line `go test ./...`  Although to do that, you'd have to do some closure tricks and reset `testAnalyzer.Run` at the start of every test.

If you want to keep the prints, I'd still prefer the closure.  Otherwise, if it was just for debugging, please remove.

internal/pkg/source/source_test.go
```diff
@@ -0,0 +1,166 @@
...
+		for _, b := range fn.Blocks {
+			for _, i := range b.Instrs {
+				fmt.Printf("%v : %T\n", i, i)
```
PurelyApplied, line 72: See about regarding `t.Logf`.

internal/pkg/source/source_test.go
```diff
@@ -0,0 +1,166 @@
...
+		fieldsPattern:      "name",
+	}
+
+	testCases := []struct {
+		desc                         string
+		pattern                      string
+		config                       *testConfig
+		wantCallConnectionIdx        []int
+		wantFieldAccessConnectionIdx []int
+		wantStoreConnectionIdx       []int
```
PurelyApplied, line 105: Testing based on index in the `ssa` result feels to me a bit brittle and resistant to future change.  See above comment about `// want` approach.

internal/pkg/source/source_test.go
```diff
@@ -0,0 +1,166 @@
...
+		},
+		{
+			desc:                         "sanitization",
+			pattern:                      "sanitization",
+			config:                       config,
+			wantFieldAccessConnectionIdx: []int{0},
+		},
+	}
+
+	for _, tt := range testCases {
```
PurelyApplied, line 128: This took me a while to understand, and I think every check done here would be performed by a `// want "allocation"`.

internal/pkg/source/testdata/src/sanitization/test.go
```diff
@@ -0,0 +1,32 @@
...
+
+import "fmt"
+
+type foo struct {
+	name     string
+	password string
+}
+
+func f1() {
+	foo := &foo{name: "bar", password: "P@ssword01"}
```
PurelyApplied, line 25: Might be best to avoid name-showing unless we're explicitly testing for it.

internal/pkg/utils/utils_test.go
```diff
@@ -27,7 +27,7 @@ import (
 )
 
 var testAnalyzer = &analysis.Analyzer{
-	Name:       "domination",
+	Name:       "dereference",
```
PurelyApplied, line 30: This rename is outside the scope of the commit message.  Please either update the commit message or separate this to another PR.

internal/levee.go
```diff
@@ -544,11 +467,11 @@ func readConfig(path string) (*config, error) {
 	return readConfigCached, readConfigCachedErr
 }
 
-func identifySources(conf *config, ssaInput *buildssa.SSA) map[*ssa.Function][]*source {
-	sourceMap := make(map[*ssa.Function][]*source)
+func identifySources(conf *config, ssaInput *buildssa.SSA) map[*ssa.Function][]*source.Source {
```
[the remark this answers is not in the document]
immutableT: I think so.
In this PR I wanted to focus on the logic of the Source type.

internal/pkg/source/source.go
```diff
@@ -0,0 +1,138 @@
...
+// Source.sanitized notes sanitizer calls that sanitize this Source
+type Source struct {
+	node       ssa.Node
+	marked     map[ssa.Node]bool
+	sanitizers []*sanitizer.Sanitizer
+	config     classifier
+}
+
+// Node returns the underlying ssa.Node of the Source.
+func (a *Source) Node() ssa.Node {
```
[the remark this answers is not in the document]
immutableT: I agree. 
I will fix this on a separate PR.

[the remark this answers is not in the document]
immutableT: In general, I agree.
However, since a Source could be referenced by Values and Instructions, it forces us to use ssa.Node to keep track of vertices, right?

internal/pkg/source/source_test.go
```diff
@@ -0,0 +1,166 @@
...
+	var result analyzerResult
+	for _, fn := range in.SrcFuncs {
+		fmt.Printf("Func: %v\n", fn.Name())
+		for _, b := range fn.Blocks {
+			for _, i := range b.Instrs {
+				fmt.Printf("%v : %T\n", i, i)
+
+				switch v := i.(type) {
+				case *ssa.Alloc:
+					result.allocations = append(result.allocations, v)
```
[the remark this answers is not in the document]
immutableT: I would certainly prefer the approach of the `analysistest`.
However, this would imply creating a new analyzer for each test case, right?
The statement above, assumes that we want to test more than just identifying the source, but also the relationships between the source and its referrers.

internal/pkg/utils/utils_test.go
immutableT: Reverted.

internal/pkg/source/source_test.go
immutableT: Removed print statements.

internal/pkg/source/source_test.go
immutableT: See my previous [response](https://github.com/google/go-flow-levee/pull/11#discussion_r424108402).

internal/pkg/source/source_test.go
immutableT: These tests go beyond just identifying where allocation is performed for a Source.
Concretely, these tests ensure that we:
1. do not build the graph of a Source beyond a sanitizer
2. do include calls to propagators in the graph

internal/pkg/source/testdata/src/sanitization/test.go
immutableT: Replaced with simpler values.

internal/pkg/source/source_test.go
```diff
@@ -0,0 +1,166 @@
...
+	dir := analysistest.TestData()
+	config := &testConfig{
+		propagatorsPattern: "propagator",
+		sanitizerPattern:   "sanitizer",
+		fieldsPattern:      "name",
+	}
+
+	testCases := []struct {
+		desc                         string
+		pattern                      string
```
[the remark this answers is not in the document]
immutableT: Removed the `desc` field.

internal/levee.go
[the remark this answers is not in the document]
PurelyApplied: Fair enough.

internal/pkg/source/source.go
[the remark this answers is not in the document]
PurelyApplied: I'm not sure that's the case.  Generally, a source will be sanitize/propagated/sunk by an `ssa.Instruction`

The following are the only `ssa` types which implement `Value` but not `Instruction`.

```
*Builtin
*Const
*FreeVar
*Function
*Global
*Parameter
```

So long as those types are part of source identification, I don't think we'll lose coverage.  But it would require another pass at the `bfs` / dominance logic, so probably shouldn't be part of this PR.

internal/pkg/source/source_test.go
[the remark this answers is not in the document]
PurelyApplied: > However, this would imply creating a new analyzer for each test case, right?
Ahh, that's a fair point.  Depending on how you split things, it might not be every test-case, but could have to be one analyzer for one or more `testdata/` files.

PurelyApplied:
As an aside, I would appreciate it if you didn't force push on your branch after the PR is open.  It makes it a harder to track changes across comments.  Since I've pulled your branch to look at the diff locally, I now have to blow the branch away for future examination rather than just pull new changes.  We can squash before merge if necessary.

internal/pkg/source/source.go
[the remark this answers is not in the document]
immutableT: I think we are on the same page.
I agree that the constructor of the source should accept `ssa.Value` and not `ssa.Node`. 

However, the graph of the source will contain Instructions (example `ssa.Store`), and since the `Source` itself is part of the graph, vertices would need to support a broader interface - probably Node.
Let's pick this up on a separate PR.

internal/pkg/source/source_test.go
immutableT: Removed.

internal/pkg/source/source_test.go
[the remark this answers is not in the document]
immutableT: The fundamental problem is that ssa instructions do not translate one-to-one to the original Go source code. 
Concretely, if I wanted to use the `analysistest` approach and assert that a given source has the expected graph, I would not be able to do so via comments alone since not all ssa instructions have an equivalent in the original source.
Therefore, a different approach is required. However, I hope that we could improve on the approach suggested by the PR by developing some helper types that reduce the cognitive overhead.

internal/pkg/source/testdata/src/allocation/test.go
```diff
@@ -0,0 +1,39 @@
...
+import (
+	"fmt"
+)
+
+type foo struct {
+	name string
+}
+
+func f1() {
+	foo := &foo{name: "bar"}
```
PurelyApplied, line 26: Also name shadowed.

internal/pkg/source/testdata/src/propagation/test.go
```diff
@@ -0,0 +1,31 @@
...
+package propagation
+
+import "fmt"
+
+type foo struct {
+	name string
+}
+
+func f1() {
+	foo := &foo{name: "bar"}
```
PurelyApplied, line 24: Also name shadowed.

internal/pkg/source/testdata/src/sanitization/test.go
```diff
@@ -0,0 +1,32 @@
...
+
+import "fmt"
+
+type foo struct {
+	name     string
+	password string
+}
+
+func f1() {
+	foo := &foo{name: "n", password: "p"}
```
PurelyApplied, line 25: Also name shadowed.

internal/pkg/source/source.go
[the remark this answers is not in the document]
PurelyApplied: > Let's pick this up on a separate PR.

+1.

PR #13, opened 2020-05-14
Extract config struct to config package.

PurelyApplied:
* Export ReadConfig
* Export Config, Config.IsSink, Config.IsSource
* Extract Config.sendsToIOWriter -> levee.sendsToIOWriter as a special case outside the scope of `Config`.
* Export NameMatcher, NameMatcher.MatchMethodName.

-----

This PR is partly in service of extracting source identification into a preliminary `Analyzer`.

### DIFF
--- a/internal/pkg/config/config.go
+++ b/internal/pkg/config/config.go
@@ -2,6 +2,7 @@ package config
 
 import (
 	"encoding/json"
+	"flag"
 	"fmt"
 	"go/types"
 	"io/ioutil"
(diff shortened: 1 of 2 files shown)


internal/pkg/config/config.go
```diff
@@ -0,0 +1,316 @@
...
+	"io/ioutil"
+	"sync"
+
+	"github.com/google/go-flow-levee/internal/pkg/config/regexp"
+	"github.com/google/go-flow-levee/internal/pkg/utils"
+	"golang.org/x/tools/go/ssa"
+)
+
+// FlagSet should be used by analyzers to reuse -config flag.
+var FlagSet flag.FlagSet
```
immutableT, line 17: WDYT about defining the flags in the main function?

PurelyApplied: I think it's preferred to use the `analysis.Analyzer.FlagSet` field.  From the `analysis` docs:

```
type Analyzer struct {
    // [...]
    // Flags defines any flags accepted by the analyzer.
    // The manner in which these flags are exposed to the user
    // depends on the driver which runs the analyzer.
    Flags flag.FlagSet
    // [...]
}
```

I'll admit I'm not familiar with `go vet`'s internals, but I think it needs to be here for `go vet -vettool=/path/to/levee` to behave as expected.

This was also partially motivated by sharing it with #17 extracting source identification analysis.  I suppose that it could have belonged to that PR rather than this one.

PR #15, opened 2020-05-14
Use goimports over gofmt in Travis-CI.

PurelyApplied:
`goimports` performs import formatting in addition to formatting performed by gofmt.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,11 +6,10 @@ go:
 script:
 - go test -v ./...
 - >
-  echo "Checking gofmt..." && (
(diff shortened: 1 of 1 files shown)


PurelyApplied:
<grumbles>

I had thought Travis would run on the first commit.

I'm going to force-push to remove the top commit so that the failure-case will be demonstrated.  Then I'll add the top commit back.

PurelyApplied:
Travis hated that, too.  The passing SHA does not match the one it should.

Closing and reopening one more time.

PR #17, opened 2020-05-14
Extract source identification as a separate analyzer

PurelyApplied:
This requires #13.  The first two commits belong to that PR.

### DIFF
--- a/internal/pkg/source/analyzer.go
+++ b/internal/pkg/source/analyzer.go
@@ -1,0 +1,59 @@
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package source
+
+import (
+	"reflect"
+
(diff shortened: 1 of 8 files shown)


internal/pkg/source/analyzer.go
```diff
@@ -0,0 +1,59 @@
...
+import (
+	"reflect"
+
+	"github.com/google/go-flow-levee/internal/pkg/config"
+	"golang.org/x/tools/go/analysis"
+	"golang.org/x/tools/go/analysis/passes/buildssa"
+	"golang.org/x/tools/go/ssa"
+)
+
+type ResultType = map[*ssa.Function][]*Source
```
immutableT, line 26: What do you think about including the result of the buildssa.SSA into the result?
In other words, do you think we would need access to buildssa.SSA in the higher level analyzers, or Sources is all we ever would need?

PurelyApplied: You can kind of think of analyzers consuming other analyzer results similar to how package dependencies work.

If `buildSSA` is a direct dependency, an analyzer should declare that dependency in `Analyzer.Requires`.  Doing so ensures the result is available via `ssaInput := pass.ResultOf[buildssa.Analyzer].(*buildssa.SSA)`.

The `analysis` package's framework only calls each analyzer once per package, so there's no extra work done if a downstream analyzer shares a dependency.  Such an analyzer can just declare all the relevant dependencies in its `Requires` field.  I.e.,

```golang
var Analyzer = &analysis.Analyzer {
  Run:        run,
  Requires:   []*analysis.Analyzer{source.Analyzer, buildssa.Analyzer},
  //...
}

func run(pass *analysis.Pass) (interface{}, error) {
  sources := pass.ResultOf[source.Analyzer].(source.ResultType)
  ssaInput := pass.ResultOf[buildssa.Analyzer].(*buildssa.SSA)
  //...
}
```

PR #18, opened 2020-05-15
Check for Apache 2.0 license link in Travis-CI

PurelyApplied:


PurelyApplied:
I'm not sure what the convention is for adding headers to JSON.  If there's a better approach than excluding them, I'm happy to do a more rigorous check.

immutableT:
I have not seen any license headers in k8s in json and yaml.

PR #20, opened 2020-05-18
Decouple the source and config packages via an interface.

immutableT:
Decouple the source and config packages via an interface.

### DIFF
--- a/internal/pkg/source/source.go
+++ b/internal/pkg/source/source.go
@@ -21,7 +21,6 @@ import (
 	"strings"
 
 	"github.com/eapache/queue"
-	"github.com/google/go-flow-levee/internal/pkg/config"
 	"github.com/google/go-flow-levee/internal/pkg/sanitizer"
 	"github.com/google/go-flow-levee/internal/pkg/utils"
 	"golang.org/x/tools/go/analysis/passes/buildssa"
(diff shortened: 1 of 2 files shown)


internal/pkg/source/source.go
```diff
@@ -21,14 +21,14 @@ import (
...
 	"github.com/google/go-flow-levee/internal/pkg/utils"
 	"golang.org/x/tools/go/analysis/passes/buildssa"
 	"golang.org/x/tools/go/ssa"
 )
 
 type classifier interface {
+	IsSource(types.Type) bool
 	IsSanitizer(*ssa.Call) bool
 	IsPropagator(*ssa.Call) bool
 	IsSourceFieldAddr(*ssa.FieldAddr) bool
```
PurelyApplied, line 34: Doesn't need to be part of this PR, but `type classifier interface` should likely belong to either its own package or to `config`.  We'll want to reuse this interface in other analyzers as those are extracted.

immutableT: Perhaps.
There is a school of thought that suggests to define interfaces where they are used. Therefore, according to this particular approach the interfaces is placed properly.
The reason for defining this interface is that config has too many responsibilities - a lot of which are not interesting to the Source package.

PR #23, opened 2020-06-02
Extract varargs' related logic into a separate package.

immutableT:
Extract varargs' related logic into a separate package.

### DIFF
--- a/internal/pkg/varargs/varargs.go
+++ b/internal/pkg/varargs/varargs.go
@@ -1,0 +1,89 @@
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package varargs contains functionality for dealing with variable length arguments.
+package varargs
+
+import (
+	"golang.org/x/tools/go/ssa"
+)
+
+// Varargs represents a variable length argument.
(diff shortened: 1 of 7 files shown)


internal/pkg/varargs/varargs.go
```diff
@@ -0,0 +1,89 @@
...
+	}
+
+	a, ok := sl.X.(*ssa.Alloc)
+	if !ok || a.Comment != "varargs" {
+		return nil
+	}
+
+	var (
+		stores []*ssa.Store
+	)
```
PurelyApplied, line 53: Nit: this should be one line.

internal/pkg/varargs/varargs.go
```diff
@@ -0,0 +1,89 @@
...
+
+	for _, r := range *a.Referrers() {
+		idx, ok := r.(*ssa.IndexAddr)
+		if !ok {
+			continue
+		}
+
+		if idx.Referrers() != nil && len(*idx.Referrers()) != 1 {
```
PurelyApplied, line 61: In what care would `len(*idx.Referrers())` be greater than one?  Should we have test coverage for that case?

internal/pkg/varargs/varargs.go
```diff
@@ -0,0 +1,89 @@
...
+// ReferredBy determines if the supplied node refers the Vararg in question via a store instruction.
+func (v *Varargs) ReferredBy(r referrer) bool {
+	for _, s := range v.stores {
+		if r.HasPathTo(s.Val.(ssa.Node)) {
+			return true
+		}
+	}
+
+	return false
+}
```
PurelyApplied, line 89: The unit tests do not cover the `return false` case, though the end-to-end `levee_test.go` does.  Should we cover that here?

internal/pkg/varargs/varargs.go
immutableT: Thanks for point this out - this has been bothering me.
I don't think there is such a case, and if there is one, I would rather panic and find out.
Removed.

internal/pkg/varargs/varargs.go
immutableT: Added test coverage.

internal/pkg/varargs/varargs.go
immutableT: Done.

PR #25, opened 2020-06-08
Follow the graph beyond the sanitizer.

### LINKED ISSUE #24: Source graph should continue beyond sanitizers.
Currently, when building a graph of a source we [stop](https://github.com/google/go-flow-levee/blob/master/internal/pkg/source/source.go#L86) following the graph when a sanitizer is encountered.
However, there is not guarantee that such sanitizer will dominate the yet to be discovered sinks.
Therefore, we should build the graph beyond sanitizers.
Once we know all the sinks, we could trip the graph by removing sinks that are dominated by a sanitizers.

immutableT:
Fixes #24

Currently, when building a graph of a source we stop following the graph when a sanitizer is encountered.
However, there is not guarantee that such sanitizer will dominate the yet to be discovered sinks.
Therefore, we should build the graph beyond sanitizers.
Once we know all the sinks, we could trip the graph by removing sinks that are dominated by a sanitizers.

### DIFF
--- a/internal/pkg/source/source.go
+++ b/internal/pkg/source/source.go
@@ -81,18 +81,25 @@ func (a *Source) bfs() {
 				continue
 			}
 
-			if c, ok := r.(*ssa.Call); ok && a.config.IsSanitizer(c) {
-				a.sanitizers = append(a.sanitizers, &sanitizer.Sanitizer{Call: c})
-			}
+			// TODO(immutableT) We are putting too much logic into this function - it should just
+			// be constructing a graph. Add specialized methods that know how to trim or interpret
+			// connections.
 
-			// Need to stay within the scope of the function under analysis.
-			if call, ok := r.(*ssa.Call); ok && !a.config.IsPropagator(call) {
-				continue
-			}
+			switch v := r.(type) {
+			case *ssa.Call:
+				// This is to avoid attaching calls where the source is the receiver, ex:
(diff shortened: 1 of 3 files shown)


immutableT:
I share your concerns and open to suggestions.

I see two options:
1. Let's stop testing at the level of SSA. In other words, we go back to the analyzer level testing - totally open to this. 
However, we will lose some testing granularity, and it will be harder to debug and iterate. 
2. Develop a framework for testing SSA, or at least something that will abstract away from indexes.

It is quite possible that we will no longer have the need for these fragile (index-based) tests once we gain more experience with SSA. Today, I find them quite useful despite their fragility. However, I think that is due to the lack of experience in translating Go code to SSA.

> being partial to test-driven development, my first step in review is to run the added tests without the non-test changes.

Don't think I understand your logic.
Were you expecting the tests to pass, even without the corresponding changes to the test's inputs?

PurelyApplied:
> Were you expecting the tests to pass, even without the corresponding changes to the test's inputs?

I left the test input as it was and only reverted the changes in `source.go`.  I certainly expected the tests to fail, but wanted to see *how* they'd fail.  That's why I was caught off guard with the `have a path to time.Now()`.  It makes me worry that something like a refactor will cause a hard-to-diagnose test failure in the future.

> However, I think that is due to the lack of experience in translating Go code to SSA.

Fair enough.

PR #28, opened 2020-06-26
Remove dependency of the index-based testing approach of Sources.

immutableT:
Remove dependency of the index-based testing approach of Sources.

The key idea is to use the String representation of a Source in the corresponding "want' comment, where the String representation is reduced to the the expected propagators, sanitizers and sinks attached to the source in question.

### DIFF
--- a/internal/pkg/source/source_test.go
+++ b/internal/pkg/source/source_test.go
@@ -16,7 +16,6 @@ package source
 
 import (
 	"go/types"
-	"reflect"
 	"regexp"
 	"testing"
 
@@ -29,13 +28,21 @@ import (
 )
 
 type testConfig struct {
+	sourcePattern      string
 	propagatorsPattern string
 	fieldsPattern      string
 	sanitizerPattern   string
 }
 
 func (c *testConfig) IsSource(t types.Type) bool {
-	return true
+	d := utils.Dereference(t)
+	_, ok := d.(*types.Named)
+	if !ok {
+		return false
+	}
+
+	match, _ := regexp.MatchString(c.sourcePattern, d.String())
+	return match
 }
 
 func (c *testConfig) IsSanitizer(call *ssa.Call) bool {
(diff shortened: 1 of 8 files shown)


internal/pkg/source/source_test.go
```diff
@@ -53,127 +60,56 @@ func (c *testConfig) IsSourceFieldAddr(field *ssa.FieldAddr) bool {
...
-					result.allocations = append(result.allocations, v)
-				case *ssa.Call:
-					result.calls = append(result.calls, v)
-				case *ssa.FieldAddr:
-					result.fieldAddr = append(result.fieldAddr, v)
-				case *ssa.Store:
-					result.store = append(result.store, v)
-				}
+	config := &testConfig{
+		sourcePattern:      ".foo",
```
PurelyApplied, line 73: Nit: Since it's a regex pattern, you mean `\.foo`?  Or possibly `\bfoo`.

internal/pkg/source/testdata/src/domination/test.go
```diff
@@ -25,7 +25,7 @@ type foo struct {
 }
 
 func f1() {
-	f := &foo{name: "n", password: "p"}
+	f := &foo{name: "n", password: "p"} // want `sanitizer\(t0\) sink\(t0\) `
```
PurelyApplied, line 28: Comment: I can't find it in the SSA documentation, so I might be misremembering, but I think register names like `t0` are in that *deterministic but undefined* category.  I still consider this to be an improvement in the tests, just wanted to mention it.

internal/pkg/source/source_test.go
immutableT: Good question. 
I am not certain, but it seems that the test framework considers the pattern as a regex when it is surrounded by ``.

internal/pkg/source/testdata/src/domination/test.go
immutableT: I agree - this is a bit of an unknown - we may need to revisit this.

Where possible, I tried to avoid relying on `t#`

internal/pkg/source/source_test.go
PurelyApplied: [That's correct](https://pkg.go.dev/golang.org/x/tools/go/analysis/analysistest?tab=doc#Run) but not what I meant here.

With `sourcePattern` being fed into `regexp.MatchString(c.sourcePattern, d.String())`, I read that as you wanting to match `<any package>.foo`.  Which is to say, I read that dot as intended to match the separator between package and Named, rather than a regexp match-any dot.

immutableT: Ah, good point.
Fixed.

PR #31, opened 2020-07-16
Field tags analyzer

mlevesquedion:
A minimal field tags analyzer to check whether I'm on the right track.
At this point this is mostly to establish feasibility and prompt further discussion.

### DIFF
--- a/internal/pkg/fieldtags/analyzer.go
+++ b/internal/pkg/fieldtags/analyzer.go
@@ -18,6 +18,7 @@ package fieldtags
 
 import (
 	"go/ast"
+
 	"golang.org/x/tools/go/analysis"
 	"golang.org/x/tools/go/analysis/passes/inspect"
 	"golang.org/x/tools/go/ast/inspector"
--- a/internal/pkg/fieldtags/analyzer.go
+++ b/internal/pkg/fieldtags/analyzer.go
@@ -18,21 +18,36 @@ package fieldtags
 
 import (
 	"go/ast"
+	"reflect"
+	"strings"
 
 	"golang.org/x/tools/go/analysis"
 	"golang.org/x/tools/go/analysis/passes/inspect"
 	"golang.org/x/tools/go/ast/inspector"
 )
 
+type keyValue struct {
+	key   string
+	value string
+}
+
+type sourcePatterns []keyValue
+
+var patterns sourcePatterns = []keyValue{
+	{"levee", "source"},
+}
+
 var Analyzer = &analysis.Analyzer{
-	Name:     "fieldtags",
-	Doc:      "This analyzer identifies fields tagged with SPII datapol annotations.",
-	Run:      run,
-	Requires: []*analysis.Analyzer{inspect.Analyzer},
+	Name:       "fieldtags",
+	Doc:        "This analyzer identifies Source fields based on their tags.",
+	Run:        run,
+	Requires:   []*analysis.Analyzer{inspect.Analyzer},
+	ResultType: reflect.TypeOf(new([]*ast.Field)).Elem(),
 }
 
 func run(pass *analysis.Pass) (interface{}, error) {
 	inspect := pass.ResultOf[inspect.Analyzer].(*inspector.Inspector)
+	results := []*ast.Field{}
 
 	nodeFilter := []ast.Node{
 		(*ast.Field)(nil),
@@ -43,22 +58,53 @@ func run(pass *analysis.Pass) (interface{}, error) {
 		if !ok {
 			return
 		}
-		if isSource(field) {
+		if patterns.isSource(field) {
+			results = append(results, field)
 			pass.Reportf(field.Pos(), "tagged field")
 		}
 	})
-	return nil, nil
+	return results, nil
 }
 
-func isSource(field *ast.Field) bool {
-	tag := field.Tag
-	if tag == nil {
+func (ps *sourcePatterns) isSource(field *ast.Field) bool {
+	if field.Tag == nil {
 		return false
 	}
 
-	value := tag.Value
-	if value == "`datapol:ST_ACCOUNT_CRENDENTIAL`" {
-		return true
+	// assume the tag is properly formatted
+	tag := field.Tag.Value
+
+	i := 1 // skip backtick
+	j := 1
+	for j < len(tag) {
+		for j < len(tag) && tag[j] != ':' {
+			j++
+		}
+		key := tag[i:j]
+		if key == "" {
+			return false
+		}
+
+		i = j + 2 // skip colon and opening quote
+		j = i
+		for j < len(tag) && tag[j] != '"' {
+			j++
+		}
+		value := tag[i:j]
+		if value == "" {
+			return false
+		}
+
+		for _, p := range *ps {
+			// value may be a list of comma separated values
+			if p.key == key && strings.Contains(value, p.value) {
+				return true
+			}
+		}
+
+		i = j + 2 // skip closing quote and space
+		j = i
 	}
+
 	return false
 }
--- a/internal/pkg/fieldtags/analyzer.go
+++ b/internal/pkg/fieldtags/analyzer.go
@@ -12,8 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// Package fieldtags defines an analyzer that identifies struct fields tagged with datapol annotations
-// indicating that they contain SPII.
+// Package fieldtags defines an analyzer that identifies struct fields identified
+// as sources via a field tag.
 package fieldtags
 
 import (
--- a/internal/pkg/fieldtags/analyzer.go
+++ b/internal/pkg/fieldtags/analyzer.go
@@ -43,14 +43,14 @@ func run(pass *analysis.Pass) (interface{}, error) {
 		if !ok {
 			return
 		}
-		if isTaggedPII(field) {
+		if isSource(field) {
 			pass.Reportf(field.Pos(), "tagged field")
 		}
 	})
 	return nil, nil
 }
 
-func isTaggedPII(field *ast.Field) bool {
+func isSource(field *ast.Field) bool {
 	tag := field.Tag
 	if tag == nil {
 		return false
--- a/internal/pkg/fieldtags/analyzer.go
+++ b/internal/pkg/fieldtags/analyzer.go
@@ -74,6 +74,7 @@ func (ps *sourcePatterns) isSource(field *ast.Field) bool {
 
 	tag := field.Tag.Value
 
+	// TODO: consider refactoring this logic into a regex matcher
 	i := 1 // skip backtick
 	j := 1
 	for j < len(tag) {
--- a/internal/pkg/fieldtags/analyzer.go
+++ b/internal/pkg/fieldtags/analyzer.go
@@ -48,7 +48,7 @@ var Analyzer = &analysis.Analyzer{
 
 func run(pass *analysis.Pass) (interface{}, error) {
 	inspector_ := pass.ResultOf[inspect.Analyzer].(*inspector.Inspector)
-	results := []*ast.Field{}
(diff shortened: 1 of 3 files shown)


internal/pkg/fieldtags/analyzer.go
```diff
@@ -0,0 +1,64 @@
...
+			return
+		}
+		if isTaggedPII(field) {
+			pass.Reportf(field.Pos(), "tagged field")
+		}
+	})
+	return nil, nil
+}
+
+func isTaggedPII(field *ast.Field) bool {
```
PurelyApplied, line 53: Let's keep to the Taint Analysis terms like "source."  While helping keep PII secure is one use-case of a data boundary, detecting possible SQL injection vulnerabilities would be another.  (This applies also to the package and analyzer docstrings.)

internal/pkg/fieldtags/analyzer.go
```diff
@@ -0,0 +1,64 @@
...
+	tag := field.Tag
+	if tag == nil {
+		return false
+	}
+
+	value := tag.Value
+	if value == "`datapol:ST_ACCOUNT_CRENDENTIAL`" {
```
PurelyApplied, line 60: A few things:

* This could also be part of a future CL, but as above, we'll want the tag(s) we're identifying to be generally applicable.  Rather than specifying credential, maybe just `levee:"source"` for now would be fine.
* We'll want this to be configurable in the near future.  How a codebase tags their structs and what is considered a source for the purpose of analysis is going to vary heavily between codebases.  If you don't opt for a placeholder hardcode of `levee:"source"`, you could use a list of key-value pairs instead.  (And hardcode `levee:"source"` in that format for the time being.)
* Strict equality is unlikely to work in the general case.  Many different tags may be applied to a given field.  For instance, I might want to annotate

```golang
type Foo struct {
  password string `levee:"source" json:"pwd,omitempty"
  user string
}
```
* In that vein, the convention for tags so as to be compatible with the `reflect` package is space-separated key-value pairs in the form `package:"list,of,comma,separated,values"`.  This can be checked by other `go vet` checkers.  Explicitly:

```bash
$ go-flow-levee> go vet -structtag internal/pkg/fieldtags/testdata/src/tests/test.go
internal/pkg/fieldtags/testdata/src/tests/test.go:18:2: struct field tag `datapol:ST_ACCOUNT_CRENDENTIAL` not compatible with reflect.StructTag.Get: bad syntax for struct tag value
internal/pkg/fieldtags/testdata/src/tests/test.go:19:2: struct field tag `datapol:ST_NAME` not compatible with reflect.StructTag.Get: bad syntax for struct tag value
```

internal/pkg/fieldtags/analyzer.go
```diff
@@ -0,0 +1,64 @@
...
+	inspect.Preorder(nodeFilter, func(n ast.Node) {
+		field, ok := n.(*ast.Field)
+		if !ok {
+			return
+		}
+		if isTaggedPII(field) {
+			pass.Reportf(field.Pos(), "tagged field")
+		}
+	})
+	return nil, nil
```
PurelyApplied, line 50: This could be part of a future PR, but for these results to be consumable by another analyzer, they need to be packaged into a `Result` of some kind.  That will then be exposed to downstream analyzers as you did above in 
```golang
	inspect := pass.ResultOf[inspect.Analyzer].(*inspector.Inspector)
```

A collection of `*ast.Field` would make a fine `Result`.

internal/pkg/fieldtags/analyzer.go
mlevesquedion: Those are great ideas!

I originally went with the direct equality checking just as a quick way to get going on a POC. I've implemented the key:value pairs, it seems like a natural way to do this.

I've setup go vet as a pre-push to avoid future mishaps involving badly formatted field tags. :sweat:

internal/pkg/fieldtags/analyzer.go
mlevesquedion: Since it's not  a big change, I went ahead and setup the Result. I'll integrate this analyzer with the main analyzer (levee) in a future PR.

internal/pkg/fieldtags/analyzer.go
```diff
@@ -0,0 +1,110 @@
...
+func (ps *sourcePatterns) isSource(field *ast.Field) bool {
+	if field.Tag == nil {
+		return false
+	}
+
+	tag := field.Tag.Value
+
+	i := 1 // skip backtick
+	j := 1
+	for j < len(tag) {
```
immutableT, line 79: Could this be refactored to a regex?

mlevesquedion: Here's a stab at implementing your suggestion : https://github.com/mlevesquedion/go-flow-levee/compare/field-tags-analyzer...mlevesquedion:refactor-field-tags-matching-into-regex

I think the regex version may be more readable and potentially more robust as well.

It seems I can't reference the current PR from the PR I linked above.

immutableT: I am totally fine with just a TODO here, and follow up with another PR to refactor to regex.

mlevesquedion: SG. Done.

internal/pkg/fieldtags/analyzer.go
```diff
@@ -0,0 +1,110 @@
...
+			j++
+		}
+		key := tag[i:j]
+		if key == "" {
+			return false
+		}
+
+		i = j + 2 // skip colon and opening quote
+		j = i
+		for j < len(tag) && tag[j] != '"' {
```
PurelyApplied, line 91: I'm not 100% certain on the proper spec for tags.  For instance, I'm not sure if

```
type Foo struct{
  var int "levee:`source`"
}
```
would match the the spec and pass the `structtag.Analyzer`.  Please verify, and if that is valid, add it as a test.

internal/pkg/fieldtags/analyzer.go
```diff
@@ -0,0 +1,110 @@
...
+var Analyzer = &analysis.Analyzer{
+	Name:       "fieldtags",
+	Doc:        "This analyzer identifies Source fields based on their tags.",
+	Run:        run,
+	Requires:   []*analysis.Analyzer{inspect.Analyzer, structtag.Analyzer},
+	ResultType: reflect.TypeOf(new([]*ast.Field)).Elem(),
+}
+
+func run(pass *analysis.Pass) (interface{}, error) {
+	inspect := pass.ResultOf[inspect.Analyzer].(*inspector.Inspector)
```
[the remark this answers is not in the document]
mlevesquedion: Yikes. Good catch! I think originally I wanted to call it `inspector`, but I noticed the clash with the `inspector` package, so I renamed it to `inspect`... :upside_down_face:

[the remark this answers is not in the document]
mlevesquedion: I renamed it to `inspector_`. I can't seem to think of a better name. Is using an underscore to avoid a name collision idiomatic in Go?

internal/pkg/fieldtags/analyzer.go
```diff
@@ -0,0 +1,110 @@
...
+	results := []*ast.Field{}
```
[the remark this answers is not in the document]
mlevesquedion: Interesting. So far (while learning Go) I have always used `mySlice := []type{}` as a way to declare a new slice. In my mind this is more obvious than using var, and it has the added benefit that if I ever want to initialize the slice with one or more elements, I just need to add them between the brackets. What is the rationale for using `var`?

internal/pkg/fieldtags/analyzer.go
mlevesquedion: Done.

internal/pkg/fieldtags/analyzer.go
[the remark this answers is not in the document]
PurelyApplied: Go eschews underscore naming.  While I personally dislike it, I think it would be idiomatic to name it with something abbreviated, such as `insp`.  If you want something more descriptive, `inspResult` or `inspectResult` could be fine.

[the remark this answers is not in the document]
mlevesquedion: SG. Not a fan of abbreviations either. I'll go with `inspectResult`.

internal/pkg/fieldtags/analyzer.go
```diff
@@ -0,0 +1,114 @@
...
+	{"levee", "source"},
+}
+
+var Analyzer = &analysis.Analyzer{
+	Name: "fieldtags",
+	Doc:  "This analyzer identifies Source fields based on their tags.",
+	Run:  run,
+	Requires: []*analysis.Analyzer{
+		inspect.Analyzer,
+		structtag.Analyzer, // required to ensure that tags are properly formatted
```
PurelyApplied, line 47: This doesn't appear to do what we thought.  Case in point, I was wrong about the tag convention in my other comment, and
```
  var int "levee:`source`"
```
is not properly tagged.

```bash
$ .../go-flow-levee > go vet -structtag internal/pkg/fieldtags/testdata/src/tests/test.go
# command-line-arguments
internal/pkg/fieldtags/testdata/src/tests/test.go:19:2: struct field secret has json tag but is not exported
internal/pkg/fieldtags/testdata/src/tests/test.go:20:2: struct field tag "levee:`source`" not compatible with reflect.StructTag.Get: bad syntax for struct tag value
```

Yet this doesn't appear to emit the finding as an error nor does it expose a `Result`.  I'm not sure there's a way to get a grip on that analyzer's output...

Given that, I think it could be alright if we say that tags must satisfy the `go vet -structtag` format, but not explicitly check for it since we can't get our hooks into it.

internal/pkg/fieldtags/testdata/src/tests/test.go
```diff
@@ -0,0 +1,23 @@
...
+// limitations under the License.
+
+package fieldtags
+
+type Person struct {
+	password              string `levee:"source"`               // want "tagged field"
+	secret                string `json:"secret" levee:"source"` // want "tagged field"
+	some_secret_number    int    "levee:`source`"               // want "tagged field"
+	name                  string `some_key:"non_secret"`
+	some_non_tagged_field int
```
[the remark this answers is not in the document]
mlevesquedion: Woops. My Python is showing. Fixed.

internal/pkg/fieldtags/testdata/src/tests/test.go
```diff
@@ -0,0 +1,23 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package fieldtags
+
+type Person struct {
+	password              string `levee:"source"`               // want "tagged field"
+	secret                string `json:"secret" levee:"source"` // want "tagged field"
+	some_secret_number    int    "levee:`source`"               // want "tagged field"
```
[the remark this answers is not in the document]
mlevesquedion: According to go vet, it does!

The "levee:`source`" case doesn't seem to, though : go vet says `bad syntax for struct tag value`. Not sure why I didn't catch this earlier, it looks like `go vet ./...` doesn't recurse into testdata :thinking:

internal/pkg/fieldtags/analyzer.go
mlevesquedion: That's kind of odd. I think it's reasonable to expect that the field tags will be properly formatted. Too bad we can't verify it in advance, though.

I've added a sentence to the doc to reflect our expectation.

internal/pkg/fieldtags/testdata/src/tests/test.go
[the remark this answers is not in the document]
mlevesquedion: This new case actually showed a bug in my parsing code. I think I managed to fix it without making it too much harder to understand.

PR #32, opened 2020-07-17
Do not mark an Alloc as a Source if it is produced by a Sanitizer

mlevesquedion:
This fixes a broken test in which a value return by a sanitizer with return type `Source` was marked as a source.

### DIFF
--- a/internal/pkg/source/source.go
+++ b/internal/pkg/source/source.go
@@ -198,7 +198,7 @@ func sourcesFromBlocks(fn *ssa.Function, conf classifier) []*Source {
 			switch v := instr.(type) {
 			// Looking for sources of PII allocated within the body of a function.
 			case *ssa.Alloc:
-				if conf.IsSource(utils.Dereference(v.Type())) {
+				if conf.IsSource(utils.Dereference(v.Type())) && !isProducedBySanitizer(v, conf) {
 					sources = append(sources, New(v, conf))
 				}
 
@@ -212,4 +212,21 @@ func sourcesFromBlocks(fn *ssa.Function, conf classifier) []*Source {
 		}
 	}
 	return sources
+}
+
+func isProducedBySanitizer(v *ssa.Alloc, conf classifier) bool {
+	for _, instr := range *v.Referrers() {
+		store, ok := instr.(*ssa.Store)
+		if !ok {
+			continue
+		}
+		call, ok := store.Val.(*ssa.Call)
+		if !ok {
+			continue
+		}
+		if conf.IsSanitizer(call) {
+			return true
+		}
+	}
+	return false
 }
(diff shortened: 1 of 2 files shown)


internal/pkg/source/source.go
```diff
@@ -213,3 +213,20 @@ func sourcesFromBlocks(fn *ssa.Function, conf classifier) []*Source {
 	}
 	return sources
 }
+
+func isProducedBySanitizer(v *ssa.Alloc, conf classifier) bool {
+	for _, instr := range *v.Referrers() {
+		store, ok := instr.(*ssa.Store)
```
immutableT, line 219: Would doing a `switch` here and acting only when the inst is a `*ssa.Call` read better?

internal/pkg/source/source.go
```diff
@@ -198,7 +198,7 @@ func sourcesFromBlocks(fn *ssa.Function, conf classifier) []*Source {
 			switch v := instr.(type) {
 			// Looking for sources of PII allocated within the body of a function.
 			case *ssa.Alloc:
-				if conf.IsSource(utils.Dereference(v.Type())) {
+				if conf.IsSource(utils.Dereference(v.Type())) && !isProducedBySanitizer(v, conf) {
```
immutableT, line 201: Good idea!

internal/pkg/source/source.go
mlevesquedion: I think we need to cast into a `*ssa.Store` first, which if successful means the Call leads to a Store that refers to the Alloc, otherwise if the Alloc is referred to by a sanitizing Call at any point it will get identified as not a source (a few tests broke when I tried it, which is good). I think the code makes a decent job of describing that, and I don't see a way to make it more to the point. Do you think a comment would help?

immutableT: My bad - did not read this carefully enough.

internal/pkg/source/source.go
```diff
@@ -213,3 +213,20 @@ func sourcesFromBlocks(fn *ssa.Function, conf classifier) []*Source {
...
+	for _, instr := range *v.Referrers() {
+		store, ok := instr.(*ssa.Store)
+		if !ok {
+			continue
+		}
+		call, ok := store.Val.(*ssa.Call)
+		if !ok {
+			continue
+		}
+		if conf.IsSanitizer(call) {
```
immutableT, line 227: Would the result (fix of the failing test) be the same if we asked a question of whether this call is dominated by any of the sanitizers?

PurelyApplied, line 227: I have a future concern that, when analyzers for each type are extracted, we risk a circular dependency here where the `source` identifier needs `sanitizer` information, which will need `source` information.  It's not an issue now, and a direct `config` call is fine here.  Just something to keep in mind for future iteration.

PR #33, opened 2020-07-21
Consolidate fieldPropogatorMatcher and NameMatcher

PurelyApplied:
* methodMatcher subsumes both, matching package, receiver, and method names.

------

I noticed that `NameMatcher.TypeRE` was unused -- its only use being in a function that was nowhere called.  

Judging by the dead `NameMatcher.matchNamedType` method, this was possibly due to a divergence in type between `fieldPropogatorMatcher` and `NameMatcher`.  I can see no reason for the underlying implementation of these to be kept separate, and so have unified them here.

This could also apply to the distinction between `FieldPropagator` and `TranformingPropagator`.  We should consider unifying both the implementation and the configuration of these to simply be `Propogator` in the future.

- [x] Tests pass

### DIFF
--- a/configuration/README.md
+++ b/configuration/README.md
@@ -28,19 +28,21 @@ Sources are identified via regexp according to package, type, and field names.
 }
 ```
 
-Sinks and sanitizers are identified via regexp according to package and method name.
+Sinks, sanitizers, and some propagators (below) are identified via regexp according to package, method, and (optional) receiver name.
 
 ```json
 {
   "Sinks": [
     {
       "PackageRE": "<package path regexp>",
+      "ReceiverRE": "<type name regexp>",
       "MethodRE": "<method name regexp>"
     }
   ],
   "Sanitizers": [
     {
       "PackageRE": "<package path regexp>",
+      "ReceiverRE": "<type name regexp>",
       "MethodRE": "<method name regexp>"
     }
   ]
@@ -69,13 +71,15 @@ Argument propagators are identified via regexp matching the fully-qualified type
   "TransformingPropagators": [
     {
       "PackageRE": "<package path regexp>",
+      "ReceiverRE": "<type name regexp>",
       "MethodRE": "<method name regexp>"
     }
   ],
   "FieldPropagators": [
     {
-      "Receiver": "<fully qualified type path regexp>",
-      "AccessorRE": "<method name regexp>"
+      "PackageRE": "<package path regexp>",
+      "ReceiverRE": "<type name regexp>",
+      "MethodRE": "<method name regexp>"
     }
   ],
   "PropagatorArgs": {
@@ -83,6 +87,9 @@ Argument propagators are identified via regexp matching the fully-qualified type
   }
 }
 ```
+
+For matchers that accept a `ReceiverRE` regexp matcher, an unspecified string will match any (or no) receiver.
+To match only methods without any receiver (i.e., a top-level function), use the matcher `^$` to match an empty-string receiver name.
 
 ### Example configuration
 
--- a/internal/pkg/config/config.go
+++ b/internal/pkg/config/config.go
@@ -38,9 +38,9 @@ func init() {
 // config contains matchers and analysis scope information
 type Config struct {
 	Sources                 []sourceMatcher
-	Sinks                   []methodMatcher
-	Sanitizers              []methodMatcher
-	FieldPropagators        []methodMatcher
+	Sinks                   []callMatcher
+	Sanitizers              []callMatcher
+	FieldPropagators        []callMatcher
 	TransformingPropagators []transformingPropagatorMatcher
 	PropagatorArgs          argumentPropagatorMatcher
 	Whitelist               []packageMatcher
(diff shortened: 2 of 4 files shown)


internal/pkg/config/config.go
```diff
@@ -268,33 +268,38 @@ func (pm packageMatcher) match(pkg *types.Package) bool {
 	return pm.PackageNameRE.MatchString(pkg.Path())
 }
 
-type NameMatcher struct {
-	PackageRE regexp.Regexp
-	TypeRE    regexp.Regexp
-	MethodRE  regexp.Regexp
+type methodMatcher struct {
```
mlevesquedion, line 271: Maybe just `matcher`, since this can match more than just methods?

internal/pkg/config/config.go
```diff
@@ -268,33 +268,38 @@ func (pm packageMatcher) match(pkg *types.Package) bool {
...
-		// In our context, both sources and sanitizers are concrete types.
+	if !r.matchPackage(callee.Pkg.Pkg) || !r.MethodRE.MatchString(callee.Name()) {
 		return false
 	}
 
-	return r.PackageRE.MatchString(n.Obj().Pkg().Path()) &&
-		r.TypeRE.MatchString(n.Obj().Name())
+	recv := c.Call.Signature().Recv()
+	var recvName string
+	if recv == nil {
```
mlevesquedion, line 296: I think you could use `!=` here and move the `else` case into the `if`, since the default value for `string` is `""`.

configuration/README.md
```diff
@@ -28,19 +28,21 @@ Sources are identified via regexp according to package, type, and field names.
...
-Sinks and sanitizers are identified via regexp according to package and method name.
+Sinks, sanitizers, and some propagators (below) are identified via regexp according to package, method, and (optional) receiver name.
 
 ```json
 {
   "Sinks": [
     {
       "PackageRE": "<package path regexp>",
+      "ReceiverRE": "<type name regexp>",
       "MethodRE": "<method name regexp>"
```
mlevesquedion, line 39: I think it would make more sense to call this field "FunctionRE" or something similar. Indeed if there is an explicit receiver, then it is a method, but if there isn't, then it is a function, but methods are functions, so "FunctionRE" would capture both.

configuration/README.md
```diff
@@ -84,6 +88,9 @@ Argument propagators are identified via regexp matching the fully-qualified type
 }
 ```
 
+For matchers that accept a `ReceiverRE` regexp matcher, an unspecified string will match any (or no) receiver.
```
mlevesquedion, line 91: Should we specify the (empty) `ReceiverRE` fields anyway in `example-config.json`? I feel like the behavior related to unspecified strings is an artifact of how Go handles JSON unmarshalling + default values, so I think it could be clearer to mention the field explicitly in the example config.

internal/pkg/config/config.go
```diff
@@ -76,7 +76,7 @@ func (c Config) shouldSkip(pkg *types.Package) bool {
 
 func (c Config) IsSink(call *ssa.Call) bool {
 	for _, p := range c.Sinks {
-		if p.MatchMethodName(call) {
+		if p.Match(call) {
```
mlevesquedion, line 79: Neat!

internal/pkg/config/config.go
```diff
@@ -268,33 +268,38 @@ func (pm packageMatcher) match(pkg *types.Package) bool {
...
 }
 
-func (r NameMatcher) MatchMethodName(c *ssa.Call) bool {
-	if c.Call.StaticCallee() == nil || c.Call.StaticCallee().Pkg == nil {
+// Match matches methods based on package, method, and receiver regexp.
+// To explicitly match a method with no receiver (i.e., a top-level function),
+// provide the ReceiverRE regexp `^$`.
+func (r methodMatcher) Match(c *ssa.Call) bool {
+	callee := c.Call.StaticCallee()
+	if callee == nil || callee.Pkg == nil {
```
mlevesquedion, line 286: It hadn't occurred to me yet to do this. Stealing this trick for future use. :smile:

internal/testdata/test-config.json
```diff
@@ -20,8 +20,9 @@
   ],
   "FieldPropagators": [
     {
-      "Receiver": "example.com/core.Source",
-      "AccessorRE": "^GetData$",
+      "PackageRE": "^example.com/core$",
+      "ReceiverRE": "Source",
+      "MethodRE": "^GetData$",
       "TODO": "Make this explicit configuration unnecessary."
```
mlevesquedion, line 26: Does this TODO apply only to `FieldPropagators`? Would you like me to make it into an Issue for discussion?

If I understand correctly, we should be able to identify field propagators automatically, i.e. a field propagator is a method on a `Source` type that returns a tainted value (by which I mean a `Source`, a `Source`'s field or a value derived from them).

configuration/README.md
mlevesquedion: (I would have also suggested adding comments directly into the example config, but JSON doesn't support comments :disappointed:)

internal/pkg/config/config.go
mlevesquedion: Actually, maybe `callMatcher` would be more expressive, since that is what we are trying to match in the end?

internal/testdata/test-config.json
mlevesquedion: (I guess we should also be able to identify TransformingPropagators automatically, e.g. in the current context any function that has a `Source` as a parameter, eventually a call that receives a `Source` value).

configuration/README.md
PurelyApplied: That's a good point.  That's kind of beyond the scope of this PR though.  I'll open a rename PR after this one.

internal/pkg/config/config.go
PurelyApplied: Astute observation.  Condensed.

internal/pkg/config/config.go
PurelyApplied: Renamed to `callMatcher`.

internal/testdata/test-config.json
PurelyApplied: I think the TODO was specifically referencing FieldPropagators, like simple getters.  Some of this comes from auto-generated code, like the structs from generated by protobuf.

I think if the getter returns a type identifiable as a source, it's moot, since the call will be identified as a source in the current, type-based identification.  But a field getter that, e.g., returns a secret string or `interface{}` would go unnoticed, and these are the things we would want to detect.

We should be able to identify these, but haven't gotten to it yet.  Thus this lingering TODO.

mlevesquedion: SG. I'll convert it to an issue once I do my TODO cleanup thing.

configuration/README.md
PurelyApplied: Good catch.  Added.

PR #34, opened 2020-07-22
Detect non variadic sinks

mlevesquedion:
Fixes test TODOs involving non variadic sinks.

Please ignore the 24 extra commits. Look at the `Files changed` tab instead.

- [x] Tests pass


internal/pkg/call/call_test.go
```diff
@@ -0,0 +1,103 @@
...
+}
+`
+
+type fakeReferrer struct {
+	hasPathTo bool
+}
+
+func (f fakeReferrer) HasPathTo(node ssa.Node) bool { return f.hasPathTo }
+
+func TestRegularCallWithNoArgsIsNotReferredBy(t *testing.T) {
```
immutableT, line 46: nit: Please take a look at table-driven-test.
https://dave.cheney.net/2019/05/07/prefer-table-driven-tests
It is the preferred method both in k8t and in google3.

internal/pkg/call/call_test.go
```diff
@@ -0,0 +1,103 @@
...
+	"go/parser"
+	"go/token"
+	"go/types"
+	"testing"
+
+	"golang.org/x/tools/go/ssa"
+	"golang.org/x/tools/go/ssa/ssautil"
+)
+
+const source = `package test
```
immutableT, line 29: nit: We prefer to keep the test data (sources) in the `testdata` directory.
https://tutorialedge.net/golang/advanced-go-testing-tutorial/#use-the-testdata-directory

internal/pkg/call/call.go
```diff
@@ -0,0 +1,51 @@
...
+// call refers to a given referrer.
+package call
+
+import (
+	"golang.org/x/tools/go/ssa"
+)
+
+// Call is a wrapper for *ssa.Call instructions, allowing Variadic and
+// non-Variadic calls to be handled polymorphically
+type Call interface {
```
immutableT, line 25: Future improvements:
I am curios what it would look like if we were to combine `varargs` pkg (which is really a call) with this one.
In other words, this package would hide the complexity involved in dealing with regular calls vs. variadic.

internal/pkg/call/call.go
```diff
@@ -0,0 +1,51 @@
...
+	// ReferredBy determines if the call is referred to by the referrer
+	ReferredBy(r Referrer) bool
+}
+
+type Referrer interface {
+	HasPathTo(node ssa.Node) bool
+}
+
+type RegularCall struct {
```
immutableT, line 34: nit: Public methods should have comments.

internal/pkg/call/call.go
mlevesquedion: Agreed. In fact, I think this new `Call` concept could become a whole `Sink` type, possibly produced by a separate `Analyzer`.

internal/pkg/call/call_test.go
mlevesquedion: Sounds good, I've merged it with the existing table driven test. I think the result is clearer!

internal/pkg/call/call.go
mlevesquedion: Done.

internal/pkg/call/call_test.go
mlevesquedion: Sounds good! In this case I felt like having the source directly in the file made for easier browsing, but I will move it to testdata. More consistent with the rest of the codebase that way.

internal/pkg/call/call_test.go
```diff
@@ -0,0 +1,97 @@
...
+		ssaCall := getCall(source, c.parentFuncName)
+		call := Regular(ssaCall)
+		got := call.ReferredBy(c.r)
+		if got != c.want {
+			t.Errorf("call.ReferredBy(%v) == %v, want %v", call, got, c.want)
+		}
+	}
+}
+
+func readFromTestData(filename string) string {
```
immutableT, line 59: Typically we pass the `t` to all helper methods and:
1. Mark methods as helpers:  `t.Helper()`
2. Instead of `panic` call t.Fatal(err)

internal/pkg/call/call_test.go
```diff
@@ -0,0 +1,97 @@
...
+
+	"golang.org/x/tools/go/ssa"
+	"golang.org/x/tools/go/ssa/ssautil"
+)
+
+type fakeReferrer struct {
+	hasPathTo bool
+}
+
+func (f fakeReferrer) RefersTo(node ssa.Node) bool { return f.hasPathTo }
```
immutableT, line 36: This style is not common in Golang.

internal/pkg/call/call_test.go
```diff
@@ -0,0 +1,97 @@
...
+	bytes, err := ioutil.ReadFile(filePath)
+	if err != nil {
+		panic(err)
+	}
+	return string(bytes)
+}
+
+// taken from golang.org/x/tools/go/ssa/example_test.go
+func getCall(source, parentFuncName string) *ssa.Call {
+	fset := token.NewFileSet()
```
immutableT, line 70: This is very cool.

However, I want us to standardize on a single approach for extracting things from testdata.
Up until this point we have been leveraging `&analysis.Analyzer` to extract interesting things for testing.

For example:
https://github.com/google/go-flow-levee/blob/master/internal/pkg/source/source_test.go#L63

Do you think we could use `&analysis.Analyzer` here is well?

internal/levee.go
```diff
@@ -18,12 +18,14 @@ import (
 	"fmt"
```
immutableT, line 18: Sorry, for taking so long to come back to this PR.
I assumed that Patrick would complete the review. In the future, please ping me directly if you are not getting your work reviewed within a day.

internal/pkg/call/call.go
```diff
@@ -0,0 +1,55 @@
...
+
+// A Referrer is something that can be queried to determine whether it refers to
+// a given ssa Node.
+type Referrer interface {
+	RefersTo(node ssa.Node) bool
+}
+
+// A RegularCall is a wrapper around a non-variadic ssa Call.
+type RegularCall struct {
+	call *ssa.Call
```
immutableT, line 38: nit: When embedding/wrapping it is common in Go to omit the name of the var.

```
type RegularCall struct {
 *ssa.Call
}
```

internal/pkg/call/call_test.go
```diff
@@ -0,0 +1,97 @@
...
+type fakeReferrer struct {
+	hasPathTo bool
+}
+
+func (f fakeReferrer) RefersTo(node ssa.Node) bool { return f.hasPathTo }
+
+func TestRegularCallReferredBy(t *testing.T) {
+	source := readFromTestData("test.go")
+	cases := []struct {
+		r              Referrer
```
immutableT, line 41: Consider adding a field for test description.

internal/pkg/call/call_test.go
```diff
@@ -0,0 +1,97 @@
...
+		parentFuncName string
+		want           bool
+	}{
+		{fakeReferrer{true}, "CallNoArgs", false},
+		{fakeReferrer{true}, "CallOneArg", true},
+		{fakeReferrer{false}, "CallOneArg", false},
+	}
+	for _, c := range cases {
```
immutableT, line 49: Typically, we structure this loop like this:
````
   for _, tt := range cases {
		t.Run(tt.description, func(t *testing.T) {
			// Test logic
		})
	}
```

internal/pkg/call/call_test.go
mlevesquedion: Fixed. Weird that gofmt allows this. I'll add it to my personal linter.

internal/pkg/call/call_test.go
mlevesquedion: I think it would be possible to use `&analysis.Analyzer` + `analysistest`, but in this case I would prefer the unit test. I think turning this into an analyzer would make the code significantly more complex than it needs to be.

internal/levee.go
mlevesquedion: No worries at all. Thank you for the detailed review! This is exactly the kind of feedback I need to make progress in my ability to write idiomatic Go.

internal/pkg/call/call.go
mlevesquedion: Very handy! Doesn't that mean the field is now exported though?

internal/pkg/call/call_test.go
```diff
@@ -0,0 +1,105 @@
...
+
+func (f fakeReferrer) RefersTo(node ssa.Node) bool { return f.hasPathTo }
+
+func TestRegularCallReferredBy(t *testing.T) {
+	source := readFromTestData(t, "test.go")
+	cases := []struct {
+		r              Referrer
+		parentFuncName string
+		want           bool
+		desc           string
```
immutableT, line 44: nit: place `desc` at the top.

internal/pkg/call/call_test.go
```diff
@@ -0,0 +1,105 @@
...
+			call := Regular(ssaCall)
+			got := call.ReferredBy(tt.r)
+			if got != tt.want {
+				t.Errorf("call.ReferredBy(%v) == %v, want %v", call, got, tt.want)
+			}
+		})
+	}
+}
+
+func readFromTestData(t *testing.T, filename string) string {
```
immutableT, line 62: This would also go away if you were to use analysistest.
`dir := analysistest.TestData()`

internal/pkg/call/call_test.go
immutableT: > I think turning this into an analyzer would make the code significantly more complex than it needs to be.

I would disagree with this. I think that using `analysistest` would simplify the logic of testing because:
1. `readFromTestData` would go away
2. `analysistest` hides the details of how SSA components (ex. calls) are extracted from sources
3. (for better of worse) we already standardized on the approach of `analysistest`.

internal/pkg/call/call_test.go
mlevesquedion: SG. Done.

internal/pkg/call/call_test.go
mlevesquedion: I agree that the *testing* code would be simpler, but the code for handling calls would be more complex (in my opinion). I think you raise important points though. I'll give it a shot.

internal/pkg/call/call.go
```diff
@@ -0,0 +1,55 @@
...
+// ReferredBy determines if the supplied referrer refers one of the Call's arguments.
+func (c *RegularCall) ReferredBy(r Referrer) bool {
+	for _, a := range c.Call.Call.Args {
+		if r.RefersTo(a.(ssa.Node)) {
+			return true
+		}
+	}
+
+	return false
+}
```
immutableT, line 55: What do you think about the pattern we used in testing the source pkg?
Concretely, `source` implements the `Stinger` interface, and the logic there describes how a source is connected to potential sinks. 
During tests, once we find a `source`, we report it (within the analyzer) by calling its String method.
https://github.com/google/go-flow-levee/blob/master/internal/pkg/source/source_test.go#L83

In our testdata, we annotate sources with the expected string representation of a `source`.
https://github.com/google/go-flow-levee/blob/master/internal/pkg/source/testdata/src/allocation/test.go#L22

How is this helpful:
1. Tests become more declarative - simply annotate the calls under tests with the expected string representation (ex. regular call, variadic call).
2. We no longer need to collect the calls in the analyzer and have these checks that the expected number of calls was collected.

WDYT?

mlevesquedion: (For context, while writing the tests I was using the `varargs` package as a reference.)

I think the pattern you are describing makes sense for the `source` package, but I don't think it makes sense for `call` or `varargs`, because we would need to turn them into full-blown analyzers, which I feel isn't necessary for them to do their job and therefore would make the code harder to understand. The testing code would certainly be cleaner, but the actual implementation code would be more complicated, which I think would be detrimental overall. 

I've reorganized the code in a way that I think makes it more readable. Specifically, I put the test function first, so the reader doesn't have to slog through the analyzer's definition, which is not necessary to understand the test.

internal/pkg/call/call_test.go
```diff
@@ -0,0 +1,113 @@
...
+
+func run(pass *analysis.Pass) (interface{}, error) {
+	in := pass.ResultOf[buildssa.Analyzer].(*buildssa.SSA)
+	var result analyzerResult
+	for _, fn := range in.SrcFuncs {
+		for _, b := range fn.Blocks {
+			for _, i := range b.Instrs {
+				switch v := i.(type) {
+				case *ssa.Call:
+					result.calls = append(result.calls, v)
```
immutableT, line 107: Here, instead of collecting results, do `v.String()` - this will emit a finding (something like: "Referred by X"). Assuming that you have a corresponding `want` comment and the source code input has referrer X, this will be all that we need to test for.

This will allow us to remove the fake referrer, but rather rely on the source code input to setup the testcase.

internal/pkg/call/call.go
immutableT: > For context, while writing the tests I was using the `varargs` package as a reference.
   
    Currently, we are inconsistent in how we organize our tests. The pattern we introduced in the source pkg came after the 
    vararg pkg was implemented. This was on my TODO list to bring consistency in how our tests are organized.

> ... because we would need to turn them into full-blown analyzers
   PTAL at my comment here: https://github.com/google/go-flow-levee/pull/34/files#r465973836

Please note that your code is well organized - this is not the issue here.
The reason why I am bringing this up is consistency, and my hope was to use the pattern we currently use in the `source` pkg.

mlevesquedion: I see, thank you for explaining!

internal/pkg/call/call_test.go
mlevesquedion: I'm not sure how to set this up. I can't just do `v.String()`, I would need to do `pass.Reportf(...)`. I'm not sure what I would put in the "want" comment though... more importantly, I don't know how I would produce a report that would allow me to test that `RegularCall.ReferredBy` is working properly. I need to create a `RegularCall` at some point, which I could do in the `Analyzer` we have here. But then, I would still need the fakeReferrer... I'm also not sure how I would cover the case where the referrer *does not* refer to the call.

PR #37, opened 2020-07-29
reorganize analysis

mlevesquedion:
This PR represents a significant change in how the analysis is conducted. I believe that these changes make it easier to understand how the analysis works, as well as offering a clearer path to the extensions that we are envisioning.

This PR also adds support for detecting the sinking of tainted slices. In fact, this is how the PR started out. The approach I took to add support for slices led me to wonder if it could be generalized to support existing detections. It can. Also, I believe this new approach obsoletes the PR I wrote implementing detection of non-variadic sinks (#34).

I propose a 2-step approach:
1. Visit each instruction in a function's body in order to construct a graph representing the flow of values. During construction, we also identify sources, sinks, and sanitizers.
2. For each source, perform a DFS. Do not recurse through sanitizers. If a sink is reached, report it.

I believe that by putting the value graph front and center, this new approach makes the analysis easier to understand and extend. Future efforts could use this analysis to produce a summary of the behavior of a function w.r.t. our analysis, e.g. if a method on a struct returns a source field, we could identify a call to this method as producing a source. This would allow us to identify propagators automatically.

At present, the code includes options to produce debugging output. This requires that a folder named "out" is present under the "internal" folder. The debugging output includes, for each function analyzed, a text representation of the SSA as well as a graphviz (DOT) source file which can be used to produce an image of the value graph. This output has been instrumental to me in producing this PR.

- [x] Tests pass


immutableT:
Do you think it would be possible to factor our the debugging enchantments into a separate PR to reduce the diff?

mlevesquedion:
Sure. I've removed the debugging code for now.

internal/pkg/config/config.go
```diff
@@ -123,21 +141,27 @@ func (c Config) IsSourceField(typ types.Type, fld *types.Var) bool {
 }
 
 func (c Config) IsSourceFieldAddr(fa *ssa.FieldAddr) bool {
```
PurelyApplied, line 143: A docstring might be helpful here now that the complexity has grown.

If I'm reading this correctly, the intent is:
* Check if the accessed field itself is a source type.  If so, return `true`.
* Check if the wrapping struct containing the field is a source type.  If so, check that the accessed field is marked.

It could be part of a future PR, but given the fact that we're matching based on names here, we're going to have a problem for fields annotated as sources by a field tag.

internal/pkg/config/config.go
```diff
@@ -123,21 +141,27 @@ func (c Config) IsSourceField(typ types.Type, fld *types.Var) bool {
...
+	field := s.Field(fa.Field)
 	for _, p := range c.Sources {
-		if p.match(n) && p.FieldRE.MatchString(fieldName) {
+		if p.match(named) && p.matchFieldName(field) {
 			return true
 		}
 	}
+
 	return false
 }
```
PurelyApplied, line 166: This loop can be replaced by the existing `c.IsSourceField`.

internal/pkg/config/config.go
```diff
@@ -222,6 +246,14 @@ func (s sourceMatcher) match(n *types.Named) bool {
 	return s.PackageRE.MatchString(n.Obj().Pkg().Path()) && s.TypeRE.MatchString(n.Obj().Name())
 }
 
+func (s sourceMatcher) matchFieldName(f *types.Var) bool {
+	return s.FieldRE.MatchString(f.Name())
+}
+
```
PurelyApplied, line 252: If the above loop in line 160 is replaced by `Config.IfSourceField`, this method becomes dead.

internal/levee.go
```diff
@@ -39,75 +41,210 @@ func run(pass *analysis.Pass) (interface{}, error) {
 	if err != nil {
 		return nil, err
 	}
-	// TODO: respect configuration scope
 
+	// TODO: respect configuration scope
```
PurelyApplied, line 45: Weird whitespace diff.  My moral opposition to `TODO`s says we should just trim this comment if we're touching the area, and open an Issue about being able to limit analysis scope.

internal/levee.go
```diff
@@ -39,75 +41,210 @@ func run(pass *analysis.Pass) (interface{}, error) {
...
+
+	// this is needed to capture variables referred to by closures
+	for _, p := range fn.FreeVars {
+		switch t := p.Type().(type) {
+		case *types.Pointer:
+			if s, ok := utils.Dereference(t).(*types.Named); ok && conf.IsSource(s) {
+				graph.addSource(p.Name(), p)
 			}
+		}
+	}
```
PurelyApplied, line 72: This work was already done by the `source` analyzer, which should remain the source of truth for Source identification.  Use `sourcesMap[fn]` for a list of `*Source`s.  Export `Source.node` if you need it for your graph.

internal/levee.go
```diff
@@ -39,75 +41,210 @@ func run(pass *analysis.Pass) (interface{}, error) {
...
+	for _, b := range fn.Blocks {
+		for _, inst := range b.Instrs {
+			value, _ := inst.(ssa.Value)
+
+			switch t := inst.(type) {
+			case *ssa.Alloc:
+				if conf.IsSource(utils.Dereference(t.Type())) && !source.IsProducedBySanitizer(t, conf) {
+					graph.addSource(value.Name(), t)
+				}
+
```
PurelyApplied, line 83: This work was already done by the `source` analyzer, which should remain the source of truth for Source identification.

internal/levee.go
```diff
@@ -39,75 +41,210 @@ func run(pass *analysis.Pass) (interface{}, error) {
...
+				case conf.IsSanitizer(t):
+					graph.addSanitizer(t)
+				default:
+					graph.addCallEdges(*t, parameterNames)
 				}
+
+			case *ssa.FieldAddr:
+				if conf.IsSourceFieldAddr(t) {
+					graph.addSource(value.Name(), t)
+				}
```
PurelyApplied, line 104: This work was already done by the `source` analyzer, which should remain the source of truth for Source identification.

internal/levee.go
```diff
@@ -39,75 +41,210 @@ func run(pass *analysis.Pass) (interface{}, error) {
...
+			case *ssa.Slice:
+				x, _ := t.X.(ssa.Value)
+				graph.addEdge(x.Name(), value.Name())
+
+			case *ssa.Store:
+				graph.addEdge(t.Val.Name(), t.Addr.Name())
+
+			case *ssa.UnOp:
+				x, _ := t.X.(ssa.Value)
+				graph.addEdge(x.Name(), value.Name())
```
PurelyApplied, line 131: If I'm not mistaken, this addition is what now enables correct taint propagation to collections?  It feels as though this could have been its own PR.

internal/levee.go
```diff
@@ -39,75 +41,210 @@ func run(pass *analysis.Pass) (interface{}, error) {
...
+		name: call.Call.StaticCallee().Name() + strconv.Itoa(i),
+		call: call,
 	}
+}
 
-	firstArg := call.Call.Signature().Params().At(0)
-	if c.PropagatorArgs.ArgumentTypeRE.MatchString(firstArg.Type().String()) {
-		if a, ok := call.Call.Args[0].(*ssa.MakeInterface); ok {
-			return a.X.(ssa.Node)
+type graph struct {
```
PurelyApplied, line 159: It's not clear to me what this new `type graph` does for us that can't be done with the existing `ssa.Node` graph.  It appears that we're reproducing a great deal of effort that's already been done for us.

More critically, this graph relies heavily on `Node.Name()` as identification.  Nodes names are not to be used semantically.  From [the `ssa` docs](https://godoc.org/golang.org/x/tools/go/ssa#Value), with focus on the third doc section:

> ```golang
> type Value interface {
>     // Name returns the name of this value, and determines how
>     // this Value appears when used as an operand of an
>     // Instruction.
>     //
>     // This is the same as the source name for Parameters,
>     // Builtins, Functions, FreeVars, Globals.
>     // For constants, it is a representation of the constant's value
>     // and type.  For all other Values this is the name of the
>     // virtual register defined by the instruction.
>     //
>     // The name of an SSA Value is not semantically significant,
>     // and may not even be unique within a function.
>     Name() string
> // ...
> ```

This graph idea is the direction we want to be heading, though.  My own thoughts were to use the existing `ssa.Node` graph, highlight those nodes which are sources, sinks, etc via independent analyzers, and then have `levee` perform a traversal of that graph to report reachability.

I'm not opposed to `putting the value graph front and center`, I just worry about reinventing a wheel.

internal/pkg/config/config.go
```diff
@@ -94,6 +94,24 @@ func (c Config) IsSanitizer(call *ssa.Call) bool {
...
+func (c Config) IsSinkName(name string) bool {
+	for _, p := range c.Sinks {
+		if p.MethodRE.MatchString(name) {
+			return true
+		}
+	}
+	return false
+}
+
```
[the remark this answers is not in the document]
mlevesquedion: Good catch. I should have deleted those along with the debugging code.

internal/pkg/config/config.go
mlevesquedion: Agreed. In fact given the complexity I think it would be useful to add unit tests and refactor, but that would definitely need its own PR.

Concerning field tags, I think it could be done in this function (that would increase the complexity a bit more, but to me it feels like the right place to put it).

internal/pkg/source/source.go
```diff
@@ -140,6 +140,10 @@ func (a *Source) String() string {
 	return b.String()
 }
 
+func (a *Source) Name() string {
+	return a.node.String()
+}
+
```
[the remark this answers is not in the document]
mlevesquedion: More debugging code. My bad.

internal/testdata/src/example.com/tests/collections/tests.go
```diff
@@ -0,0 +1,27 @@
...
+import (
+	"example.com/core"
+)
+
+func TestSlices(s core.Source) {
+	slice := []core.Source{s}
+	core.Sink(slice)                   // want "a source has reached a sink"
+	core.Sink([]core.Source{s})        // want "a source has reached a sink"
+	core.Sink([]interface{}{s})        // want "a source has reached a sink"
+	core.Sink([]interface{}{0, "", s}) // want "a source has reached a sink"
```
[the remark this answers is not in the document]
mlevesquedion: The SSA with and without spreading is nearly identical. I've added the case just to be explicit.

I was planning on handling other collections in a future PR. Is that ok?

internal/levee.go
mlevesquedion: I'm not sure I see the benefits of keeping the `source` analyzer as the source of truth for Source identification. One benefit of having separate analyzers is that we can test them separately and use `analysistest` to test Source detection specifically. However, I think Source detection could be tested more simply via unit tests. If I replace the Requires on the `source` analyzer with a Requires on the `buildssa` analyzer and iterate directly on `buildssa.ResultType`'s `srcFuncs` field, the tests still pass. Given that, we could transform the `source` analyzer into a `source` package that would provide utilities for Source identification (if needed; this would be on top of config). To me the biggest benefit with this approach would be simpler code. Not that having separate analyzers makes things particularly complex.

Are there benefits/limitations to each approach that I am missing?

internal/levee.go
mlevesquedion: The concerns you raise about `Name()` are valid. I was aware of the warnings in the documentation, but given that the test cases that we have right now pass and that I didn't find anything weird or inconsistent when viewing the SSA and the value graph for a dozen or so functions in our tests, I thought it was reliable enough for us to use.

In my original investigation of Nodes, I had a hard time constructing a graph that was useful for our analysis. See e.g. [this doc](https://docs.google.com/document/d/1gp6FbYoZKqLN5AR6_1Tuqjpyd5oi5ycNrUu1FiiIEGQ/preview).

For what it's worth, the documentation for Node reads : 
```
// Node is provided to simplify SSA graph algorithms.  Clients should
// use the more specific and informative Value or Instruction
// interfaces where appropriate.
```

I'm not overly concerned about reinventing the wheel if it doesn't require much code and allows us to see our core logic more clearly. I think this is the case right now, but I may well be missing something.

internal/levee.go
PurelyApplied: > we are not respecting configuration scope right now? Do you mean packages? If so I believe we are at least somewhat respecting that right now, is that correct? 🤔

At present, `Config.AnalysisScope` isn't used.  It is reachable through `Config.shouldSkip(pkg)`, but that method isn't called anywhere.

The idea had been that we don't necessarily want to waste cycles scanning every dependency when the sources and sinks we care about are in some higher-level packages.  A call to `conf.shouldSkip(pass.Pkg)` used to be here, but it got replaced by this `TODO` when this project was open-sourced.

By "respect configuration scope," I meant  "we should be able to skip packages that we recognize as irrelevant."

There had been some other hard-coded logic about automatically skipping any package that imports `testing`, since that's presumably a test package.  Looking at it now, there was also a contradicting allow-list that would skip some packages even if they were "in scope."

All this to say, we're due to reexamine both how configuration is provided and how that configuration is consumed to decide which packages to actually analyze.

internal/levee.go
PurelyApplied: Avoiding duplicated effort was my primary concern.  Whether that comes from a `source` analyzer or `source` package functions is somewhat immaterial to me.

> However, I think Source detection could be tested more simply via unit tests.

So far, immutableT and I haven't managed to find a clean way to unit-test SSA steps that don't rely on `analysis.Analyzer`s and that aren't either brittle or heavily hard-coded.  If you see a good path forward that doesn't include us writing an entire test harness, I'm all for it.

> To me the biggest benefit with this approach would be simpler code. Not that having separate analyzers makes things particularly complex.

Is there an extra negative in here?  This sounds like you prefer distinct analyzers as a means to simpler code, with which I agree.  But that also seems to contradict the previous portion of the paragraph.

internal/levee.go
PurelyApplied: > but given that the test cases that we have right now pass and that I didn't find anything weird or inconsistent when viewing the SSA and the value graph for a dozen or so functions in our tests, I thought it was reliable enough for us to use.

Not to be glib, but our relatively simple test suite and some spot-checking is hardly reason enough to disregard a documented best practice.  But more to the point, there's no reason we need to be using strings as keys here.  Case in point...

> It seems to me like the documentation for Node suggests that in some cases using the Values might be ok?

Yes, using a `ssa.Node`, `ssa.Value`, or `ssa.Instruction` as the key to a `map` would be fine.  But at that point, the graph being produced should be the same as that presented via `Value.Referrers()` and `Instruction.Operands`.

internal/testdata/src/example.com/tests/collections/tests.go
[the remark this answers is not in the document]
PurelyApplied: Yep, that's fine.  Just making sure it was a matter of intent and not an oversight.

internal/levee.go
mlevesquedion: > So far, immutableT and I haven't managed to find a clean way to unit-test SSA steps that don't rely on analysis.Analyzers and that aren't either brittle or heavily hard-coded. If you see a good path forward that doesn't include us writing an entire test harness, I'm all for it.

I believe I do have an approach that isn't too brittle. See #34. If we want to work with the SSA, I don't see a way to avoid writing **some** boilerplate. In my opinion, if we have functions that operate directly on SSA values/instructions, then it makes more sense to write a bit of code to get just the SSA than to write an analyzer and use analysistest.

> Is there an extra negative in here? This sounds like you prefer distinct analyzers as a means to simpler code, with which I agree. But that also seems to contradict the previous portion of the paragraph.

I believe I didn't express myself clearly. I think having separate analyzers can lead to simpler code, but in this case I don't see a clear case for having a separate analyzer just for source identification, so to me it actually makes the code more complex. Identifying sources seems to me a simple task that can be done while constructing/traversing the graph, so having a separate analyzer just to do that seems like unnecessary complexity.

internal/levee.go
mlevesquedion: I believe you make valid points. Going forward I will use `Referrers/Operands`.

PurelyApplied:
[Continuing offline conversation]

Were you intending to close this PR?  I see you've subsumed parts of it in new PRs, as well.

PR #38, opened 2020-08-03
Handle slices

mlevesquedion:
This PR adds support for propagating taint to slices in order to report "a source has reached a sink" when a slice containing a tainted value gets sinked.

In order to propagate the taint to slices, we need to DFS into the operands of nodes.

I have also modified the source analyzer to not visit sink functions. This was required to prevent a spurious diagnostic in the "domination" test.

- [x] Tests pass
- [x] Appropriate changes to README are included in PR (N/A)

### DIFF
--- a/internal/pkg/config/config.go
+++ b/internal/pkg/config/config.go
@@ -84,6 +84,16 @@ func (c Config) IsSink(call *ssa.Call) bool {
 	return false
 }
 
+func (c Config) IsSinkFunction(f *ssa.Function) bool {
+	for _, p := range c.Sinks {
+		if p.MethodRE.MatchString(f.Name()) {
+			return true
+		}
+	}
+
+	return false
+}
+
 func (c Config) IsSanitizer(call *ssa.Call) bool {
 	for _, p := range c.Sanitizers {
 		if p.Match(call) {
--- a/internal/pkg/source/source.go
+++ b/internal/pkg/source/source.go
@@ -31,6 +31,7 @@ type classifier interface {
 	IsSanitizer(*ssa.Call) bool
 	IsPropagator(*ssa.Call) bool
 	IsSourceFieldAddr(*ssa.FieldAddr) bool
+	IsSinkFunction(fn *ssa.Function) bool
 }
 
 // Source represents a Source in an SSA call tree.
@@ -68,15 +69,22 @@ func (a *Source) dfs(n ssa.Node) {
 	a.preOrder = append(a.preOrder, n)
 	a.marked[n.(ssa.Node)] = true
 
-	if n.Referrers() == nil {
-		return
+	if n.Referrers() != nil {
+		a.visitReferrers(n.Referrers())
 	}
 
-	for _, r := range *n.Referrers() {
+	var operands []*ssa.Value
+	operands = n.Operands(operands)
+	if operands != nil {
+		a.visitOperands(operands)
+	}
+}
+
+func (a *Source) visitReferrers(referrers *[]ssa.Instruction) {
+	for _, r := range *referrers {
 		if a.marked[r.(ssa.Node)] {
 			continue
 		}
-
 		switch v := r.(type) {
 		case *ssa.Call:
 			// This is to avoid attaching calls where the source is the receiver, ex:
@@ -84,7 +92,6 @@ func (a *Source) dfs(n ssa.Node) {
 			if v.Call.Signature().Recv() != nil {
 				continue
 			}
-
 			if a.config.IsSanitizer(v) {
 				a.sanitizers = append(a.sanitizers, &sanitizer.Sanitizer{Call: v})
 			}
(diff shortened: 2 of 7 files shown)


internal/pkg/config/config.go
```diff
@@ -84,6 +84,16 @@ func (c Config) IsSink(call *ssa.Call) bool {
 	return false
 }
 
+func (c Config) IsSinkFunction(f *ssa.Function) bool {
```
vinayakankugoyal, line 87: go/gocomments#doc-comments

PurelyApplied: go/ links are a feature of your local DNS.  They are opaque to a wider audience.

> All top-level, exported names should have doc comments, as should unexported type or function declarations with unobvious behavior or meaning. See http://golang.org/doc/effective_go.html#commentary for more information about commentary conventions.

internal/testdata/test-config.json
```diff
@@ -9,7 +9,7 @@
   "Sinks": [
     {
       "PackageRE": "^example.com/core$",
-      "MethodRE": "Sink"
+      "MethodRE": "Sinkf?$"
```
PurelyApplied, line 12: Why this change?  Reverting this fails no tests.

internal/pkg/source/source.go
```diff
@@ -68,23 +69,29 @@ func (a *Source) dfs(n ssa.Node) {
...
+func (a *Source) visitReferrers(referrers *[]ssa.Instruction) {
+	for _, r := range *referrers {
 		if a.marked[r.(ssa.Node)] {
 			continue
 		}
-
 		switch v := r.(type) {
 		case *ssa.Call:
 			// This is to avoid attaching calls where the source is the receiver, ex:
 			// core.Sinkf("Source id: %v", wrapper.Source.GetID())
```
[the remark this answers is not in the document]
mlevesquedion: Judging by the fact that all the tests keep passing if we remove it, it looks like we don't. :thinking: I'll add it to my personal list of things to investigate.

internal/pkg/source/source.go
```diff
@@ -144,6 +160,11 @@ func identify(conf classifier, ssaInput *buildssa.SSA) map[*ssa.Function][]*Sour
 	sourceMap := make(map[*ssa.Function][]*Source)
 
 	for _, fn := range ssaInput.SrcFuncs {
+		// no need to analyze the body of sinks
```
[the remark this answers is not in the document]
mlevesquedion: Gotcha. So in this specific case, ideally I would have made this change in a separate PR and wrote the slice PR separately, with a reference to the first one?

internal/testdata/test-config.json
mlevesquedion: Hmm. On my machine it breaks a bunch of tests. Specifically, `core.Sinkf` doesn't get identified as a Sink, so it doesn't get reported. Not sure why these were passing before.

internal/pkg/config/config.go
mlevesquedion: One thing I've been wondering about: do we still want a comment even if it would be redundant? e.g. here the comment would be `IsSinkFunction determines whether a Function is a Sink`...

internal/testdata/test-config.json
PurelyApplied: As far as regexp goes, `Sink` should match `Sinkf` without the `$` anchor...

PurelyApplied:
Please don't force pushes onto an open PR.  The various `View Changes` links in comment threads are now broken.

internal/pkg/config/config.go
PurelyApplied: I'm not dogmatic about it, but it is in the `Effective Go` standard and linters often expect a doc-comment.

PurelyApplied: From the `#commentary` link above:

> Inside a package, any comment immediately preceding a top-level declaration serves as a doc comment for that declaration. Every exported (capitalized) name in a program should have a doc comment.

mlevesquedion:
Gotcha.

(It seemed like the simpler path in the moment, I didn't know it would break those links.)

internal/pkg/source/source.go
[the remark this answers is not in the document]
PurelyApplied: > So in this specific case, ideally I would have made this change in a separate PR and wrote the slice PR separately, with a reference to the first one?

Is this change necessary for slice identification?  If it is, then yes, a reference to the required PR is good.  If not, then let the orthogonal branches be separate.

internal/testdata/test-config.json
mlevesquedion: My bad, my explanation was wrong. The problem is that because "Sink" functions are no longer analyzed, the analyzer doesn't look at the bodies of the test functions under `sinks/tests.go` (which have `Sink` in their names) so it misses all the reports. Making the regex more precise allows us to analyze these functions and also catch the sinks.

Therefore the change to the regex is actually to prevent false *negatives*, not false *positives* as I originally mentioned. You are correct w.r.t. regex matching.

internal/pkg/config/config.go
mlevesquedion: SG. It's not a big deal either way. In this case, besides the redundancy issue I also noticed that most (none?) of the other exported symbols had docs, so I went with the locally observed "standard".

internal/testdata/test-config.json
PurelyApplied: Ahh, against the `TestSinkEtCetera`.  Gotcha.

PR #40, opened 2020-08-05
Integrate field propagators analyzer with levee

mlevesquedion:
Depends on #39.

- [x] Tests pass


mlevesquedion:
It looks like I can't set my other PR (#39) as a reference instead of `google/master`. The first 2 commits don't belong in this PR.

PurelyApplied:
You would still want to set the PR against `google/master` in any case.  Just a comment that it's based on an existing PR like you have is fine.

mlevesquedion:
Since the first PR has been merged, the diff is now valid.

internal/levee.go
```diff
@@ -118,3 +123,16 @@ func getArgumentPropagator(c *config.Config, call *ssa.Call) ssa.Node {
 
 	return nil
 }
+
+func isFieldPropagator(c *ssa.Call, fieldPropagators map[types.Object]bool) bool {
```
immutableT, line 127: WDYT about placing this logic into the `fieldpropagator` result?
I am looking for a better encapsulation of the logic of identifying propagators.

mlevesquedion: 100% agree.

internal/levee.go
```diff
@@ -118,3 +123,16 @@ func getArgumentPropagator(c *config.Config, call *ssa.Call) ssa.Node {
...
+	for o, _ := range fieldPropagators {
```
PurelyApplied, line 128: Aside: if you only need the keys from a map, you can do `for key := range myMap`.

internal/levee.go
```diff
@@ -118,3 +123,16 @@ func getArgumentPropagator(c *config.Config, call *ssa.Call) ssa.Node {
...
+		cf, ok := c.Call.Value.(*ssa.Function)
+		if !ok {
+			continue
+		}
+		if cf.Object() == o {
+			return true
+		}
```
PurelyApplied, line 135: This doesn't need to loop over the map.  Go maps return the zero-value when accessing a key outside the map, so you could skip the loop with

```golang
func isFieldPropagator(c *ssa.Call, fieldPropagators map[types.Object]bool) bool {
  cf, ok := c.Call.Value.(*ssa.Function)
  return ok && fieldPropagators[cf.Object()]
}

If `cf.Object()` isn't in the map, the map returns the `bool` zero-value, which is `false`.
```

internal/levee.go
PurelyApplied, line 127: This being here feels awkward to me...

My first thought was that this function should belong to `fieldpropagator/analyzer.go`.  It's unpacking the result of that analysis, so it belongs to that package.

My second thought was that `fieldpropagator.ReturnType` should be a named type instead of an alias, and that this function should be a method on the `ReturnType`.  It's unpacking the result contained in that type, so it belongs to that type.

My third thought was that this is probably fine and we could iterate on it in the future if need be.

I don't know.  Weigh it yourself and I'll trust your judgement on it.

internal/levee.go
mlevesquedion: Thank you, I keep forgetting that :blush:

internal/levee.go
mlevesquedion: Very nice! Thanks, reads much better.

internal/levee.go
PurelyApplied: [Tone: playful ribbing of a `vim` user]
<cough> an IDE would highlight it <cough>.

mlevesquedion: Haha, `vim-go` actually nits me on it too, but it's not super visible :/

internal/levee.go
mlevesquedion: Good thinking. I think now is a good time to change it, I'll go with what you suggest in your 2nd thought.

PR #41, opened 2020-08-06
Add tests for propagation through calls

mlevesquedion:
- [x] Tests pass

### DIFF
--- a/internal/testdata/src/example.com/tests/propagation/tests.go
+++ b/internal/testdata/src/example.com/tests/propagation/tests.go
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package sinks
+package propagation
 
 import (
 	"fmt"
@@ -29,6 +29,10 @@ func TestIdentityPropagator(s core.Source) {
 	core.Sink(i) // want "a source has reached a sink"
 }
 
+func ReturnFive(arg interface{}) interface{} {
+	return 5
+}
+
 func ToString(arg interface{}) string {
 	return fmt.Sprintf("%v", arg)
 }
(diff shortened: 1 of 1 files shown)


internal/testdata/src/example.com/tests/propagation/tests.go
```diff
@@ -0,0 +1,39 @@
...
+	"fmt"
+
+	"example.com/core"
+)
+
+func Identity(arg interface{}) interface{} {
+	return arg
+}
+
+func TestIdentityPropagator(s core.Source) {
```
PurelyApplied, line 27: It's undoubtedly redundant with other tests, but consider adding a chained test here:

```golang
core.Sink(Identity(s)) // want "et cetera"
```

internal/testdata/src/example.com/tests/propagation/tests.go
```diff
@@ -0,0 +1,39 @@
...
+	i := Identity(s)
+	core.Sink(i) // want "a source has reached a sink"
+}
+
+func ToString(arg interface{}) string {
+	return fmt.Sprintf("%v", arg)
+}
+
+func TestToStringPropagator(s core.Source) {
+	v := ToString(s)
```
PurelyApplied, line 37: Consider adding a `fnt.Sprintf` reference directly:

```golang
st := fmt.Sprintf("%v", s)
core.Sink(st)

core.Sink(fmt.Sprintf("yet another %v", s)
```

mlevesquedion: Those tests actually fail.

I will keep those tests for a future PR.

In a future PR, I will also tackle false positives such as the `return 5` case we discussed elsewhere.

PR #44, opened 2020-08-11
Expose levee externally

mlevesquedion:
This PR exposes levee externally by creating an external `pkg/levee` package which exposes the internal `levee.Analyzer`.

In order to make the external pkg structure mirror the internal pkg structure, this PR *also* moves `levee` to its own package.

- [x] Tests pass
- [ ] Appropriate changes to README are included in PR (N/A?)


pkg/levee/levee.go
```diff
@@ -0,0 +1,7 @@
+// Package levee exports the levee Analyzer.
+package levee
+
+import internal "github.com/google/go-flow-levee/internal/pkg/levee"
+
+// Analyzer reports attempts to source data to sinks.
```
PurelyApplied, line 6: Let's avoid using `source` as a verb, so as to avoid the heavy usage of it as a noun throughout the codebase.  Consider instead:
```golang
/// Analyzer reports instances of source data reaching a sink.
```

mlevesquedion:
Thank you for the review!

Sorry about the description. In my mind while I was making the changes to the code, moving levee to its own pkg and exposing the analyzer registered as one thing so I didn't think to explain it in the description.

PR #46, opened 2020-08-13
Add tests for taint propagation through maps

mlevesquedion:
Add tests showing how propagation through maps is handled.

Currently, the entire map is considered to be tainted. This is not unreasonable considering that if a map containing a tainted value reaches a sink, the tainted value will almost certainly "reach the sink" as well, since e.g. printing a map implies printing all of its keys and values.

This does have the unfortunate side effect that any key access on the map will yield a value that is _considered_ to be tainted. For example, analyzing the following instructions yields a diagnostic, even though the value that reaches the sink is not _really_ tainted:
```
m := map[string]interface{}{"source": s}
core.Sink(m[""]) // diagnostic: "a source has reached a sink"
```

This sequence of instructions, in which a map is properly "sanitized" (ahem...) also yields a diagnostic:
```
m := map[string]interface{}{"source": s}
delete(m, "source")
core.Sink(m) // diagnostic: "a source has reached a sink"
```

- [x] Tests pass
- [x] Appropriate changes to README are included in PR. (N/A)


internal/pkg/levee/testdata/src/example.com/tests/collections/maps.go
```diff
@@ -0,0 +1,56 @@
...
+	m = nil
+	core.Sink(m)
+}
+
+func TestValueObtainedFromTaintedMapIsTainted(s core.Source) {
+	m := map[interface{}]string{s: "source"}
+	v := m[0]
+	core.Sink(v) // want "a source has reached a sink"
+	delete(m, s)
+	core.Sink(m) // want "a source has reached a sink"
```
immutableT, line 49: Should we be commenting on these type of tests with TODOs to fix such false positives?

mlevesquedion: I agree. I feel like filing an issue would be a better way to track this though. WDYT?

mlevesquedion: (GitHub marked this as outdated because I removed the 2 lines which overlap with the next test.)

PurelyApplied: I'd +1 to opening an Issue about it.  I've been meaning to backfill some known issues into GitHub, too.  Anything that makes the project more accessible to someone wanting to join the effort here is a good thing in my book.

I think identifying "sectional" taint in a collection is going to be A Whole Big Thing.  My initial impression is that it will only be detectable if all assignments and accesses are done with `const` values.  Any other access must be presumed tainted.

`delete(m, s)` as sanitizer is an interesting case here.  Is it possible currently to specify a built-in as a sanitizer?  Does it meaningfully change if the source were a value instead of a key?

mlevesquedion: Agreed on the Issues thing. Issue here: #47

Agreed on the Whole Big Thing. Probably we will need some kind of abstraction over collections that would allow us to say "this exact key (map key or slice/array index) is tainted/no longer tainted".

I originally though so, because the call to the builtin appears as a regular `ssa.Call`. Here's the ssa:
```
0: entry
	1(*ssa.Store          ): *t0 = s
	2(*ssa.MakeMap        ): t1 = make map[interface{}]string 1:int
	3(*ssa.UnOp           ): t2 = *t0
	4(*ssa.MakeInterface  ): t3 = make interface{} <- example.com/core.Source (t2)
	5(*ssa.MapUpdate      ): t1[t3] = "source":string
	6(*ssa.UnOp           ): t4 = *t0
	7(*ssa.MakeInterface  ): t5 = make interface{} <- example.com/core.Source (t4)
	8(*ssa.Call           ): t6 = delete(t1, t5)
	9(*ssa.Alloc          ): t7 = new [1]interface{} (varargs)
	10(*ssa.IndexAddr      ): t8 = &t7[0:int]
	11(*ssa.MakeInterface  ): t9 = make interface{} <- map[interface{}]string (t1)
	12(*ssa.Store          ): *t8 = t9
	13(*ssa.Slice          ): t10 = slice t7[:]
	14(*ssa.Call           ): t11 = example.com/core.Sink(t10...)
	15(*ssa.Return         ): return
```
However, when I looked deeper into it we can't actually do it right now because calls to builtins have a nil `Call.StaticCallee()`, so we would need to change the way `callMatcher.Match` works.
 
In any case, I don't think a user would actually want to list `delete` as a sanitizer because then any deletion would sanitize the whole map, meaning we could stop tracking a value that is still tainted. I think handling this properly would require the Whole Big Thing alluded to above.

I don't think having the source as a value changes anything. :thinking:

PR #48, opened 2020-08-17
Simplify propagation code

mlevesquedion:
## Context

The [reorganize-analysis](https://github.com/google/go-flow-levee/pull/37) PR I opened recently showed, albeit rather clumsily, that we did not actually need the `call` or `varargs` packages. In fact, currently we do not need an explicit representation of the concept of a `propagator` either. This is due to the fact that by traversing the SSA graph, we are already passing through slices and calls, which allows us to propagate taint without additional work.

## This PR

The goal of the current PR is to simplify the code we have for doing propagation. The main change that I made is in `source.go`. (Note that I refactored this code for robustness and clarity, so you won't find the following snippet in the actual code). Here it is:

*Before*
```go
al, ok := (*o).(*ssa.Alloc)
if !ok || al.Comment == "slicelit" || a.config.IsSource(al.Type()) {
  a.dfs(n)
```

*After*
```go
if ... || al.Comment == "varargs" || a.config.IsSource(utils.Dereference(al.Type())) {
```

There are 2 small changes here:
1. Checking the `Comment` for "varargs", not just "slicelit".
2. Dereferencing the `Alloc`'s type before checking whether it is a Source.

Checking the comment on the `Alloc` is brittle, which is why I refactored the code. (Again, you won't find this exact code in the PR, but it is presented here to illustrate the essence of the change).

Every other change in this PR is a natural consequence of the small change described above. Mostly, it leads to the deletion of a lot of code, which I think greatly simplifies the overall logic.

- [x] Tests pass


mlevesquedion:
I might have gotten carried away with the deleting.

However, most of the deletions in this PR are the direct consequence of removing the reliance on the `call` and `varargs` packages, so this shouldn't be a 5 line PR. That would just set us up for more forgotten dead code. I also think it is a significant improvement to the propagation logic, because it removes a lot of unnecessary complexity.

I'll wait for your PR to be merged and then I'll merge master into this one to remove the extraneous deletions from the diff.

internal/pkg/source/source.go
```diff
@@ -109,13 +109,23 @@ func (a *Source) visitOperands(operands []*ssa.Value) {
 		if !ok || a.marked[n] {
 			continue
 		}
-		al, ok := (*o).(*ssa.Alloc)
-		if !ok || al.Comment == "slicelit" || a.config.IsSource(al.Type()) {
+		al, isAlloc := (*o).(*ssa.Alloc)
```
immutableT, line 112: Please add some comments to explain the logic.

mlevesquedion: Certainly, done. I think a comment is super relevant here. When adding the comment, I realized I could also refactor the code to be more in line with the "line of sight" idiom, and I think it reads much better now. Thanks!

internal/pkg/source/source.go
```diff
@@ -109,13 +109,25 @@ func (a *Source) visitOperands(operands []*ssa.Value) {
 		if !ok || a.marked[n] {
 			continue
 		}
-		al, ok := (*o).(*ssa.Alloc)
-		if !ok || al.Comment == "slicelit" || a.config.IsSource(al.Type()) {
-			a.dfs(n)
+		al, isAlloc := (*o).(*ssa.Alloc)
+		// An Alloc represents the allocation of space for a variable. If a Node is an Alloc,
+		// and the thing being allocated is not a slice, then either:
+		// a) it is a Source value, in which case it will get its own DFS
```
immutableT, line 115: Could you please explain how it will get its DFS?

internal/pkg/source/source.go
```diff
@@ -109,13 +109,25 @@ func (a *Source) visitOperands(operands []*ssa.Value) {
...
+		// b) it is not a Source value, in which case we should not DFS through it.
+		// However, if the Alloc is a slice, then that means the source that we are DFSing from
```
immutableT, line 117: Could we be certain that the `Alloc` is for a `Source`?

internal/pkg/source/source.go
mlevesquedion: Do you mean explain it in the comment or explain it here?

If the latter, it will get its own DFS when the code in `source.go:sourcesFromBlocks` catches it. The relevant line is line 239 in `source/source.go`.

internal/pkg/source/source.go
mlevesquedion: I'm not sure I understand your question.

We could check its type, yes. But if it is a `Source` it will get its own DFS. If it isn't a `Source`, then we shouldn't DFS through it, because that would mean we are tainting a new variable that isn't a `Source`.

internal/pkg/source/source.go
```diff
@@ -109,13 +109,25 @@ func (a *Source) visitOperands(operands []*ssa.Value) {
...
+		// is being placed into that slice, so we do need to keep DFSing.
+		if isAlloc && !isSlice(utils.Dereference(al.Type())) {
+			return
 		}
+		a.dfs(n)
 	}
 }
 
+func isSlice(t types.Type) bool {
+	_, ok := t.(*types.Array)
```
PurelyApplied, line 127: Is slice checks if the type is an Array?  Not `types.Slice`?

internal/pkg/source/source.go
```diff
@@ -109,13 +109,25 @@ func (a *Source) visitOperands(operands []*ssa.Value) {
...
-		if !ok || al.Comment == "slicelit" || a.config.IsSource(al.Type()) {
-			a.dfs(n)
+		al, isAlloc := (*o).(*ssa.Alloc)
+		// An Alloc represents the allocation of space for a variable. If a Node is an Alloc,
+		// and the thing being allocated is not a slice, then either:
+		// a) it is a Source value, in which case it will get its own DFS
+		// b) it is not a Source value, in which case we should not DFS through it.
+		// However, if the Alloc is a slice, then that means the source that we are DFSing from
+		// is being placed into that slice, so we do need to keep DFSing.
+		if isAlloc && !isSlice(utils.Dereference(al.Type())) {
```
PurelyApplied, line 119: My previous comment appears to have been marks out of date, but `isSlice` should be in-lined into this `if` statement.

internal/pkg/source/source.go
```diff
@@ -109,13 +109,25 @@ func (a *Source) visitOperands(operands []*ssa.Value) {
 		if !ok || a.marked[n] {
 			continue
 		}
-		al, ok := (*o).(*ssa.Alloc)
-		if !ok || al.Comment == "slicelit" || a.config.IsSource(al.Type()) {
-			a.dfs(n)
+		al, isAlloc := (*o).(*ssa.Alloc)
```
PurelyApplied, line 112: Nit: this line should go below the comment block.

internal/pkg/source/source.go
mlevesquedion: I think the ssa allocates a `types.Array` as the backing array for a slice. Since I've inlined the code `isSlice` became a variable. In retrospect in think `isArray` is less confusing. I've also updated the doc comment.

(The reason I called it `isSlice` initially is that in actual code usually the `Array` is intended to be used for a slice, but that is a poor justification.)

internal/pkg/source/source.go
mlevesquedion: Done.

internal/pkg/source/source.go
mlevesquedion: Done.

internal/pkg/source/source.go
immutableT: Thanks.
I think it would be useful to include this in the comment.

mlevesquedion: Gotcha. Done.

internal/pkg/source/source.go
```diff
@@ -109,13 +109,28 @@ func (a *Source) visitOperands(operands []*ssa.Value) {
...
-		al, ok := (*o).(*ssa.Alloc)
-		if !ok || al.Comment == "slicelit" || a.config.IsSource(al.Type()) {
-			a.dfs(n)
+		// An Alloc represents the allocation of space for a variable. If a Node is an Alloc,
+		// and the thing being allocated is not an array, then either:
+		// a) it is a Source value, in which case it will get its own DFS when sourcesFromBlocks
+		//    finds this Alloc
+		// b) it is not a Source value, in which case we should not DFS through it.
+		// However, if the Alloc is an array, then that means the source that we are DFSing from
+		// is being placed into an array, slice or varags, so we do need to keep DFSing.
```
PurelyApplied, line 118: Nit: I'd avoid an implementation-specific term like `DFSing` in favor of a more general `traversing`.  It wasn't long ago that we were using breadth-first.

internal/pkg/source/source.go
PurelyApplied: > Since I've inlined the code

That makes this dead, then.

internal/pkg/source/source.go
```diff
@@ -109,13 +109,28 @@ func (a *Source) visitOperands(operands []*ssa.Value) {
...
+		al, isAlloc := (*o).(*ssa.Alloc)
+		if isAlloc {
+			if _, isArray := utils.Dereference(al.Type()).(*types.Array); !isArray {
+				return
+			}
 		}
```
PurelyApplied, line 124: I don't know why I didn't notice it when the line was above the comment block and I suggested you move it down...

You can in-line the declaration on both of these checks.
```golang
if al, isAlloc := (*o).(*ssa.Alloc); isAlloc {
	if _, isArray := utils.Dereference(al.Type()).(*types.Array); !isArray {
		return
	}
}
```

And actually, since `Type()` is on the interface `o` satisfies, you could even go with 

```golang
if _, isAlloc := (*o).(*ssa.Alloc); isAlloc {
	if _, isArray := utils.Dereference((*o).Type()).(*types.Array); !isArray {
		return
	}
}
```

Although I think the first block reads better.

internal/pkg/source/source.go
mlevesquedion: Good nit. Done.

internal/pkg/source/source.go
mlevesquedion: Good stuff, thanks! Done.

I actually don't really like this style of `if` statement. It's good because it limits the scope of the variables, but I find it hard to read. Anyhoo, this seems to be universally accepted as idiomatic so I'll just get used to it.

Thanks for including the second block, very informative.

PR #49, opened 2020-08-17
Use fieldtags analyzer in fieldpropagator analyzer

mlevesquedion:
## This PR

This PR enhances the `fieldpropagator` analyzer's ability to detect field propagators by allowing it to use the results of the `fieldtags` analyzer. (The `fieldtags` analyzer was unused up to now.)

## Future PRs (in order)
1. Make this configurable, with the current setting of `levee:"source"` as a default.
2. Use the `fieldtags` analyzer in the `source` analyzer.

## Outstanding
* Should we mark a tagged field as a Source even if it is not on a Source type?
  - I would say no.
  - Currently, we do. This would be easy to change, however.
* Should we produce a warning when we detect a field tagged as a Source that is not on a Source type?
  - I would say yes, e.g. "warning: field tagged as source on non-source type; this field will not be used as a source"

- [x] Tests pass

### DIFF
--- a/internal/pkg/fieldpropagator/analyzer.go
+++ b/internal/pkg/fieldpropagator/analyzer.go
@@ -50,7 +50,7 @@ A field propagator is a function that returns a source field.`,
 	FactTypes:  []analysis.Fact{new(isFieldPropagator)},
 }
 
-type Analysis struct {
+type analysisState struct {
 	pass     *analysis.Pass
 	conf     *config.Config
 	ssaInput *buildssa.SSA
@@ -66,7 +66,7 @@ func run(pass *analysis.Pass) (interface{}, error) {
 		return nil, err
 	}
 
-	a := &Analysis{
+	a := &analysisState{
 		pass:     pass,
 		conf:     conf,
 		ssaInput: ssaInput,
@@ -81,7 +81,7 @@ func run(pass *analysis.Pass) (interface{}, error) {
 	return FieldPropagators(isFieldPropagator), nil
 }
 
-func (a *Analysis) run() {
+func (a *analysisState) run() {
 	ssaProg := a.ssaInput.Pkg.Prog
 	for _, mem := range a.ssaInput.Pkg.Members {
 		ssaType, ok := mem.(*ssa.Type)
@@ -96,7 +96,7 @@ func (a *Analysis) run() {
 	}
 }
 
-func (a *Analysis) analyzeBlocks(meth *ssa.Function) {
+func (a *analysisState) analyzeBlocks(meth *ssa.Function) {
 	// Function does not return anything
 	if res := meth.Signature.Results(); res == nil || (*res).Len() == 0 {
 		return
@@ -114,7 +114,7 @@ func (a *Analysis) analyzeBlocks(meth *ssa.Function) {
 	}
 }
 
-func (a *Analysis) analyzeResults(meth *ssa.Function, results []ssa.Value) {
+func (a *analysisState) analyzeResults(meth *ssa.Function, results []ssa.Value) {
 	for _, r := range results {
 		fa, ok := fieldAddr(r)
 		if !ok {
--- a/internal/pkg/fieldpropagator/analyzer.go
+++ b/internal/pkg/fieldpropagator/analyzer.go
@@ -50,13 +50,6 @@ A field propagator is a function that returns a source field.`,
 	FactTypes:  []analysis.Fact{new(isFieldPropagator)},
 }
 
-type analysisState struct {
-	pass     *analysis.Pass
-	conf     *config.Config
-	ssaInput *buildssa.SSA
-	tf       fieldtags.TaggedFields
-}
-
 func run(pass *analysis.Pass) (interface{}, error) {
 	taggedFields := pass.ResultOf[fieldtags.Analyzer].(fieldtags.ResultType)
 	ssaInput := pass.ResultOf[buildssa.Analyzer].(*buildssa.SSA)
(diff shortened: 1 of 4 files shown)


internal/pkg/fieldpropagator/analyzer.go
```diff
@@ -44,40 +45,58 @@ var Analyzer = &analysis.Analyzer{
...
 	ResultType: reflect.TypeOf(new(ResultType)).Elem(),
 	FactTypes:  []analysis.Fact{new(isFieldPropagator)},
 }
 
+type Analysis struct {
+	pass     *analysis.Pass
+	conf     *config.Config
+	ssaInput *buildssa.SSA
+	tf       fieldtags.TaggedFields
+}
```
mlevesquedion, line 58: WDYT about this? I find it makes the function signatures much lighter, and the code more readable.

The name might not be ideal though, I am open to suggestions.

internal/pkg/fieldpropagator/analyzer.go
```diff
@@ -44,40 +45,58 @@ var Analyzer = &analysis.Analyzer{
 A field propagator is a function that returns a source field.`,
 	Flags:      config.FlagSet,
 	Run:        run,
-	Requires:   []*analysis.Analyzer{buildssa.Analyzer},
+	Requires:   []*analysis.Analyzer{buildssa.Analyzer, fieldtags.Analyzer},
 	ResultType: reflect.TypeOf(new(ResultType)).Elem(),
 	FactTypes:  []analysis.Fact{new(isFieldPropagator)},
 }
 
+type Analysis struct {
```
mlevesquedion, line 53: WDYT about this? I find it makes the function signatures much lighter, and the code more readable.

The name might not be ideal though, I am open to suggestions.

PurelyApplied:
> Should we mark a tagged field as a Source even if it is not on a Source type?
I would take a in-code annotation as a higher truth than a runtime config.  Otherwise, a user would have to essentially configure analysis twice: once as an annotation and once in the config JSON.

More to the point, what is the benefit of using a tag if it must also be specified in a JSON, where fields can be explicitly configured in any case?

mlevesquedion:
Hmm. What if a user already has relevant annotations scattered all over their code base?

Wouldn't they rather add a couple patterns to the JSON config?

Also, users might want to avoid "polluting" their struct fields with tags intended specifically for our analysis. If they don't mind doing that, they can use our default. However, if they have their own convention (e.g. "`data:"secret"` or whatever) then they would need to duplicate that information everywhere by adding our `levee:"source"` annotation.

PurelyApplied:
> Hmm. What if a user already has relevant annotations scattered all over their code base?
> 
> Wouldn't they rather add a couple patterns to the JSON config?
> 
> Also, users might want to avoid "polluting" their struct fields with tags intended specifically for our analysis. If they don't mind doing that, they can use our default. However, if they have their own convention (e.g. `"data:"secret"` or whatever) then they would need to duplicate that information everywhere by adding our `levee:"source"` annotation.

I'm not sure I understand your premise here.

A user should be able to specify a source by type, configuring via package path, member name, and optional field name regexp.

A user should also be able to specify a field tag to automate that.  We're currently hard-coded at `levee:"source"`, but that will need to be configurable in the future.

A user should absolutely not have to do both.

mlevesquedion:
Oh my bad. In your original comment I thought you were also giving an opinion about whether this should be configurable or not. I agree with your point of view.

> We're currently hard-coded at levee:"source", but that will need to be configurable in the future.

Also agreed.

internal/pkg/fieldpropagator/analyzer.go
mlevesquedion: Hmm. I'm not sure I see any clear drawbacks here. The abstraction would be the state/functionality that is necessary to run the analysis itself. A 4-argument signature is never fun to see in any case.

It is indeed not meant to be exported, good catch.

I don't see how this hurts the reviewability of this PR, and I don't think this change would hold its own in a future PR. I made this change because I thought it was justified by the changes in the PR, i.e. it got to a point where we had a bunch of functions with large signatures, and I was looking for a way to make a clear distinction between "these variables are specifically important to this function" and "these variables are generally important to the analysis as a whole".

PurelyApplied: > A 4-argument signature is never fun to see in any case.
> ...
> I was looking for a way to make a clear distinction between "these variables are specifically important to this function" and "these variables are generally important to the analysis as a whole".

This isn't Java.  Go is a functional language, not an object-oriented one.  Structs are abstracted data bundles, not namespaces.  There's no benefit in calling a niladic function over a four-parameter one.  A function should have as many arguments as it needs and not rely on a struct to globalize parameters.

> I don't see how this hurts the reviewability of this PR

As your reviewer, I am letting you know that it has slowed my review.

> I don't think this change would hold its own in a future PR.

There's no `You must be this tall to open a PR` here.  If it is a meaningful, atomic improvement, it could constitute a quick-and-easy PR.  If it isn't, it doesn't belong here either.

mlevesquedion: > This isn't Java. 

I don't see the point in comparing Go to other languages, or in discussing programming paradigms. My understanding of your point of view is that the `struct` keyword should never be used unless it is defining a meaningful abstraction. Is that correct?

> There's no benefit in calling a niladic function over a four-parameter one.

Let's agree to disagree then. I think a function is much easier to read when it has fewer parameters.

> A function should have as many arguments as it needs ...

So in your opinion a function with, say, 8 arguments would not be a problem?

> As your reviewer, I am letting you know that it has slowed my review.

Gotcha. I will try to avoid creating unnecessary structs. 

> There's no ...

Ok, that is good to know. To reiterate my initial stance, I thought the change was warranted as a natural extension of the changes in this PR. It certainly wasn't *required* by the changes though, so I guess it could indeed have been its own PR.

mlevesquedion:
I've reverted the `struct`. Looking back, I don't think it makes a huge difference, so if it can help with the review, that's a clear win.

PR #50, opened 2020-08-17
Introduce pointer analysis

mlevesquedion:
## Context

This PR introduces pointer analysis into `levee` in order to handle some tricky cases that we have been having trouble with. In particular, the following case, and some other cases similar to it, are now handled correctly:

*Wrapper through empty interface*
```
func TestSinkWrapper(s core.Source) {
    SinkWrapper(s) // want "a source has reached a sink"
}

func SinkWrapper(arg interface{}) {
    core.Sink(arg) 
}
```
Note that the diagnostic is produced at the call to `SinkWrapper`.

This other case is still not handled properly, however, because pointer analysis can't work with `int`s.
*Untainted value identified as tainted*
```
func TestReturnsFive(s core.Source) {
    five := ReturnsFive(s)
    core.Sink(five) // an unexpected diagnostic is produced here
}

func ReturnsFive(arg interface{}) int {
    return 5
}
```

## Implementation
In order to make the pointer analysis play well with the existing analysis, pointer analysis is executed first. If pointer analysis reports something at a call, there is no need to analyze it with our existing analysis. Running both analyses also allows us to handle some cases which pointer analysis does not handle, such as this one:
*Source value written to a string*
```
func TestStringify(s core.Source) {
    str := Stringify(s)
    core.Sink(str) // pointer analysis cannot handle this case, because a string is not a reference type
}

func Stringify(arg interface{}) string {
    return fmt.Sprintf("%v", arg)
}
```
Performing the `pointer` analysis itself is fairly straightforward. For each sink call, we collect the `ssa.Value`s that are arguments to the call. If the sink is variadic, this requires a bit of additional work. For each value, if it is queryable, we add it to the list of queries. After performing the analysis, we go over the queries' `PointsToSet`s and report our findings.

## Outstanding issues
At present, pointer analysis does not allow us to handle every case we might be interested in. For example, the following test case fails:
```
func TestSinkWrapperSlice(s core.Source) {
	SinkWrapperSlice("not a source", s, 0) // TODO want "a source has reached a sink"
}

func SinkWrapperSlice(args ...interface{}) {
	core.Sink(args)
}
```
My understanding of this failure is that within `SinkWrapperSlice`, the `args` parameter is represented as an `ssa.Parameter`, which represents the whole slice. Since the only `Referrer` for this `ssa.Parameter` is the call to `core.Sink`, and since it has no `Operands` (because it is an `ssa.Value`), using the current approach I see no way to formulate a relevant query for pointer analysis. We are not interested in the whole slice, merely in its elements (more specifically, in its 2nd element, but we can't know that in advance), and we have no way to get at these elements (at least, not within the body of `SinkWrapperSlice`).

- [x] Tests pass


internal/pkg/levee/levee.go
```diff
@@ -94,20 +118,77 @@ func run(pass *analysis.Pass) (interface{}, error) {
...
+		pointerConf.AddQuery(v)
+	}
+
+	result, err := pointer.Analyze(pointerConf)
+	if err != nil {
+		return false
+	}
+
+	isSafeCall = true
+	for _, pSet := range result.Queries {
```
immutableT, line 150: Since documentation here https://godoc.org/golang.org/x/tools/go/pointer is very terse, do you mind putting a few comments that explain the flow?

mlevesquedion:
> I think invoking `doPointerAnalysis` within `run` is problematic. [...]
I'm not sure I understand what you mean by "analysis should be ordered according to imports"? Do you mean pointer analysis might run incorrectly if it is not run in topological import order?

> Notably, the nearest similar test using `ToString` has been removed with no replacement.
I was on the fence about including that test. I decided to remove it because it was already passing on master, and I thought we had a similar test elsewhere. However, revisiting that test case allowed me to realize that the logic to determine whether a call is `safe` is backwards. In fact, the concept of a `safe` call is misleading. I meant for it to represent the fact that a call was already successfully analyzed with pointer analysis, so we don't need to run the usual analysis on it. I made some changes, and the `ToString` case is handled correctly now.

Unfortunately, this does mean that the `ReturnsFive` case no longer works, because pointer analysis can't handle it, and our existing analysis incorrectly produces a diagnostic for it.

> To remediate these issues, I think a better approach would be to [...]
I don't think going up one level to the parent would allow us to scale this. We might need to go up several levels, which is what we want the pointer analysis to do for us. I'm also not sure how we could mark arguments as "possibly sources", or more importantly, how we could use "possibly sources" in propagation analysis. It sounds like that could get us a lot of false positives.

internal/pkg/levee/levee.go
```diff
@@ -122,3 +203,25 @@ func getArgumentPropagator(c *config.Config, call *ssa.Call) ssa.Node {
...
+func callValues(c *ssa.Call) (values []ssa.Value) {
+	for _, a := range c.Call.Args {
+		slice, ok := a.(*ssa.Slice)
+		if !ok {
+			values = append(values, a)
+			continue
+		}
+		refs := *slice.X.Referrers()
+		// the last referrer is the slice
+		for i := 0; i < len(refs)-1; i++ {
```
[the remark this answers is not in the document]
mlevesquedion: I could add an explicit check for it to make sure we avoid panics completely, but it is my understanding that getting an empty slice is impossible. Since the `slice` is an `Arg` of the `Call`, then it will at least be referred to by the `Call`.

internal/pkg/levee/levee.go
```diff
@@ -122,3 +203,25 @@ func getArgumentPropagator(c *config.Config, call *ssa.Call) ssa.Node {
...
+// We don't want to analyze what the variadic argument points to (a slice), but we do
+// want to analyze the values *within* the slice.
+func callValues(c *ssa.Call) (values []ssa.Value) {
+	for _, a := range c.Call.Args {
+		slice, ok := a.(*ssa.Slice)
+		if !ok {
+			values = append(values, a)
+			continue
+		}
+		refs := *slice.X.Referrers()
```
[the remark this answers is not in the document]
mlevesquedion: Good question. I am removing the `varags` package entirely in PR #48. Given that, I think it would be preferable not to, and I don't think this code deserves its own package.

internal/pkg/levee/levee.go
mlevesquedion: Sure. Done, let me know if it's better now.

internal/pkg/levee/levee.go
```diff
@@ -122,3 +203,25 @@ func getArgumentPropagator(c *config.Config, call *ssa.Call) ssa.Node {
 
 	return nil
 }
+
+// callValues collects the ssa.Values that are arguments to an ssa.Call.
+// in particular, it handles variadic arguments to some extent.
+// We don't want to analyze what the variadic argument points to (a slice), but we do
+// want to analyze the values *within* the slice.
+func callValues(c *ssa.Call) (values []ssa.Value) {
```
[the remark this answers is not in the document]
mlevesquedion: I think this logic should remain close to other pointer-analysis-related functions, since it is very closely related to it and isn't likely to be used anywhere else.

In fact I think we may want to rename the `utils` package to something more evocative, or even remove it completely. Inspired by [this post](https://dave.cheney.net/2019/01/08/avoid-package-names-like-base-util-or-common).

internal/pkg/levee/levee.go
```diff
@@ -94,20 +118,77 @@ func run(pass *analysis.Pass) (interface{}, error) {
 	return nil, nil
 }
 
+func doPointerAnalysis(pass *analysis.Pass, analysisConf *config.Config, c *ssa.Call) (isSafeCall bool) {
+	pkg := c.Parent().Pkg
```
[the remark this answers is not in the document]
mlevesquedion: Very nice, thanks for the link!

At this point I think I would need to add more tests before I change this logic though. Also, I'm not sure in what situation we would want to have more than one main package? (this is likely just a failure of my imagination)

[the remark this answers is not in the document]
immutableT: k8s (as a product) consists of dozens of components (main functions): kube-apiserver, scheduler, controller-manager, etc. 
So, at least, in k8s case we will have multiple mains.

internal/pkg/levee/levee.go
```diff
@@ -94,20 +118,95 @@ func run(pass *analysis.Pass) (interface{}, error) {
...
+		return false
+	}
+
+	for _, v := range vs {
+		// If we ask the pointer package to analyze a value it can't analyze (i.e., not a pointer or
+		// pointer-like type), the analysis will fail and produce error output in stderr.
+		if !pointer.CanPoint(v.Type()) {
+			continue
+		}
+		// Let the configuration know that we are interested in computing this value's points-to set.
```
immutableT, line 151: Is `points-to set` is a set of all call-sites which may contribute to the `v` under analysis?

mlevesquedion: It is the set of values that may be pointed to by the given pointer value. I've added that to the comment.

internal/pkg/levee/testdata/src/example.com/tests/propagation/tests.go
```diff
@@ -20,21 +20,77 @@ import (
 	"example.com/core"
 )
 
-func Identity(arg interface{}) interface{} {
-	return arg
+func main() {
```
immutableT, line 23: Do you think these new tests are related to propagation?

I think we are testing something else here, but I can't quite sure how to name this. WDYT?

internal/pkg/levee/testdata/src/example.com/tests/propagation/tests.go
```diff
@@ -20,21 +20,77 @@ import (
...
+	TestSinkWrapperSlice(core.Source{})
+	TestSinkWrapperSpread(core.Source{})
+	TestStringify(core.Source{})
 }
 
-func TestIdentityPropagator(s core.Source) {
-	i := Identity(s)
-	core.Sink(i)           // want "a source has reached a sink"
-	core.Sink(Identity(s)) // want "a source has reached a sink"
+func TestSinkWrapperWrapper(s core.Source) {
```
immutableT, line 33: How would this play with sanitizers?
Could we add a test? Even if it is a failing test.

mlevesquedion: It would fail to recognize that the pointed-to value has been sanitized and produce a report. :/

internal/pkg/levee/levee.go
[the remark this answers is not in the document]
immutableT: It think that the logic of finding main functions is reasonably straightforward to test.

internal/pkg/levee/levee.go
```diff
@@ -79,16 +102,114 @@ func run(pass *analysis.Pass) (interface{}, error) {
...
+			v := lab.Value()
+			switch t := v.(type) {
+			case *ssa.MakeInterface:
+				v = t.X
+			}
+			if !analysisConf.IsSource(utils.Dereference(v.Type())) {
+				continue
+			}
+			analyzedSuccesfully = true
+			reportAtSource(pass, v.(ssa.Node), (ssa.Node)(c))
```
immutableT, line 168: Do we take into account sanitizers?

mlevesquedion: No, not at all. This is a very important point you have raised here. See my comment on your other review comment.

internal/pkg/levee/testdata/src/example.com/tests/propagation/tests.go
mlevesquedion: Good question. I think propagation may be too broad. Indeed, *taint propagation* is the whole point of our analysis. Maybe propagat***ors*** would be more appropriate?

internal/pkg/levee/testdata/src/example.com/tests/propagation/tests.go
immutableT: Do you think that asking the pointer analysis to also produce a CallGraph somehow could help us with this?

mlevesquedion: Hmm. That is a good question.

I don't think so. Given an `ssa.Value`, we can get the function it was defined in by using its `Parent()` function. By traversing the call graph we might be able to determine e.g. that the function that contains the source value calls a sanitizer at some point, but we would still need to look at the function's body to determine whether sanitization is done properly. So I don't think `pointer` analysis would help us much, we would need to do a lot of work outside of the analysis.

internal/pkg/levee/levee.go
immutableT, line 168: Perhaps, instead of reporting, we should emit a Fact, and later check for the presence of sanitizers?

mlevesquedion: We could, but I think the "checking for sanitizers" part would require a lot of work, and in fact would likely require us to implement things that we were hoping pointer analysis would do for us, and that if we do implement, would obviate the need to use pointer analysis. For example:
```
func CallSinkWrapperWrapper(s core.Source) {
    SinkWrapperWrapper(s)
}
func SinkWrapperWrapper(i interface{}) {
    safe := Sanitize(i)
    SinkWrapper(safe)
}
func SinkWrapper(i interface{}) {
    core.Sink(i)
}
func Sanitize(i interface{}) interface{} {
    return i
}
```
`pointer` analysis would tell us that the `i` argument to `core.Sink` could point to the source `s` defined in `CallSinkWrapperWrapper`. However, that value has been properly sanitized. But `pointer` analysis doesn't tell us anything about the path the value may have traveled, so we have no easy way to know that. Having the call graph would tell us where to look, and we might be able to reconstruct that path (there could easily be many paths, though), but then we would need to fully analyze the bodies of every function on the path from source to sink in order to piece together the path the value followed, and see that it was sanitized before reaching the sink.

immutableT: If there are no sanitizers on the CallGraph then we can safely report a finding, correct?
However, if there is one - we need to analyze bodies of the functions on the call-graph to ensure that the sink was dominated by a sanitizer, correct?

immutableT:
WDYT about the following approach:

1. Perform the per-function analysis in the same way we are doing this today
    - Report leaks identified within each function (same as we do today)
    - **New**: return to the parent analyzers a list that represents sanitized sources and sinks (including potential sinks). 
   -  New optimization: Remove sources from the above list if such sources are either sanitized prior to being passed as to other calls or never passed to other calls.
    
2. Perform pointer analysis (as implemented in this PR)
    - Trim each Points-To-Set by removing items which already have been reported (this knowledge would also need to be passed by the child analyzer).
    - For each remaining item in the Points-To-Set, identify the `Package`.`Function` and the argument that we now believe to be a Source (probably there is a better way to do this). 
    - Add such newly discovered source to the result of this pointer-analyzer. The result should uniquely identify a new source, probably package:function:argument-idx is sufficient.
3. Perform a version of our current per-function analyzer which uses a slightly different definition (location-based) of a Source.

internal/pkg/levee/levee.go
```diff
@@ -79,16 +102,114 @@ func run(pass *analysis.Pass) (interface{}, error) {
 	return nil, nil
 }
 
+func doPointerAnalysis(pass *analysis.Pass, analysisConf *config.Config, c *ssa.Call) (analyzedSuccesfully bool) {
+	// Pointer analysis requires at least one main package containing a main function.
+	// If we fail to provide such a package, the analysis will fail and produce error output in stderr.
+	pkg := c.Parent().Pkg
+	main := pkg.Func("main")
```
PurelyApplied, line 109: Mentioned offline / in my above top-level comment, but I'm concerned about how this interacts with the `analysis` partition into package `pass`es.  Add tests that cross package boundaries.

mlevesquedion:
I think that could work for some cases, but I'm worried the logic is going to get convoluted. I also don't think that this would allow us to handle the case that I presented in a previous comment, where a `Source` gets sanitized *somewhere* on a path from source to sink over many functions. Also, if a `Source` gets sanitized before every call in the body of the function that "owns" the `Source`, then yes, we would be able to avoid a false positive. But since `pointer` analysis does not tell us through what call the source reached the sink, if the `Source` reaches even *one* call without being sanitized, then we will have to assume that it was not sanitized.

There's also the issue of `ssa.Parameters`, which I don't currently see a way around. See the "outstanding issues" section in the PR description, as well as the comments in `levee/testdata/src/example.com/tests/propagation/tests.go`, lines 66 and 79. In short, in some (non-degenerate) cases it seems we will not be able to collect the arguments to a function call, and therefore not be able to perform pointer analysis on them.

internal/pkg/levee/levee.go
```diff
@@ -79,16 +102,114 @@ func run(pass *analysis.Pass) (interface{}, error) {
...
+	pass.Reportf(source.Pos(), b.String())
+}
+
+// callValues collects the ssa.Values that are arguments to an ssa.Call.
+// In particular, it handles variadic arguments to some extent.
+// We don't want to analyze what the variadic argument points to (a slice), but we do
+// want to analyze the values *within* the slice.
+func callValues(c *ssa.Call) (values []ssa.Value) {
+	for _, a := range c.Call.Args {
+		slice, ok := a.(*ssa.Slice)
```
PurelyApplied, line 201: Is the intent to unpack all slices, or only those for variadic calls?  This should do the former, correct?  That could be fine, but the function's docstring suggests the latter.

internal/pkg/levee/levee.go
```diff
@@ -79,16 +102,114 @@ func run(pass *analysis.Pass) (interface{}, error) {
...
+		if !ok {
+			values = append(values, a)
+			continue
+		}
+		refs := *slice.X.Referrers()
+		// the last referrer is the slice
```
PurelyApplied, line 207: This feels imprecise.  We have a `slice` above and are examining its referrers.  The last referrer of the slice is the slice?  And that's why we're only iterating to `len(refs)-1`?

I thought `Referrers()` order wasn't guaranteed.  A simple equality check and `continue` within the loop would be more explicit, and would resolve immutableT's concern about emptyness.

internal/pkg/levee/levee.go
```diff
@@ -18,22 +18,28 @@ import (
...
 	"github.com/google/go-flow-levee/internal/pkg/varargs"
 	"golang.org/x/tools/go/analysis"
+	"golang.org/x/tools/go/analysis/passes/buildssa"
+	"golang.org/x/tools/go/pointer"
 	"golang.org/x/tools/go/ssa"
 
 	"github.com/google/go-flow-levee/internal/pkg/source"
 )
 
+const sourceReachedSinkMessage = "a source has reached a sink"
```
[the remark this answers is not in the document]
mlevesquedion: I think you make a valid point. As I understand it, the only difference between the two kinds of report is the *location*, i.e. at the source or at the sink. In both cases we are reporting on the same kind of event, i.e. that "a source has reached a sink". This difference is already present in the messages: `, source: %v` vs `, sink: %v`, but that may not be obvious because the reporting functions' bodies are largely the same. Do you think a refactoring is needed?

I've updated the `want` messages in `propagators` to show what kind of message we expect.

internal/pkg/levee/levee.go
mlevesquedion: Correct on all counts. I've updated the docstring. The intent in the docstring should be more in line with the implementation now.

internal/pkg/levee/levee.go
mlevesquedion: > The last referrer of the slice is the slice? And that's why we're only iterating to len(refs)-1?

Correct.

> I thought Referrers() order wasn't guaranteed.

I couldn't find anything in the `ssa` package's docs that explicitly says that the order of `Referrers()` is not guaranteed. 

> A simple equality check and continue within the loop would be more explicit, and would resolve immutableT's concern about emptyness.

Agreed. Done.

internal/pkg/levee/levee.go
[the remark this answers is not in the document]
PurelyApplied: > This difference is already present in the messages: , source: %v vs , sink: %v, but that may not be obvious because the reporting functions' bodies are largely the same. Do you think a refactoring is needed?

It was more of an issue of testing readability than reporting.  Since `// want "text"` checks `text` as a regexp, and since our `// want` lines aren't anchored and we hadn't included the appended suffix.  Including it in the `// want` as you have suffices, though.

internal/pkg/levee/levee.go
mlevesquedion: Please see my comment on the PR itself for further discussion.

mlevesquedion:
Concerning cross-package analysis, I think the problem boils down to the way the `pointer` package performs its analysis. For pointer analysis to be able to give us the points-to set for an argument to a `sink` call, the pointer package needs to be provided with a `main` package that has a `main` function that has a call in it that eventually leads to a call to a `sink`. So it needs something like `package main ... func main() { sinkWrapper(core.Source{}) }`.

What that means is, if we have a function `sinkWrapper(a interface{}) { core.Sink(a) }` that is defined in a non-main package that doesn't have a main function (i.e., a library), then the `pointer` package does not provide us with a way to determine the points-to set for that argument. In the package's defense, this is a bit of a tall order, because it would require it to find the set of packages that import and use this function, which it currently has no way to do. 

Given these limitations, I will begin working on a different approach.

PurelyApplied:
> What that means is, if we have a function `sinkWrapper(a interface{}) { core.Sink(a) }` that is defined in a non-main package that doesn't have a main function (i.e., a library), then the `pointer` package does not provide us with a way to determine the points-to set for that argument.

If a call is not reachable by a `main`, I don't think we need to have `pointer` produces a `PointsTo()` set for that call.  If I'm only using one function in a third-party library, I'm okay with skipping the functions in that library I'm not using.

My concern was that this approach wasn't going to play well with the `analysis` package's partitioning of execution and reporting.  What is the behavior of, say,

* `main.main()` calls `other.sendsToSinkWrapper()`
* `sendsToSinkWrapper()` instantiates a `core.Source`, which it sends to `other.sinkWrapper(interface{})`
* `other.sinkWrapper(interface{})` passes its arg to `core.Sink(interface{})`

Since all this would be happening with `pass.Pkg == "main"`, I don't think we would not be allowed to report an `Diagnostic` in `other`.

PurelyApplied:
> Given these limitations, I will begin working on a different approach.

But if you have another tack in mind, it rather moots my other concern.

Issue #47, opened 2020-08-14
Improve the granularity of taint propagation in maps


mlevesquedion:
Currently, when a tainted value is placed in a map, the entire map is considered to be tainted. This is not unreasonable considering that if a map containing a tainted value reaches a sink, the tainted value will almost certainly "reach the sink" as well, since e.g. printing a map implies printing all of its keys and values.

This does have the unfortunate side effect that any key access on the map will yield a value that is considered to be tainted. For example, analyzing the following instructions yields a diagnostic, even though the value that reaches the sink is not really tainted:
```
m := map[string]interface{}{"source": s}
core.Sink(m[""]) // diagnostic: "a source has reached a sink"
```

This sequence of instructions, in which a map is properly "sanitized" (ahem...) also yields a diagnostic:
```
m := map[string]interface{}{"source": s}
delete(m, "source")
core.Sink(m) // diagnostic: "a source has reached a sink"
```

We would like to be able to determine which keys/values in a map are actually tainted, in order to avoid producing false positives.

mlevesquedion:
Some additional cases to consider:

```go
func TestIndexByIndexSanitization(sources []core.Source) {
	sanitizedSources := make([]core.Source, len(sources))
	for i := range sources {
		sanitizedSources[i] = core.Sanitize(sources[i])[0].(core.Source)
	}
	core.Sink(sanitizedSources)
}
```
Index-by-index sanitization does not work in this case because `sanitizedSources` is considered a Source because of its type.

```go
func TestIndexByIndexSanitizationEface(sources []core.Source) {
	sanitizedSources := make([]interface{}, len(sources))
	for i := range sources {
		sanitizedSources[i] = core.Sanitize(sources[i])[0].(core.Source)
	}
	core.Sink(sanitizedSources)
}
```
This case is more complicated. Among other things, it may involve #155.

Currently, one way to circumvent this is to define a sanitizer that can handle collections, and to sanitize the whole collection in one go:
```go
func TestWholeSliceSanitization(sources []core.Source) {
	sanitizedSources := core.Sanitize(sources)
	core.Sink(sanitizedSources)
}
```

PR #53, opened 2020-08-18
Add tests for propagation through arrays

mlevesquedion:
This PR adds tests demonstrating the behavior of our analyzer concerning taint propagation through arrays.

At present, this behavior is subject to some limitations. We consider the whole array to be tainted as soon as it is touched by a `Source`, and we do not consider the array to be safe even if the `Source` value is overwritten. This is in relation with issue #47.

- [x] Tests pass


PurelyApplied:
I guess as a nit, maybe mention that this is test-focused in the commit message.

PR #58, opened 2020-08-24
Introduce function abstraction for cross-function analysis

mlevesquedion:
(This PR is intended as a precursor to #57 to reduce the diff.)

This PR introduces a `Function` abstraction that can be used to query a function about what it does with its arguments. Among other things, having such an abstraction will allow us to handle the following cases:

```
func SinksFive(s core.Source) {
	five := ReturnsFive(s)
	core.Sink(five)
}

func ReturnsFive(a interface{}) int {
	return 5
}
```
Currently, when a function is called with a tainted value, we traverse through its return value no matter what the function's behavior is. Here, `ReturnsFive` returns a value that is not tainted, as can be seen easily by analyzing its body. Our current analysis does not know about that though, and it assumes the return value is tainted, so it incorrectly produces a diagnostic for `core.Sink(five)`. Using the abstraction introduced in this PR would allow us to handle this case in the following way:
1. Analyze the body of the `ReturnsFive` function. Discover that it does not sink its argument, nor does its argument taint any of its return values.
2. While analyzing the body of `SinksFive`, we find a call to `ReturnsFive` that takes a `Source` as argument. We ask the `ReturnsFive` function if it sinks that argument, and it says no. We then ask it if its return value is tainted by this argument. It once again says no, so we know we do not need to traverse through the return value, preventing us from reaching the sink, avoiding a false positive diagnostic.

```
func TestStringify(s core.Source) {
	s := Stringify(e)
	core.Sink(s)
}

func Stringify(e interface{}) string {
	return fmt.Sprintf("%v", e)
}
```
In this case, the value returned by `Stringify` *is* tainted if its `e interface{}` argument is tainted. Analyzing the body of `Stringify` would tell us that, and then while analyzing `TestStringify` we would traverse through the tainted return value to find the call to `core.Sink`.

The following are not handled by the proposed abstraction:
- Methods
- Tainting of colocated arguments, e.g. `fmt.Fprintf(w, Source{})`

(I think for a first stab at cross-function analysis, handling those is out of scope.)

- [x] Tests pass

### DIFF
--- a/internal/pkg/cfa/function.go
+++ b/internal/pkg/cfa/function.go
@@ -86,12 +86,23 @@ func newGenericFunc(f *ssa.Function) GenericFunc {
 	}
 }
 
-func (g GenericFunc) Sinks(arg int) bool {
-	return g.sinks[arg]
+// Sinks returns whether the GenericFunc's nth argument reaches a sink.
+// This function will panic if `n` is out of range,
+// i.e. n <= 0 || n >= len(g.sinks)
+// Such a panic can only occur due to a programming error, which should
+// be caught during development.
+func (g GenericFunc) Sinks(n int) bool {
+	return g.sinks[n]
 }
 
-func (g GenericFunc) Taints(arg int) []int {
-	return g.taints[arg]
+// Taints returns the positions of the return values reached by GenericFunc's
+// nth argument.
+// This function will panic if `n` is out of range,
+// i.e. n <= 0 || n >= len(g.sinks)
+// Such a panic can only occur due to a programming error, which should
+// be caught during development.
+func (g GenericFunc) Taints(n int) []int {
+	return g.taints[n]
 }
 
 func (g GenericFunc) String() string {
@@ -123,7 +134,7 @@ func (g GenericFunc) String() string {
 	return b.String()
 }
 
-// Results returns the number of return values that this function has.
+// Results returns the number of results (return values) that this function has.
 func (g GenericFunc) Results() int {
 	return g.results
 }
--- a/internal/pkg/cfa/function.go
+++ b/internal/pkg/cfa/function.go
@@ -67,29 +67,34 @@ func (s sanitizer) String() string {
 	return "sanitizer"
 }
 
-type genericFunc struct {
+// A GenericFunc is a generic Go function, i.e. neither a sink nor a sanitizer.
+// As such, each of its arguments may or may not reach a sink. Also, each of
+// its arguments may taint 0 or more of its return values.
+// Because it may taint its return values, GenericFunc provides a way for
+// users to know how many results (return values) it has.
+type GenericFunc struct {
 	sinks   []bool
 	taints  [][]int
 	results int
 }
 
-func newGenericFunc(f *ssa.Function) genericFunc {
+func newGenericFunc(f *ssa.Function) GenericFunc {
 	params := f.Signature.Params().Len()
-	return genericFunc{
+	return GenericFunc{
 		sinks:  make([]bool, params),
 		taints: make([][]int, params),
 	}
 }
 
-func (g genericFunc) Sinks(arg int) bool {
+func (g GenericFunc) Sinks(arg int) bool {
 	return g.sinks[arg]
 }
 
-func (g genericFunc) Taints(arg int) []int {
+func (g GenericFunc) Taints(arg int) []int {
 	return g.taints[arg]
 }
 
-func (g genericFunc) String() string {
+func (g GenericFunc) String() string {
 	var b strings.Builder
 	b.WriteString("genericFunc{ ")
 
@@ -119,6 +124,6 @@ func (g genericFunc) String() string {
 }
 
 // Results returns the number of return values that this function has.
-func (g genericFunc) Results() int {
+func (g GenericFunc) Results() int {
 	return g.results
 }
--- a/internal/pkg/cfa/function.go
+++ b/internal/pkg/cfa/function.go
@@ -67,17 +67,17 @@ func (s sanitizer) String() string {
 	return "sanitizer"
 }
 
-// A GenericFunc is a generic Go function, i.e. neither a sink nor a sanitizer.
+// A genericFunc is a generic Go function, i.e. neither a sink nor a sanitizer.
 // As such, each of its arguments may or may not reach a sink. Also, each of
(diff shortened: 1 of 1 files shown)


internal/pkg/cfa/function.go
```diff
@@ -0,0 +1,124 @@
...
+func newGenericFunc(f *ssa.Function) genericFunc {
+	params := f.Signature.Params().Len()
+	return genericFunc{
+		sinks:  make([]bool, params),
+		taints: make([][]int, params),
+	}
+}
+
+func (g genericFunc) Sinks(arg int) bool {
+	return g.sinks[arg]
```
mlevesquedion, line 85: This assumes that we never call `Sinks` (same with `Taints` below) with an out of range argument (e.g. `arg < 0` or `arg >= len(g.sinks)`). I think with appropriate test coverage this is a reasonable assumption to make.

internal/pkg/cfa/function.go
```diff
@@ -0,0 +1,124 @@
...
+)
+
+// A Function is an abstraction for a Go function.
+// It can be queried about what it does with its arguments.
+type Function interface {
+	// Does this function's nth argument reach a sink?
+	Sinks(arg int) bool
+
+	// If this argument is tainted, which return values are tainted?
+	Taints(arg int) []int
```
PurelyApplied, line 39: Mentioned offline / in another thread, but this interface should also support argument tainting.  The taint on, for instance, [writing to a buffer](https://golang.org/pkg/bytes/#Buffer.Write) is not on the return values.

internal/pkg/cfa/function.go
```diff
@@ -0,0 +1,124 @@
...
+}
+
+type sink struct{}
```
PurelyApplied, line 42: Are these structs meant to represent how we currently handle sinks and taint?  Are they meant for consumption outside this package as we iterate towards integrating this interface into existing code?  Should they be exported?

A comment would be good.  Otherwise, it's appears to be dead code.

internal/pkg/cfa/function.go
mlevesquedion: I agree that this should be handled eventually, but for a first stab at cross-function analysis I think that would be too much. Even when the current PR is merged, #57 will have a large diff, and it is already rather complex, so I think it would be preferable to wait until a future PR.

internal/pkg/cfa/function.go
mlevesquedion: `sink` and `sanitizer` are only intended for in-package use. Outside users will be expected to interact with them through the interface. Should I add comments even though they're unexported?

`genericFunc` should be exported though, since it has an additional method outside the interface that users will need. I've exported it and added a comment.

internal/pkg/cfa/function.go
mlevesquedion: Sounds good.

Originally I did not put a comment because it's an implementation of an interface function, so I would assume users would look at the docstring in the interface. Is that a reasonable assumption? In this case since there is something special going so I agree that we need a comment, but do we need a comment in general? (e.g. for `(sink|sanitizer).(Sinks|Taints)`?

internal/pkg/cfa/function.go
```diff
@@ -0,0 +1,140 @@
...
+
+func (s sink) Sinks(_ int) bool {
```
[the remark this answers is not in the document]
mlevesquedion: It may be a little weird, but currently I think it is the correct behavior: a sink is assumed to be sinking all of its arguments

internal/pkg/cfa/function.go
```diff
@@ -0,0 +1,140 @@
...
+func (s sanitizer) String() string {
+	return "sanitizer"
+}
+
+// A GenericFunc is a generic Go function, i.e. neither a sink nor a sanitizer.
+// As such, each of its arguments may or may not reach a sink. Also, each of
+// its arguments may taint 0 or more of its return values.
+// Because it may taint its return values, GenericFunc provides a way for
+// users to know how many results (return values) it has.
+type GenericFunc struct {
```
immutableT, line 75: What does the term 'Generic' signify here? Do we expect non-generic functions?

internal/pkg/cfa/function.go
```diff
@@ -0,0 +1,140 @@
...
+	"sort"
+	"strconv"
+	"strings"
+
+	"golang.org/x/tools/go/ssa"
+)
+
+// A Function is an abstraction for a Go function.
+// It can be queried about what it does with its arguments.
+type Function interface {
```
immutableT, line 34: Why is this better than adding `Sinks` method to `Source`?
This assumes we fix the graph logic in the `Source`.

internal/pkg/cfa/function.go
mlevesquedion: `Generic` means a regular function. Would `RegularFunction` be more evocative?

Non-generic functions would be "special" functions such as `sink`s and `sanitizer`s.

internal/pkg/cfa/function.go
mlevesquedion: I think these are orthogonal concerns. The `Sinks` method here is for cross-function analysis, it allows us to know if a function sinks its nth argument. A `Sinks` method on `Source` (if I correctly recall what you suggested elsewhere) would return the sinks that are reachable from a source. I think the latter would eventually use the former.

internal/pkg/cfa/function.go
mlevesquedion: I must have been writing too much Go code lately, because I forgot what the term "generic" usually means when applied to functions. To be clear, I did not mean generic functions such as `func<T> sort(s []T)`.

Do you think the name is confusing?

mlevesquedion:
I found a way to rewrite the code in the main PR #57 such that the `GenericFunc.Results()` function is no longer required. Consequently, I am removing it from this PR.

I think this makes a clearer case for the abstraction I'm proposing: it is sufficient to support basic cross-function analysis.

internal/pkg/cfa/function.go
mlevesquedion: I've unexported `genericFunc` due to no longer needing the extra out-of-interface method. I've kept the comment though.

mlevesquedion:
Closing because of staleness. This will likely be revisited in the future, but right now isn't a priority.

PR #62, opened 2020-08-24
To discuss: Issues and prioritization for a '1.0'

PurelyApplied:
I wanted to discuss how we are prioritizing certain issues.

While I don't necessarily think this document should be merged into `master`, I wanted to have a place where we could fork discussions by line item, since I foresee many sub-discussions occurring.  After discussion, each pending item should have an Issue opened to track it, and those we deem higher priority should be tagged with a `1.0 target` tag.

### DIFF
--- a/design/1.0.md
+++ b/design/1.0.md
@@ -13,6 +13,8 @@ Given a spec defining sources and sinks (below), diagnostics are reported for
 any direct call to a sink that includes a source.  Additionally, we detect any
 taint propagation from a source to a sink within the scope of a single function.
 
+Analysis does not explicitly track across the use of reflection.
+
 ## Analysis Configuration
 
 Configuration will consist of the following key components:
@@ -32,20 +34,24 @@ In that light, specification should follow:
 A ValueSpecifier marks an `ssa.Value`, e.g., for identification as a source or as the safe result of a sanitizer.
 Values should be specifiable via any or all of:
 
+* Specifier name: For identification during reporting
 * Type / Field: Package path and type name, optional field name
 * Field tags
 * Scope: local variable, free variable, or global
 * Context: within a specific package or function call.  See CallSpecifier below.
 * Is a reference
 * Is "reference-like," e.g., slices, maps, chans.
+* Const value (e.g., `"PASSWORD"`)
+* Argument position (for use below)
 
 #### CallSpecifier
 A CallSpecifier marks an `*ssa.Call`, e.g., as a sink or sanitizer.
 Calls should be specifiable via any or all of:
 
+* Specifier name: For identification during reporting
 * Symbol: package path, function name, optional receiver name.
-* Context: A CallSpecifier indicating  
-* Based on arguments (ValueSpecifier)
+* Context: A CallSpecifier indicating scope, such as package or specific functions.
+* Based on argument value and position (ValueSpecifier).
 * Based on return values (ValueSpecifier)
 
 ## Inference
@@ -63,6 +69,7 @@ These items are open to interpretation and warrant discussion.
 
 ## Improved Inference
 
+* Detect partial sanitization of collections.
 * Detect sanitization via zero-value.
     * Explicit sanitization of `type Source struct{ data string }` is easily done via `s.data = ""`.
     * Alternatively, a user may specify redaction placeholders, e.g. `"REDACTED"` rather than require a zero value.
--- a/design/1.0.md
+++ b/design/1.0.md
@@ -1,0 +1,84 @@
+I recently opened a Kubernetes Enhancement Proposal to introduce `go-flow-levee` during testing as a defense against accidental credential logging.
+See [KEP-1933](https://github.com/kubernetes/enhancements/pull/1936) details.
+Such integration demands a stable API and well-defined scope.
+
+I propose the following.
+Portions may already be implemented, but are included here as a "big picture."
+
+# 1.0 Targets
+
+## Basic taint propagation detection
+
+Given a spec defining sources and sinks (below), diagnostics are reported for
+any direct call to a sink that includes a source.  Additionally, we detect any
+taint propagation from a source to a sink within the scope of a single function.
+
+## Analysis Configuration
+
+Configuration will consist of the following key components:
+
+* Identification of sources, sinks, and sanitizers
+* Identification of analysis scope, e.g. to skip third-party or testing packages.
+* Specify diagnostics to be ignored, e.g false-positives
+
+### Identification of sources, sinks, and sanitizers
+
+Much of our work is directly with `ssa.Value`s and `ssa.Instruction`s.
+Whether those values or instructions represent one element or another,
+ the configuration to identify these members should be as uniform as possible.
+In that light, specification should follow:
+
+#### ValueSpecifier
+A ValueSpecifier marks an `ssa.Value`, e.g., for identification as a source or as the safe result of a sanitizer.
+Values should be specifiable via any or all of:
+
+* Type / Field: Package path and type name, optional field name
+* Field tags
+* Scope: local variable, free variable, or global
+* Context: within a specific package or function call.  See CallSpecifier below.
+* Is a reference
+* Is "reference-like," e.g., slices, maps, chans.
+
+#### CallSpecifier
+A CallSpecifier marks an `*ssa.Call`, e.g., as a sink or sanitizer.
+Calls should be specifiable via any or all of:
+
+* Symbol: package path, function name, optional receiver name.
+* Context: A CallSpecifier indicating  
+* Based on arguments (ValueSpecifier)
+* Based on return values (ValueSpecifier)
+
+## Inference
+
+Inference represents the scope of what a user does *not* need to configure.
+For instance, we can currently detect getter methods and mark as propagating
+a marked field without requiring any explicit configuration.
+
+If a source type is aliased, the alias is properly identified as equivalent as the aliased source type.
+
+# Beyond a 1.0
+
+The following seem like significant improvements but seem larger in scope than what constitutes a stable 1.0.
+These items are open to interpretation and warrant discussion.
+
+## Improved Inference
+
+* Detect sanitization via zero-value.
+    * Explicit sanitization of `type Source struct{ data string }` is easily done via `s.data = ""`.
+    * Alternatively, a user may specify redaction placeholders, e.g. `"REDACTED"` rather than require a zero value.
+* A type defined from a source type (e.g., `type Foo core.Source`) should also be considered a source type.
+    * Note: as tags take part in type identify of the underlying struct literal, field tags are "inherited" in defined types based on existing types.
+      This may result in conflicting current behavior between those sources identified by tag and those identified by path/package/name.
+* A type which contains a known source type as a field should itself be a source type and associated field.  E.g., `type Wrapper struct { data core.Source }`
+* Current effort under the umbrella of *cross functional analysis* greatly refines inference of safety in scenarios that would currently produce a finding.
+
+## Reporting
+
+At present, reporting of Diagnostics is a generic `a source has reached a sink` message with source and sink positions included.
+As we iterate to include wider inference and definitions of sources, sinks, cross-function analysis, etc, the path between a source and a sink can become ambiguous.
+If much inference is targeted for a 1.0 release, reporting will need to be improved to remain meaningful.
+
+## Extensibility
+
+A user may have highly specific considerations for the definition of a source/sink/sanitizer and may wish to implement custom specification code that interacts directly with `ssa` / `types` packages.
+We should expose an extensible interface for this purpose.
--- a/design/1.0.md
+++ b/design/1.0.md
@@ -62,6 +62,12 @@ a marked field without requiring any explicit configuration.
 
 If a source type is aliased, the alias is properly identified as equivalent as the aliased source type.
 
+If `Source` is identified as a source type, collection types based on that should also be identified as a source type.
+
+## Internal Process
+
+Add benchmarks.
+
 # Beyond a 1.0
 
 The following seem like significant improvements but seem larger in scope than what constitutes a stable 1.0.


design/1.0.md
```diff
@@ -0,0 +1,84 @@
...
+* Context: within a specific package or function call.  See CallSpecifier below.
+* Is a reference
+* Is "reference-like," e.g., slices, maps, chans.
+
+#### CallSpecifier
+A CallSpecifier marks an `*ssa.Call`, e.g., as a sink or sanitizer.
+Calls should be specifiable via any or all of:
+
+* Symbol: package path, function name, optional receiver name.
+* Context: A CallSpecifier indicating
```
mlevesquedion, line 47: Looks like this bullet point is incomplete.

design/1.0.md
```diff
@@ -0,0 +1,84 @@
...
+* Based on arguments (ValueSpecifier)
+* Based on return values (ValueSpecifier)
+
+## Inference
+
+Inference represents the scope of what a user does *not* need to configure.
+For instance, we can currently detect getter methods and mark as propagating
+a marked field without requiring any explicit configuration.
+
+If a source type is aliased, the alias is properly identified as equivalent as the aliased source type.
```
mlevesquedion, line 57: For context, we already handle aliases e.g. `type MySource = core.Source`, but not named types e.g. `type MySource core.Source`. (Interestingly, in the short investigation I just did, the SSA was exactly the same for both cases).

Do we want named types to also count as "aliases"? Playing devil's advocate: what if a user defines a named type `type SafeSource core.Source` and uses it to explicitly differentiate between sources that have and have not been properly sanitized?

design/1.0.md
```diff
@@ -0,0 +1,84 @@
...
+A ValueSpecifier marks an `ssa.Value`, e.g., for identification as a source or as the safe result of a sanitizer.
+Values should be specifiable via any or all of:
+
+* Type / Field: Package path and type name, optional field name
+* Field tags
+* Scope: local variable, free variable, or global
+* Context: within a specific package or function call.  See CallSpecifier below.
+* Is a reference
+* Is "reference-like," e.g., slices, maps, chans.
+
```
mlevesquedion, line 41: In addition to those, users might want to be able to specify specific _values_. For a 1.0, we could choose to support only basic literals (strings, ints, etc.), e.g. : 

`os.Getenv("PASSWORD")`.

design/1.0.md
```diff
@@ -0,0 +1,84 @@
...
+#### CallSpecifier
+A CallSpecifier marks an `*ssa.Call`, e.g., as a sink or sanitizer.
+Calls should be specifiable via any or all of:
+
+* Symbol: package path, function name, optional receiver name.
+* Context: A CallSpecifier indicating  
+* Based on arguments (ValueSpecifier)
```
mlevesquedion, line 48: I think it would be valuable to have more information in addition to the ValueSpecifier, e.g. the argument's position in the call. This would be useful if, for example, I have a function `DoWork(a interface{}, b interface{})` which only sinks its first argument.

design/1.0.md
```diff
@@ -0,0 +1,84 @@
...
+* Based on return values (ValueSpecifier)
```
mlevesquedion, line 49: I'm not sure it makes sense to identify a sink/sanitizer based on the return value. Could you give an example?

design/1.0.md
```diff
@@ -0,0 +1,84 @@
...
+
+# Beyond a 1.0
+
+The following seem like significant improvements but seem larger in scope than what constitutes a stable 1.0.
+These items are open to interpretation and warrant discussion.
+
+## Improved Inference
+
+* Detect sanitization via zero-value.
+    * Explicit sanitization of `type Source struct{ data string }` is easily done via `s.data = ""`.
```
mlevesquedion, line 67: This means eventually, we could detect some sanitizers automatically. We may still want to let the user configure sanitizer detection though, e.g. if their sanitizer uses reflection.

design/1.0.md
```diff
@@ -0,0 +1,84 @@
...
+    * Alternatively, a user may specify redaction placeholders, e.g. `"REDACTED"` rather than require a zero value.
+* A type defined from a source type (e.g., `type Foo core.Source`) should also be considered a source type.
+    * Note: as tags take part in type identify of the underlying struct literal, field tags are "inherited" in defined types based on existing types.
+      This may result in conflicting current behavior between those sources identified by tag and those identified by path/package/name.
+* A type which contains a known source type as a field should itself be a source type and associated field.  E.g., `type Wrapper struct { data core.Source }`
+* Current effort under the umbrella of *cross functional analysis* greatly refines inference of safety in scenarios that would currently produce a finding.
```
mlevesquedion, line 73: Beyond cross function analysis, would we eventually want to get to *whole program* analysis? If so, we would need to discuss what that means in the context of our analysis. Here's an example scenario that I think we may eventually want to handle:
```
var s = core.Source{data: "SECRET"}

func main() {
  sanitizeSource() // this sanitizes the global
  sinkSource() // this sinks the global
}
```
This scenario is pretty terrible in terms of programming practices, but there might be situations where it would make sense for a user to have a global `Source`.

design/1.0.md
```diff
@@ -0,0 +1,84 @@
...
+## Reporting
+
+At present, reporting of Diagnostics is a generic `a source has reached a sink` message with source and sink positions included.
+As we iterate to include wider inference and definitions of sources, sinks, cross-function analysis, etc, the path between a source and a sink can become ambiguous.
+If much inference is targeted for a 1.0 release, reporting will need to be improved to remain meaningful.
+
+## Extensibility
+
+A user may have highly specific considerations for the definition of a source/sink/sanitizer and may wish to implement custom specification code that interacts directly with `ssa` / `types` packages.
+We should expose an extensible interface for this purpose.
```
mlevesquedion, line 84: I think a reasonably easy way for us to do this would be to allow users to register a custom `analysis.Analyzer` as a `Requires` dependency of one of our analyzers, with a convention concerning the kind of results that it such an analyzer could produce. Our analyzer would then extract the results and use them in our analysis. E.g. if we wanted custom source detection, our `source` analyzer would extract the results from the user's custom analyzer and create sources from them.

design/1.0.md
```diff
@@ -0,0 +1,84 @@
...
+a marked field without requiring any explicit configuration.
+
+If a source type is aliased, the alias is properly identified as equivalent as the aliased source type.
+
+# Beyond a 1.0
+
+The following seem like significant improvements but seem larger in scope than what constitutes a stable 1.0.
+These items are open to interpretation and warrant discussion.
+
+## Improved Inference
```
mlevesquedion, line 64: I assume propagation/sanitization involving collections would fall under here as well?

mlevesquedion:
This is great!

mlevesquedion:
Would it help for us to give specific examples of cases we want/do not want to handle?

design/1.0.md
```diff
@@ -0,0 +1,84 @@
...
+Given a spec defining sources and sinks (below), diagnostics are reported for
+any direct call to a sink that includes a source.  Additionally, we detect any
+taint propagation from a source to a sink within the scope of a single function.
+
+## Analysis Configuration
+
+Configuration will consist of the following key components:
+
+* Identification of sources, sinks, and sanitizers
+* Identification of analysis scope, e.g. to skip third-party or testing packages.
```
mlevesquedion, line 21: If we skip certain packages (or directories/paths?), we might want to identify contradictions in the config, e.g. : 
```
PackageDenyList: [
  "third_party",
],
Sink: {
  package: "third_party",
  method: "sink"
}
```

design/1.0.md
```diff
@@ -0,0 +1,84 @@
...
+* Specify diagnostics to be ignored, e.g false-positives
```
mlevesquedion, line 22: Should we also give users a way to report false-positives? This could be a valuable source of ideas for how to improve the analyzer.

design/1.0.md
```diff
@@ -0,0 +1,84 @@
...
+These items are open to interpretation and warrant discussion.
+
+## Improved Inference
+
+* Detect sanitization via zero-value.
+    * Explicit sanitization of `type Source struct{ data string }` is easily done via `s.data = ""`.
+    * Alternatively, a user may specify redaction placeholders, e.g. `"REDACTED"` rather than require a zero value.
+* A type defined from a source type (e.g., `type Foo core.Source`) should also be considered a source type.
+    * Note: as tags take part in type identify of the underlying struct literal, field tags are "inherited" in defined types based on existing types.
+      This may result in conflicting current behavior between those sources identified by tag and those identified by path/package/name.
```
mlevesquedion, line 71: I think I understand what you mean, but could you give an example here?

design/1.0.md
```diff
@@ -0,0 +1,84 @@
...
+* A type which contains a known source type as a field should itself be a source type and associated field.  E.g., `type Wrapper struct { data core.Source }`
+* Current effort under the umbrella of *cross functional analysis* greatly refines inference of safety in scenarios that would currently produce a finding.
+
+## Reporting
+
+At present, reporting of Diagnostics is a generic `a source has reached a sink` message with source and sink positions included.
+As we iterate to include wider inference and definitions of sources, sinks, cross-function analysis, etc, the path between a source and a sink can become ambiguous.
+If much inference is targeted for a 1.0 release, reporting will need to be improved to remain meaningful.
```
mlevesquedion, line 79: I think it could also be useful for users to know *why* a given source/sink was identified as such. E.g. _this_ matcher in your config said that this call is a call to a sink. This would likely become more important when we implement value/call specifiers.

Also, currently we do not report anything when a source is sanitized, but it might be interesting for some users to be able to see those, e.g. with a `-verbose` flag or something.

design/1.0.md
PurelyApplied: > Do we want named types to also count as "aliases"?

Opinion: I think we should include named types.  The only difference between the two types would be the method set associated, which to me suggests that the kind of data stored is reasonably similar.  Also, mentioned below, I think it would be the easier way to handle the fact that field tags are naturally inherited in a way that a source specified by type path and name would not be.

Although...

> Playing devil's advocate: what if a user defines a named type type SafeSource core.Source and uses it to explicitly differentiate between sources that have and have not been properly sanitized?

I hadn't thought of that.  I think I might prefer an edge-case false-positive to a systematic false-negative, though.

Also and alternatively: this could be enabled/disabled in configuration.  And/or we could allow a user to include a safe-list for types that should not be considered sources, though that feels like it's getting a bit over-elaborate.

design/1.0.md
PurelyApplied: I'd need to test it, but I believe that...

```golang
type SourceA struct {
  Data string    // specified in config by package, name, and field
}

type SourceB struct {
  Data string  // `levee:"source"`
}

type NamedA SourceA  // Not identified as a source

type NamedB SourceB  // Is identified by source
```

Since the underlying struct for both `SourceB` and `NamedB` is the struct literal
```golang
struct {
  Data string // `levee:"source"`
}```

then both will satisfy the identification by field tags.  But since the other is identified by name-matching regexp, the change of name will lost track of it.

Both of these should be handled uniformly by the `sourceType` analyzer (which also need to actually be consumed by the `source` analyzer, and either be unified with or consume the output of the `fieldtags` analyzer).

File: design/1.0.md (hunk 13 above)
Comment (PurelyApplied) [reply design/1.0.md:79]: Agreed.  Added `Specifier Name` bullets to both specifier outlines.

File: design/1.0.md (hunk 3 above)
Comment (PurelyApplied) [reply design/1.0.md:41]: I had had "Const" in my noted but couldn't remember what I meant by it.  Good job.

It does occur to me (and this is perhaps implementation detail beyond the current scope) that if we're concerned about `os.Getenv("PASSWORD")`, we need to be also concerned about `func(s string) { os.Genenv(s) }`... I'm not sure how I'd want to specify "This is a source when called on this `const` term, or on any ambiguous term, but not when called on a `const` term like `"PATH"`."

File: design/1.0.md (hunk 9 above)
Comment (PurelyApplied) [reply design/1.0.md:64]: Hmmm...

I think propagation in general should be part of 1.0, and I believe you've made strides to cover most collections and are working on the remaining?

I think propagation and sanitization of specific elements within a collection might be a `Beyond 1.0` target.

Comment (mlevesquedion) [reply design/1.0.md:70]: Ah. Important distinction here. Yes, propagation/sanitization of specific elements is what I had in mind.

I think for single-function propagation all that's left is `chan`s.

File: design/1.0.md [hunk 14]
```diff
@@ -0,0 +1,91 @@
...
+
+### Identification of sources, sinks, and sanitizers
+
+Much of our work is directly with `ssa.Value`s and `ssa.Instruction`s.
+Whether those values or instructions represent one element or another,
+ the configuration to identify these members should be as uniform as possible.
+In that light, specification should follow:
+
+#### ValueSpecifier
+A ValueSpecifier marks an `ssa.Value`, e.g., for identification as a source or as the safe result of a sanitizer.
```
immutableT, line 34: Do you expect that your audience (ex. kubernetes contributors) would be familiar with these concepts of SSA?

I think  that, at the most, we could assume that our audience/users are familiar with Go. Therefore, consider rephrasing in terms of concepts familiar to Go developers.

design/1.0.md
```diff
@@ -0,0 +1,91 @@
...
+Values should be specifiable via any or all of:
+
+* Specifier name: For identification during reporting
+* Type / Field: Package path and type name, optional field name
+* Field tags
+* Scope: local variable, free variable, or global
```
immutableT, line 40: We don't do this today, right?

design/1.0.md
```diff
@@ -0,0 +1,91 @@
+I recently opened a Kubernetes Enhancement Proposal to introduce `go-flow-levee` during testing as a defense against accidental credential logging.
+See [KEP-1933](https://github.com/kubernetes/enhancements/pull/1936) details.
+Such integration demands a stable API and well-defined scope.
```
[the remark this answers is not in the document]
mlevesquedion: To me our analyzer's "API" is its observable behavior, which includes the way it is configured, the cases that it handles/does not handle, and the way it reports what it finds.

design/1.0.md
```diff
@@ -0,0 +1,91 @@
...
+
+I propose the following.
+Portions may already be implemented, but are included here as a "big picture."
+
+# 1.0 Targets
```
mlevesquedion, line 8: Would benchmarking be a 1.0 target? I think speed should be a part of the promise we make to users. Benchmarks would let users know what to expect and would allow us to make sure that our analyzer doesn't slow down as we make changes to it.

design/1.0.md
mlevesquedion: I think it would indeed be difficult to say "This is a source when called on an ambiguous term". I think the `func(s string)` case you mention would require cross-function analysis to do well. If the variable is defined within the same function we might be able to do better though, e.g. I think we could handle `func() { pass := "PASSWORD"; os.Getenv(pass) }`.

design/1.0.md
mlevesquedion: Correct.

design/1.0.md
mlevesquedion: I think you make a good case for named type inference.

I think the scenario I proposed is sufficiently unlikely that we should probably wait until we actually hit it before we think too deeply about how to handle it.

I think as a user I would expect a named type that aliases a Source to also be considered as Source.

design/1.0.md
mlevesquedion: I think asking for a name is a good idea. If the user doesn't provide a name, we could easily fall back on a string representation for the matcher. In fact we might want to do that even if the user specifies a name, so they can clearly see why it matched.

design/1.0.md
mlevesquedion: Good explanation, thanks! This will be something to keep in mind then.

design/1.0.md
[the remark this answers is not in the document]
PurelyApplied: > which includes the way it is configured

That's mostly how I meant it, although explicitly commenting as to what is and is not reported is also important.

design/1.0.md
```diff
@@ -0,0 +1,91 @@
...
+The following seem like significant improvements but seem larger in scope than what constitutes a stable 1.0.
+These items are open to interpretation and warrant discussion.
+
+## Improved Inference
+
+* Detect partial sanitization of collections.
+* Detect sanitization via zero-value.
+    * Explicit sanitization of `type Source struct{ data string }` is easily done via `s.data = ""`.
+    * Alternatively, a user may specify redaction placeholders, e.g. `"REDACTED"` rather than require a zero value.
+* A type defined from a source type (e.g., `type Foo core.Source`) should also be considered a source type.
```
mlevesquedion, line 76: Should this also include collection type wrappers, e.g. :
```
type SourceSlice []Source
type SourceMap map[string]Source
```

design/1.0.md
```diff
@@ -0,0 +1,91 @@
...
+* Is "reference-like," e.g., slices, maps, chans.
+* Const value (e.g., `"PASSWORD"`)
+* Argument position (for use below)
+
+#### CallSpecifier
+A CallSpecifier marks an `*ssa.Call`, e.g., as a sink or sanitizer.
+Calls should be specifiable via any or all of:
+
+* Specifier name: For identification during reporting
+* Symbol: package path, function name, optional receiver name.
```
mlevesquedion, line 52: Do we want to handle the following cases:
1. `func`s stored in `struct` fields, e.g.
```go
type MySinker struct {
  sinkFunc func(args ...interface{})
}
```

2. `func`s passed as function parameters:
```go
func callSink(arg interface{}, sink func(...interface{})) {
  sink(arg)
}
```
In theory, since `func`s are first class in Go, the actual range of cases we could potentially handle is much wider, e.g. a user could put functions in a `map[string]func()`, store functions in local variables, etc.

There are some cases we already handle, e.g.:
```go
func Test(s core.Source) {
  f := core.Sink
  f(s) // a diagnostic is produced here
}
```

But it is quite easy to throw our analysis off in this case:
```go
func Test(s core.Source) {
  var f func(...interface{})
  if true {
    f = core.Sink
  }
  f(s) // no diagnostic
}
```

My opinion: I think we should handle `func`s stored in `struct` fields, because that seems to be a fairly common pattern. The other cases seem more outlandish and seem like they would require some kind of cross-function or even whole-program analysis, so I think for a 1.0 they are out of scope.

PurelyApplied: > My opinion: I think we should handle funcs stored in struct fields, because that seems to be a fairly common pattern. The other cases seem more outlandish and seem like they would require some kind of cross-function or even whole-program analysis, so I think for a 1.0 they are out of scope.

We currently rely on the static dispatch in our call matching.  In effect, we cover only case (a) of the four kinds of Call described [in the docs](https://godoc.org/golang.org/x/tools/go/ssa#CallCommon).

I think we should make a reasonable effort to cover all four cases.  (b) seems achievable in our current framework, and (c) seems a bit out-of-scope.  Dynamic dispatch in case (d) is probably a whole can of worms.  We could rely on `pointer.PointsTo` to get a set of possible functions, but that predicates whole-program analysis.

I don't think the fact that a given `func` is stored in a struct field will make the problem any simpler.  I'd be inclined to put this at a "post-1.0 but preferably 1.1" sort of target.

design/1.0.md
PurelyApplied: I have been targeting "Willing to reread the Go spec" as the audience.  In this theoretic future configuration, I might reference SSA as additional reading, but I would think everything below should be intuitive to someone familiar with Go.

design/1.0.md
PurelyApplied: \shrug.

I feel until we get to more expensive approaches of cross-function / whole-program analysis, it costs what it costs.  But you're not wrong that they'd be good to have.

I'll add an `Internal Process` section and drop benchmarks in there, though.

design/1.0.md
PurelyApplied: > Beyond cross function analysis, would we eventually want to get to whole program analysis?

I meant this as a catch-all term.  Whole program analysis is a subset of cross-function analysis, right?

> If so, we would need to discuss what that means in the context of our analysis.

Unless we want to include it in 1.0, I think it would be premature to try to nail it down now.

design/1.0.md
PurelyApplied: I think this would be easy to add to the `sourceType` analyzer, which should be wired for downstream consumption Soon™️.  Do you think this inference belongs in 1.0?

design/1.0.md
```diff
@@ -0,0 +1,91 @@
...
+taint propagation from a source to a sink within the scope of a single function.
+
+Analysis does not explicitly track across the use of reflection.
+
+## Analysis Configuration
+
+Configuration will consist of the following key components:
+
+* Identification of sources, sinks, and sanitizers
+* Identification of analysis scope, e.g. to skip third-party or testing packages.
```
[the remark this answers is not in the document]
PurelyApplied: (I believe that) for `buildssa` to produce the SSA graph, the code visible to analysis must be compiled and built.  Consequently, anything relying on `buildssa` necessitates visibility in any third-party dependencies.

I was proposing providing an option to skip *our* analyzers from examining third-party packages.  The pre-submit proposed in the referenced KEP could go either way, depending on what K8s ends up being worried about.

design/1.0.md
mlevesquedion: Hmm. The way I see it it's the other way around, haha! I agree with you that the distinction is not useful right now though.

Just for informational purposes, here is how I see it:
* Cross-function analysis = tracking values across function calls
* Whole-program analysis = tracking any and all values in an entire program, including e.g. globals

design/1.0.md
PurelyApplied: I would think they such problems would be reported via an Issue.  That could perhaps be elaborated in the README / Contributing docs.

mlevesquedion:
I don't think it's something that we need to concern ourselves with immediately, but we might want to keep the following edge case in mind:
`log.Printf("%T", source)`. In this case we shouldn't produce a report because only the `Source`'s type is being sinked.

mlevesquedion:
While analyzing kubernetes, I had a thought that it could be useful to be able to specify a `FieldRE` independently of a type. Right now, we can't do that, because e.g. the following definition for a source:
```
    {
      "PackageRE": "",
      "TypeRE": "",
      "FieldRE": "Secret"
    },
```
will lead to a lot of false positives because pretty much everything will be considered a source.

This ties into the discussion concerning `ValueSpecifier`s, I just want to make sure that we have a way to support this kind of usage, because I think it could be very useful.

PurelyApplied:
I'll create Issues / Labels / Milestones for these topics in the next few days.  I've pushed a copy of this branch to [`design/1.0`](https://github.com/google/go-flow-levee/blob/design/1.0/design/1.0.md) for posterity, but will close this PR once I finish opening all the relevant items.

Starting next week, feel free to open any new Issues directly and we can continue conversation there.

design/1.0.md
PurelyApplied: Rereading this, I think well-defined specifiers catch both of these cases.  It's just a difference of a sink being identified by a `ValueSpecifier` rather than a `CallSpecifier`.  Pseudo-config in yaml for readability and ignoring implementation details:

```yaml
Sinks:
  ValueSpecifiers:
    - name: "MySinker matcher"
      symbol: "path/to/my/pkg.MySinker.sinkFunc"
    - name: "log wrapper"
      context:
       # This is a CallSpecifier
        symbol: "path/to/my/pkg.callSink"
      argPosition: 2
```

design/1.0.md
PurelyApplied: I'm not sure either.  I was just throwing stuff out there.

All the instances that come to my mind are specific runtime values, like a buffer writer returning non-zero bytes written.  But we won't see that in SSA.

Probably best to scratch from the list.

PurelyApplied:
I've open what I believe are all the Issues discussed here and gathered the more immediate priority into [milestone/1](https://github.com/google/go-flow-levee/milestone/1).

Some Issues, such as future interprocedural analysis, are ambiguously defined and should be split into sub-Issues as necessary.

If I've missed one, feel free to open one and we'll continue discussion there.

PR #63, opened 2020-08-25
Take instruction order into account when visiting calls

mlevesquedion:
## The bug
Currently, the analyzer does not differentiate between the following two (abstract) scenarios:
1. Bad
```
taint(value)
sink(value)
```
2. Ok
```
sink(value)
taint(value)
```
Specifically, in this second case, a diagnostic is produced at the `sink` call, even though the `value` has not yet been `taint`ed. This is because we rely mostly on the SSA value graph to determine if a source reaches a sink, and that graph has no concept of the "order" of instructions: it only contains `Referrers`/`Operands` relationships.

## The fix
While traversing from a `Source`, for each `ssa.BasicBlock`, we keep track of the max index of an instruction visited within it. 
When we visit an instruction, we find its index in its block and update the max seen so far if necessary. When we encounter a `Call` instruction, we check that it is higher than the current max for its block, i.e. that it occurs after the value that led us to the call was tainted.

## Future PR's
The solution proposed here only works for single-block functions. If a function has multiple blocks, in most cases this approach will not work as is, so I will need to extend it in a future PR.

- [x] Tests pass

### DIFF
--- a/internal/pkg/source/source.go
+++ b/internal/pkg/source/source.go
@@ -87,7 +87,10 @@ func (a *Source) dfs(n ssa.Node) {
 
 func (a *Source) recordIndex(target ssa.Instruction) {
 	b := target.Block()
-	i := indexInBlock(target)
+	i, ok := indexInBlock(target)
+	if !ok {
+		return
+	}
 	if a.maxInstrReached[b] < i {
 		a.maxInstrReached[b] = i
 	}
@@ -111,7 +114,12 @@ func (a *Source) visitReferrers(referrers *[]ssa.Instruction) {
 
 			// If this call's index is lower than the highest in its block,
 			// then this call is "in the past" and we should stop traversing.
-			if indexInBlock(r) < a.maxInstrReached[r.Block()] {
+			i, ok := indexInBlock(r)
+			if !ok {
+				break
+			}
+
+			if i < a.maxInstrReached[r.Block()] {
 				continue
 			}
 
@@ -288,12 +296,14 @@ func isProducedBySanitizer(v *ssa.Alloc, conf classifier) bool {
 	return false
 }
 
-func indexInBlock(target ssa.Instruction) int {
+// indexInBlock returns this instruction's index in its parent block.
+func indexInBlock(target ssa.Instruction) (int, bool) {
 	for i, instr := range target.Block().Instrs {
 		if instr == target {
-			return i
+			return i, true
 		}
 	}
-	// this cannot happen
-	return -1
+	// we can only hit this return if there is a bug in the ssa package
+	// i.e. an instruction does not appear within its parent block
+	return 0, false
 }
--- a/internal/pkg/source/source.go
+++ b/internal/pkg/source/source.go
@@ -38,11 +38,12 @@ type classifier interface {
 // its referrers.
 // Source.sanitized notes sanitizer calls that sanitize this Source
 type Source struct {
-	node       ssa.Node
-	marked     map[ssa.Node]bool
-	preOrder   []ssa.Node
-	sanitizers []*sanitizer.Sanitizer
-	config     classifier
+	node            ssa.Node
+	marked          map[ssa.Node]bool
+	preOrder        []ssa.Node
+	sanitizers      []*sanitizer.Sanitizer
+	config          classifier
+	maxInstrReached map[*ssa.BasicBlock]int
 }
 
 // Node returns the underlying ssa.Node of the Source.
@@ -53,9 +54,10 @@ func (a *Source) Node() ssa.Node {
 // New constructs a Source
 func New(in ssa.Node, config classifier) *Source {
 	a := &Source{
-		node:   in,
-		marked: make(map[ssa.Node]bool),
-		config: config,
+		node:            in,
+		marked:          make(map[ssa.Node]bool),
+		config:          config,
+		maxInstrReached: map[*ssa.BasicBlock]int{},
 	}
 	a.dfs(in)
 	return a
@@ -68,6 +70,10 @@ func (a *Source) dfs(n ssa.Node) {
 	a.preOrder = append(a.preOrder, n)
 	a.marked[n.(ssa.Node)] = true
 
+	if instr, ok := n.(ssa.Instruction); ok {
+		a.recordIndex(instr)
+	}
+
 	if n.Referrers() != nil {
 		a.visitReferrers(n.Referrers())
 	}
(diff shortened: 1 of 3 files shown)


immutableT:
Do you think it would be possible to leverage SSA's dominance to answer the question of whether a taint instruction dominates the sink instruction.
In other words, could we use the same pattern for exempting a sink if it was sanitized?

mlevesquedion:
I originally thought of using dominance, but then I ran into some issues which made me look for a simpler way. I think certainly dominance could be used in some way, especially in the multi-block case. For example, we could check that at least one "tainting instruction" dominates the call to a sink. However, the problem is that the `Call` instruction can easily end up sandwiched between "tainting" instructions:
```
0: entry
	// top bun
	1(*ssa.Store          ): *t0 = s

	// sandwich contents
	2(*ssa.Alloc          ): t1 = new [1]interface{} (varargs)
	3(*ssa.IndexAddr      ): t2 = &t1[0:int]
	4(*ssa.ChangeInterface): t3 = change interface interface{} <- io.Writer (w)
	5(*ssa.Store          ): *t2 = t3
	6(*ssa.Slice          ): t4 = slice t1[:]
	7(*ssa.Call           ): t5 = example.com/core.Sink(t4...)
 
	// bottom bun
	8(*ssa.UnOp           ): t6 = *t0
	9(*ssa.Alloc          ): t7 = new [1]interface{} (varargs)
	10(*ssa.IndexAddr      ): t8 = &t7[0:int]
	11(*ssa.MakeInterface  ): t9 = make interface{} <- example.com/core.Source (t6)
	12(*ssa.Store          ): *t8 = t9
	13(*ssa.Slice          ): t10 = slice t7[:]
	14(*ssa.Call           ): t11 = fmt.Fprintf(w, "%v":string, t10...)
	15(*ssa.Extract        ): t12 = extract t11 #0
	16(*ssa.Extract        ): t13 = extract t11 #1
	17(*ssa.Return         ): return
```
The tainting instructions at the top dominate the call, but they actually have nothing to do with it. So we would need to find what values are involved in the call, and only consider the tainting instructions that also involve those values (which in this case, are at the bottom and do not dominate the call). So the problem becomes "what values are involved in the call"? In general, figuring that out also requires a graph traversal (e.g. here it is not immediate that the call to `Sink` involves the `Writer` `w`, because of the varargs), and I'm worried that it could get convoluted and that we might end up running in circles: when trying to find what values go into a call, how do we know when to stop? If we don't stop "early enough", we'll end up back at the `Source` value, and then we'll consider the instructions at the top to be using the same values as those the call uses, and we'll incorrectly assume that the call is dominated.

I think for now, it is simpler to use the approach I propose in the current PR. Handling the multi-block elegantly will require more

immutableT:
Could you please add the original Go code for the example above.

mlevesquedion:
Certainly, here it is:
```go
func TestSinkBeforeTainting(s core.Source, w io.Writer) {
	core.Sink(w)
	_, _ = fmt.Fprintf(w, "%v", s)
}
```

mlevesquedion:
Looking at the SSA graph, I think finding the values + instructions involving the values might not be so difficult after all:
image

Starting from the call to `Sink` on the left:
image

If we traverse through `Instructions` and we stop traversing when we hit `Value`s, that will allow us to find `w`:
image

Then, we could ask `w` for its `Referrers`, which would allow us to find the call to `Fprintf`:
image

Asking whether `Fprintf` dominates the call to `Sink` would give us the answer we want.

I think this is more likely to scale than the approach using block indexes that I am proposing here. There might be problems that I don't see right now, but this is promising enough that I want to give it a shot.

Thanks! I'll try implementing this tomorrow morning and we'll see what happens.

PurelyApplied:
> Do you think it would be possible to leverage SSA's dominance to answer the question of whether a taint instruction dominates the sink instruction.

We are interested in the inverse, are we not?  If there exists a taint-path, regardless of whether or not that instruction dominates a sink call, we want to report it.

```golang
func Foo(s Source) {
  var st string
  if time.Now().Day == 0 {
    st := fmt.Sprintf("%v", s)
  }
  Sink(st)
}
```

I think dominance will only come into play when we're examining sanitization.

internal/pkg/source/source.go
```diff
@@ -266,3 +287,13 @@ func isProducedBySanitizer(v *ssa.Alloc, conf classifier) bool {
...
 }
+
+func indexInBlock(target ssa.Instruction) int {
+	for i, instr := range target.Block().Instrs {
+		if instr == target {
+			return i
+		}
+	}
+	// this cannot happen
+	return -1
```
PurelyApplied, line 298: Those are betting words.

If there is a contract of correct use, put it in a docstring.

I'm mildly opposed to sentinel values.  Consider using the return pair `(int, ok)` or `(int, error)` instead.

mlevesquedion: Hitting that `return -1` would mean that we've encountered a pretty severe bug in the SSA package, namely that an instruction isn't actually in the block it claims to be in.

The `ssa` package actually has a sanity checker that verifies that this does not happen (`ssa/sanity.go:320`), but it doesn't look like it gets called when using `buildssa`, so I modified the code to use the `(int, bool)` form.

mlevesquedion:
> The tainting of st does not dominate the sink instruction, but we still should report a Diagnostic. I think dominance will only come into play when we're examining sanitization.

It doesn't? Intuitively, if we've made it into the `if`, then we have no choice but to get through the `sink` call once we leave the `if`.

mlevesquedion:
If I tweak your example a little:
```
func Foo(s Source) {
  st := fmt.Sprintf("%v", s)
  if time.Now().Day() == 0 {
    Sink(st)
  }
}
```
The call to `Sink` here is not dominated by the tainting of `st`, so we would fail to produce a report.

With this in mind, I think I'm going to hold off until the next PR, where I'll look into this more deeply as I add test cases for multi-block functions.

mlevesquedion:
I think we could certainly be more careful with whether an edge is a `Referrers` or an `Operands` relationship. The edge type is already included in the graph, though. Doing it in `HasPathTo` would require some significant changes, since that method merely checks whether a node was visited during the traversal.

With the current set of tests, I don't think a more complex approach is justified. If we want to have a good confidence that whatever solution we come up with is working properly in the multi-block case, then we will need a lot of tests. I know this because when I tried to solve the full problem before opening this PR, I ended up writing a lot of extra tests that I thought would pass right away, but ended up failing and requiring me to change my approach. Of course implementing a more complex approach would also require more code.

Mostly I'm trying to avoid making this PR too big. I think handling swaps within a given block is a good start and I will be able to build on it in later PR's.

PR #66, opened 2020-08-26
Handle sources inlined in sink calls (and other similar cases)

mlevesquedion:
## The bug
Consider the following piece of code:
```go
func CreateSource() core.Source {
	return core.Source{}
}

func TestWrappedCall() {
	core.Sink(CreateSource()) // want "a source has reached a sink"
}

func TestNonWrappedCall() {
	s := CreateSource()
}
```

Certainly, it should not matter whether the call to `CreateSource` is inlined into the call to `core.Sink`, right? Unfortunately, it does. The reason why is plain to see in the SSA:

This is the inlined version:
```
0: entry
	0(*ssa.Call           ): t0 = CreateSource()
	...
	(instructions setting up the call to core.Sink)
	...
	6(*ssa.Call           ): t5 = example.com/core.Sink(t4...)
	7(*ssa.Return         ): return
```

This is the non-inlined version:
```
0: entry
	1(*ssa.Call           ): t1 = CreateSource()
	...
	(the remainder is identical to the inlined version)
```

(The inlined version does not have an Alloc instruction for the `Source`, therefore we do not detect the `Source` at all.)

Similar problems occur for the following when they are inlined:
- getting a source from an array by index
- getting a source from a map by key
- getting a source from a channel

There are probably other cases that I haven't though of.

## The fix
The fix I propose is simply to handle more types of value-creating instructions when we scan instructions in `source.go:sourcesFromBlocks`.

## Discussion
Some of these test cases are a little bit "weird" in the sense that it would be weird that a user would write such code. For example, it is unlikely that a user will call a source-producing function right into a call to a sink. However, such code is still incorrect and therefore we should produce a report for it.

- [x] Tests pass


internal/pkg/source/source.go
```diff
@@ -232,25 +233,49 @@ func sourcesFromBlocks(fn *ssa.Function, conf classifier) []*Source {
 
 		for _, instr := range b.Instrs {
 			switch v := instr.(type) {
```
mlevesquedion, line 235: The diff is really weird, I recommend looking at the code before/after in an editor.

internal/pkg/source/source.go
```diff
@@ -232,25 +233,49 @@ func sourcesFromBlocks(fn *ssa.Function, conf classifier) []*Source {
...
 			case *ssa.FieldAddr:
-				if conf.IsSource(utils.Dereference(v.Type())) {
-					sources = append(sources, New(v, conf))
+
+			// source obtained from array by index
+			case *ssa.IndexAddr:
+
+			// source obtained from map by key
+			case *ssa.Lookup:
+
```
PurelyApplied, line 259: Are these intentionally no-ops?

mlevesquedion: Yep. The intent is that in all other cases the `default` will get hit, which will cause the `for` to loop. Hitting one of the `case`s will lead to the code after the switch being executed. I did this because every case is pretty much the same otherwise (lots of duplicated code).

From your comment though I understand that this way of structuring the code way be misleading. Would you prefer that I duplicate the code? Or that I find some other way to avoid the duplication?

PurelyApplied: Ha!  I guess I'm acclimated to seeing the `default` at the bottom, I just banner-blindness'd right past it.  (It's fine where it is, just my own patterns.)

Stylistically, Go does allow multiple cases in-line.  You could consider collapsing the no-op cases

```golang
			switch v := instr.(type) {
			// Personally, I could go either way on collecting these by case, since it requires a type-assertion.  Just demonstrating here.
			case *ssa.Alloc, *ssa.Call:
				if isProducedBySanitizer(v.(ssa.Value), conf) {
					continue
				}

			// source obtained from field
			case *ssa.FieldAddr, *ssa.IndexAddr, *ssa.Lookup:
				// pass
```

The intent of the switch is to act as a filter, then?  "Get Allocs and Calls, unless they're from a sanitizer.  Get unary operator values other than channels.  Get these three.  Skip everything else."

Are we intentionally skipping `Value`s that don't fit the above?

All this to say, a docstring for reference on the function would be helpful now that it's grown in complexity.

internal/pkg/source/source.go
```diff
@@ -232,25 +233,49 @@ func sourcesFromBlocks(fn *ssa.Function, conf classifier) []*Source {
...
+			case *ssa.UnOp:
+				// not a <-chan operation
+				if v.Op != token.ARROW {
+					continue
 				}
 			}
+
+			// all of the above instructions are values as per ssa/doc.go
+			v := instr.(ssa.Value)
+			if conf.IsSource(utils.Dereference(v.Type())) {
```
PurelyApplied, line 270: Consider the more idiomatic

```golang
if v := instr.(ssa.Value); conf.IsSource(utils.Dereference(v.Type())) {
```

mlevesquedion: Argh. I always miss these opportunities. Good catch, thanks!

mlevesquedion:
For some reason I can't reply to your comment in-line.

> The intent of the switch is to act as a filter, then? "Get Allocs and Calls, unless they're from a sanitizer. Get unary operator values other than channels. Get these three. Skip everything else."
> Are we intentionally skipping Values that don't fit the above?

Yep! We may want to add more cases eventually, but for now these cover our tests and I can't think of other compelling cases. There might be the `Field` instruction, but I've never seen it in our ssa dumps (I do see plenty of `FieldAddr`s though). There might also be the `Extract` instruction, but in the tests I ran it was paired with other instructions that we already handle, so I don't think we need it.

> Stylistically, Go does allow multiple cases in-line. You could consider collapsing the no-op cases

Ah, very nice. I tried to do that one time, and the code wouldn't compile, so I thought type switches were somehow different from regular switches and you couldn't have many types in one `case`. Turns out, the error I got that time was due to the fact that when you use multiple types in a `case`, the type you get is the original type of the value you're switching on, so for `Alloc` and `Call`, the type was `ssa.Instruction`, and I was using it as an `ssa.Value`  in the body of the `case` (they are both `ssa.Value`s, but the compiler doesn't check for that). Anyhoo. Thanks, reads much better now!

internal/pkg/source/source.go
mlevesquedion: For some reason I can reply to your comment inline now.

>The intent of the switch is to act as a filter, then? "Get Allocs and Calls, unless they're from a sanitizer. Get unary operator values other than channels. Get these three. Skip everything else."
>Are we intentionally skipping Values that don't fit the above?

Yep! We may want to add more cases eventually, but for now these cover our tests and I can't think of other compelling cases. There might be the Field instruction, but I've never seen it in our ssa dumps (I do see plenty of FieldAddrs though). There might also be the Extract instruction, but in the tests I ran it was paired with other instructions that we already handle, so I don't think we need it.

>Stylistically, Go does allow multiple cases in-line. You could consider collapsing the no-op cases

Ah, very nice. I tried to do that one time, and the code wouldn't compile, so I thought type switches were somehow different from regular switches and you couldn't have many types in one case. Turns out, the error I got that time was due to the fact that when you use multiple types in a case, the type you get is the original type of the value you're switching on, so for Alloc and Call, the type was ssa.Instruction, and I was using it as an ssa.Value in the body of the case (they are both ssa.Values, but the compiler doesn't check for that). Anyhoo. Thanks, reads much better now!

>All this to say, a docstring for reference on the function would be helpful now that it's grown in complexity.

Good point. Done.

PR #67, opened 2020-08-27
Use facts in fieldtags analyzer

mlevesquedion:
This PR is a precursor to #68.

While integrating the `fieldtags` analyzer with the `source` analyzer, I ran into some issues that made me want to rework the `fieldtags` analyzer.

For one thing, currently the `fieldtags` analyzer uses strings to represent fields, which is not very robust. It is much simpler and more reliable to use `types.Object`s. We can get the `types.Object` for an `ast.Ident` node by using `pass.TypesInfo.ObjectOf`.

The rest of the code mirrors the pattern used in the `fieldpropagators` analyzer, namely:
1. while analyzing, export facts for tagged fields
2. when the main analysis is done, collect the exported facts and wrap them up into a result other analyzers can use.

**More fundamentally,** we need to have some way to carry information about tagged fields over the course of the analysis. Consider:
```go
package core

type Source struct {
  field string `levee:"source"`
}
```

```go
package usescore

import "core"

func main() {
  f := core.Source{}.field
}
```
Under the current implementation, while the `source` analyzer is running on the `usescore` package, it does not receive any tagged fields from the `fieldtags` analyzer, because the tagged fields are in `core`, not `usescore`.

Typically, analyzer `Result`s are the results for the analysis performed on a specific package. However, in this case we need to carry the knowledge of which tagged fields are sources to the importing packages. Otherwise, when a tagged field is accessed in an importing package, we have no way to tell.

It seems to me the most straightforward way to do this is to use `Fact`s. For additional context, see the `fieldpropagators` analyzer, in which we are already using this pattern.

- [x] Tests pass

### DIFF
--- a/internal/pkg/fieldtags/analyzer.go
+++ b/internal/pkg/fieldtags/analyzer.go
@@ -17,13 +17,10 @@
 package fieldtags
 
 import (
-	"fmt"
 	"go/ast"
 	"go/types"
 	"reflect"
 	"strings"
-
-	"github.com/google/go-flow-levee/internal/pkg/utils"
 
 	"golang.org/x/tools/go/analysis"
 	"golang.org/x/tools/go/analysis/passes/inspect"
@@ -42,15 +39,18 @@ var patterns sourcePatterns = []keyValue{
 	{"levee", "source"},
 }
 
-// TaggedFields is named type representing a slice of TaggedFields.
-type TaggedFields []TaggedField
-
-// ResultType is a slice of TaggedFields.
+// ResultType is a map telling whether an object is a tagged field.
 type ResultType = TaggedFields
 
-// A TaggedField contains the necessary information to identify a tagged struct field.
-type TaggedField struct {
-	pkg, typ, field string
+// TaggedFields is map telling whether an object is a tagged field.
+type TaggedFields map[types.Object]bool
+
+type isTaggedField struct{}
+
+func (i *isTaggedField) AFact() {}
+
+func (i *isTaggedField) String() string {
+	return "tagged field"
 }
 
 var Analyzer = &analysis.Analyzer{
@@ -61,11 +61,11 @@ var Analyzer = &analysis.Analyzer{
 		inspect.Analyzer,
 	},
 	ResultType: reflect.TypeOf(new(ResultType)).Elem(),
+	FactTypes:  []analysis.Fact{new(isTaggedField)},
 }
 
 func run(pass *analysis.Pass) (interface{}, error) {
 	inspectResult := pass.ResultOf[inspect.Analyzer].(*inspector.Inspector)
-	var taggedFields TaggedFields
 
 	nodeFilter := []ast.Node{
 		(*ast.TypeSpec)(nil),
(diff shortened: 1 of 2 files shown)


internal/pkg/fieldtags/analyzer.go
```diff
@@ -149,30 +148,7 @@ func (sp *sourcePatterns) isSource(field *ast.Field) bool {
...
-}
-
-func (t TaggedField) matches(typeName, fieldName string) bool {
-	return fmt.Sprintf("%s.%s", t.pkg, t.typ) == typeName && t.field == fieldName
-}
-
-func (t TaggedField) String() string {
-	return fmt.Sprintf("%s.%s.%s", t.pkg, t.typ, t.field)
+	// incantation plundered from the ssa package
+	field := f.X.Type().Underlying().(*types.Pointer).Elem().Underlying().(*types.Struct).Field(f.Field)
```
PurelyApplied, line 152: This seems prone to panic.  Are the assumptions in our use-case the same as those from wherever you plundered this?

internal/pkg/fieldtags/analyzer.go
```diff
@@ -61,11 +61,11 @@ var Analyzer = &analysis.Analyzer{
 		inspect.Analyzer,
 	},
 	ResultType: reflect.TypeOf(new(ResultType)).Elem(),
+	FactTypes:  []analysis.Fact{new(isTaggedField)},
```
PurelyApplied, line 64: Are you using `Fact`s just as an organizational / testing tool?  This imposes a import-ordering and doesn't buy us much otherwise.  What issues in #68 were you hitting that requires this?

internal/pkg/fieldtags/analyzer.go
```diff
@@ -82,17 +82,16 @@ func run(pass *analysis.Pass) (interface{}, error) {
...
-				taggedFields = append(taggedFields, tf)
-				pass.Reportf(f.Pos(), "tagged field: %s", tf)
+				pass.ExportObjectFact(pass.TypesInfo.ObjectOf(f.Names[0]), &isTaggedField{})
 			}
 		}
 	})
-	return taggedFields, nil
+
+	taggedFields := TaggedFields{}
+	for _, of := range pass.AllObjectFacts() {
```
PurelyApplied, line 91: I believe this will also return the objects for any imported dependency, which do not seem useful to downstream analyzers.  See about about `Fact`s in general.

internal/pkg/fieldtags/analyzer.go
mlevesquedion: We need to have some way to carry information about tagged fields over the course of the analysis. Consider:
```
package core

type Source struct {
  field string `levee:"source"`
}
```

```
package usescore

import "core"

func main() {
  f := core.Source{}.field
}
```

Under the current implementation, while the `source` analyzer is running on the `usescore` package, it does not receive any tagged fields from the `fieldtags` analyzer, because the tagged fields are in `core`, not `usescore`.

Typically, analyzer `Results` are the results for the analysis performed on a specific package. However, in this case we need to carry which tagged fields are sources to the importing packages. Otherwise, when a tagged field is accessed in an importing package, we have no way to tell.

It seems to me the most straightforward way to do this is to use `Fact`s. We are using this pattern in the `fieldpropagators` analyzer. The way it is being done right now is rather inefficient. It would be more efficient to use `PackageFact`s instead of `ObjectFact`s. That way we could bundle up all the tagged fields into a slice or map, and moving it forward from importee to importer would be very efficient. Right now I don't see this hindering performance significantly, and without a benchmark I think it is preferable to use `ObjectFact`s since the code is simpler that way.

internal/pkg/fieldtags/analyzer.go
mlevesquedion: I think they are. The way it's documented seems to suggest that this is the canonical way to get the field, so I would expect it not to panic.

I plundered this from the documentation for `ssa.FieldAddr`. I've made the comment more specific in that regard. I could extract this to a function that checks every step of the way to ensure a panic cannot occur.

mlevesquedion:
Closing this PR as I will integrate with `source` through `sourcetype`, so this change is no longer needed.

PR #68, opened 2020-08-27
Integrate fieldtags analyzer with source analyzer

mlevesquedion:
This PR is built on #67.

This PR integrates the `fieldtags` analyzer into the `source` analyzer, allowing identification of tagged fields as sources.

- [x] Tests pass

### DIFF
--- a/internal/pkg/fieldtags/analyzer.go
+++ b/internal/pkg/fieldtags/analyzer.go
@@ -17,13 +17,10 @@
 package fieldtags
 
 import (
-	"fmt"
 	"go/ast"
 	"go/types"
 	"reflect"
 	"strings"
-
-	"github.com/google/go-flow-levee/internal/pkg/utils"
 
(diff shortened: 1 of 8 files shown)


PurelyApplied:
Would it make more sense to link the `fieldtags` analyzer into the `sourceTypes` analyzer?  I know the `sourceTypes` analyzer still needs to be consumed by `source`, but that seems like a more streamlined path.

mlevesquedion:
I think it could, yes. In that case my proposed plan of action would be:
1. Use `fieldtags` analyzer in `sourceTypes` analyzer.
2. Use `sourceTypes` analyzer in `source` analyzer.

This does mean however that it will no longer be necessary for the `fieldtags` analyzer to use `Fact`s, because the `sourcetype` analyzer will be able to use its `Result`s directly to create its own `Fact`s.

mlevesquedion:
Closing this PR as I will integrate with `source` through `sourcetype`.

PR #70, opened 2020-08-27
Prepare fieldtags analyzer for integration with sourcetype analyzer

mlevesquedion:
This PR is a precursor to #71.

This PR prepares integration with the `sourcetype` analyzer by modifying the `fieldtags` analyzer to work with `types.Object`s. This is useful because the `sourcetype` analyzer works with `types.Var`s, which are `types.Object`s.

Working with `types.Object`s also makes the `fieldtags` analyzer more robust. Currently, it is working with strings, and in fact it contains a bug: it uses `Pkg.Name()` instead of `Pkg.Path()` when recording the string that represents a field's package. Using `types.Object`s makes the code simpler and more likely to be correct.

- [x] Tests pass

### DIFF
--- a/internal/pkg/fieldtags/analyzer.go
+++ b/internal/pkg/fieldtags/analyzer.go
@@ -39,12 +39,9 @@ var patterns sourcePatterns = []keyValue{
 	{"levee", "source"},
 }
 
-// TaggedFields is a map from types.Object to bool.
+// ResultType is a map from types.Object to bool.
 // It can be used to determine whether a field is a tagged Source field.
-type TaggedFields map[types.Object]bool
-
-// ResultType is a slice of TaggedFields.
-type ResultType = TaggedFields
+type ResultType map[types.Object]bool
 
 var Analyzer = &analysis.Analyzer{
 	Name: "fieldtags",
--- a/internal/pkg/fieldtags/analyzer.go
+++ b/internal/pkg/fieldtags/analyzer.go
@@ -143,7 +143,7 @@ func (sp *sourcePatterns) isSource(field *ast.Field) bool {
 
 // IsSourceFieldAddr determines whether a ssa.FieldAddr is a source, that is whether it refers to a field previously identified as a source.
 func (t TaggedFields) IsSourceFieldAddr(fa *ssa.FieldAddr) bool {
-	// incantation plundered from the documentation for ssa.FieldAddr
+	// incantation plundered from the docstring for ssa.FieldAddr.Field
 	field := fa.X.Type().Underlying().(*types.Pointer).Elem().Underlying().(*types.Struct).Field(fa.Field)
 	return t.IsSource(field)
 }
--- a/internal/pkg/fieldtags/analyzer.go
+++ b/internal/pkg/fieldtags/analyzer.go
@@ -17,13 +17,10 @@
 package fieldtags
 
 import (
-	"fmt"
 	"go/ast"
 	"go/types"
 	"reflect"
 	"strings"
-
-	"github.com/google/go-flow-levee/internal/pkg/utils"
 
 	"golang.org/x/tools/go/analysis"
 	"golang.org/x/tools/go/analysis/passes/inspect"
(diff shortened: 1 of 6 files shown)


internal/pkg/fieldtags/analyzer.go
```diff
@@ -147,32 +138,14 @@ func (sp *sourcePatterns) isSource(field *ast.Field) bool {
...
-	deref := utils.Dereference(structType)
-	named, ok := deref.(*types.Named)
-	return named, ok
-}
-
-func (t TaggedField) matches(typeName, fieldName string) bool {
-	return fmt.Sprintf("%s.%s", t.pkg, t.typ) == typeName && t.field == fieldName
+// IsSourceFieldAddr determines whether a ssa.FieldAddr is a source, that is whether it refers to a field previously identified as a source.
+func (t TaggedFields) IsSourceFieldAddr(fa *ssa.FieldAddr) bool {
+	// incantation plundered from the documentation for ssa.FieldAddr
```
mlevesquedion, line 143: Carrying over a thread from a previous PR: would you like me to extract this to a function that makes all the intermediate checks to make sure that this can never panic?

(Since I took it straight from the documentation I would expect it to be fine.)

PurelyApplied: I saw it's from the docstring on `ssa.FieldAddr`.  So I think the expression is fine, but I'd update `incantation plundered from` to explicitly point to `ssa.FieldAddr.Field docstring`

mlevesquedion: SG. Done.

internal/pkg/fieldtags/analyzer.go
```diff
@@ -42,17 +39,13 @@ var patterns sourcePatterns = []keyValue{
 	{"levee", "source"},
 }
 
-// TaggedFields is named type representing a slice of TaggedFields.
-type TaggedFields []TaggedField
+// TaggedFields is a map from types.Object to bool.
+// It can be used to determine whether a field is a tagged Source field.
+type TaggedFields map[types.Object]bool
```
PurelyApplied, line 44: I maybe commented on the last one, but I don't see why we need to export both `TaggedFields` and `ResultType`.  I have no opinion about which it should be.

mlevesquedion: I find having a `ResultType` symbol is useful when you want to get the analyzer's result, e.g. `pass.ResultOf[fieldtags.Analyzer].(fieldtags.ResultType)`.

Apart from that I agree that it's not super useful, and may contribute to noise. Also it duplicates documentation, and in this case it also led to the documentation for `ResultType` being out of date.

I've removed `TaggedFields`. It's not super useful to have two aliases, and I think `ResultType` is the more useful of the two.

Since this "pattern" was also in use in the fieldpropagators analyzer, I've made a PR for that one too, for consistency: #78

internal/pkg/fieldtags/analyzer.go
```diff
@@ -82,17 +75,15 @@ func run(pass *analysis.Pass) (interface{}, error) {
...
 			if patterns.isSource(f) && len(f.Names) > 0 {
-				tf := TaggedField{
-					pkg:   pass.Pkg.Name(),
-					typ:   t.Name.Name,
-					field: f.Names[0].Name,
-				}
-				taggedFields = append(taggedFields, tf)
-				pass.Reportf(f.Pos(), "tagged field: %s", tf)
+				fName := f.Names[0]
+				obj := pass.TypesInfo.ObjectOf(fName)
```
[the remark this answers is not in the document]
mlevesquedion: When I first saw this "Names" business I assumed it had to do with some implementation decision, e.g. future-proofing. Turns out this was a bad assumption to make.

From the spec:
`FieldDecl     = (IdentifierList Type | EmbeddedField) [ Tag ] .`
`IdentifierList = identifier { "," identifier } .`

So, today I learned that Go fields can have more than one name. This is valid Go:
```
type Source struct {
    password, secret string
}
```

This field has 2 names, each with its own `types.Object`. 

I've modified the code to handle all the names, and I've modified the tests.

PR #73, opened 2020-08-28
Avoid future json errors

mlevesquedion:
This PR aims to avoid typos such as #72 from happening again by validating the JSON files in the repository using `jq`.

While the existing `go test ./...` check already validates that all of the configs used in testing are properly formed, any other JSON files are not currently being validated.

- [x] Tests pass


travis_scripts/check_json.sh
```diff
@@ -0,0 +1,14 @@
...
+failed=false
+while read -r -d '' f; do 
+	if ! OUT=$(jq type "$f"); then
+		echo "$OUT"
+		failed=true
+	fi
+done < <(find . -iname '*.json' -print0)
+if $failed; then
+	exit 1
+fi
```
PurelyApplied, line 13: This feels rather opaque.  Consider
```bash
set -e
for f in $(find . -name '*.json'); do
  jq type $f
done
```

mlevesquedion: Hmm. Are you suggesting that as a replacement for the whole thing, or are you suggesting using `set -e` + `for loop` instead of a `while` loop? That's what I wrote initially (minus the `set -e`), but I didn't want to pollute the CI stdout with `jq`'s output (when `jq type` executes correctly on our json files, it outputs `"object"`). If that's not an issue, then I agree that the script can be simplified quite a lot.

I turned to a `while` loop because `shellcheck` told me the following:
`For loops over find output are fragile. Use find -exec or a while read loop. [SC2044]`.

PurelyApplied: > Are you suggesting that as a replacement for the whole thing

That 4-liner would replace the whole thing.

>  didn't want to pollute the CI stdout with jq's output

That's fair, but that's easily done with a `jq type $f &> /dev/null`

> I turned to a while loop because shellcheck told me the following:

Ehhhh.  The linter isn't wrong.  `find` output is fragile if one allows spaces in filenames.  We don't have any yet.  But the linter isn't wrong...

`find . -name '*.json' -exec jq {} +` reads pretty clearly though.  Redirect to `/dev/null` to suppress output.

Or whatever.

</bikeshed>

mlevesquedion: Thanks!

Great point on the redirect there.

I like the one-liner. I've made the change.

PR #76, opened 2020-08-28
Only avoid Source methods when traversing from a Source

mlevesquedion:
## Description

This PR is a successor to #75. See that PR for additional context.

In short, currently we cannot find sink calls when the sink is a method, because in `source.go:visitReferrers`, we avoid all methods.

This PR rectifies that by avoiding only methods where the receiver is a `Source`.

## Discussion

While I think this is a good bug fix, in some cases we will need to visit `Source` methods if we want to do a good job with propagation. For example, currently this case does not lead to a report:
```
s := core.Source{}
str := s.String()
core.Sink(str)
```
Yet depending on what the `String` method does (i.e., whether it returns tainted data or not), we may want to consider it to be a propagator. I think the most natural way to handle this right now would be to extend the `fieldpropagators` analyzer to handle not just getters and getter-like methods, but to actually perform taint propagation analysis within the bodies of `Source` methods to determine whether we should traverse or not. Hopefully this can be implemented by reusing existing code, although I suspect that this will not be simple to do and that some refactoring will need to be done. On a much longer horizon, all of this could eventually be handled via cross-function analysis.

Currently, `fieldpropagators` are handled in `levee.go`, but I feel like it might make more sense to have that logic in the traversal code in `source.go`, since field propagators are a subset of `Source` type methods.

- [x] Tests pass

### DIFF
--- a/internal/pkg/source/source.go
+++ b/internal/pkg/source/source.go
@@ -17,6 +17,7 @@ package source
 
 import (
 	"fmt"
+	"go/token"
 	"go/types"
 	"strings"
 
@@ -252,6 +253,7 @@ func sourcesFromClosure(fn *ssa.Function, conf classifier) []*Source {
 	return sources
 }
 
+// sourcesFromBlocks finds Source values created by instructions within a function's body.
 func sourcesFromBlocks(fn *ssa.Function, conf classifier) []*Source {
 	var sources []*Source
 	for _, b := range fn.Blocks {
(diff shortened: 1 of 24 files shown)


internal/pkg/source/source.go
```diff
@@ -106,9 +106,10 @@ func (a *Source) visitReferrers(referrers *[]ssa.Instruction) {
 		case *ssa.Call:
 			// This is to avoid attaching calls where the source is the receiver, ex:
 			// core.Sinkf("Source id: %v", wrapper.Source.GetID())
-			if v.Call.Signature().Recv() != nil {
+			if recv := v.Call.Signature().Recv(); recv != nil && a.config.IsSource(utils.Dereference(recv.Type())) {
 				continue
```
PurelyApplied, line 110: For the case `GetData()` rather than `GetID()`, are we explicitly superseding this elsewhere?

mlevesquedion: Yeah, we're using the `fieldpropagators` analyzer's result in `levee.go:61`:
```
				case fieldPropagators.Contains(v):
					sources = append(sources, source.New(v, conf))
```

I think it would make more sense to use the `fieldpropagator` results in the `source` analyzer, WDYT?

PurelyApplied: That sounds good to me.

PR #79, opened 2020-09-02
Fix bugs in cross-block propagation

mlevesquedion:
## Context

This PR is a successor to #63 (merged).

## This PR

This PR fixes propagation bugs in functions that have multiple blocks in their bodies. Currently, there are several cases where we are producing incorrect diagnostics, e.g.:
```
if true {
    fmt.Fprintf(w, "%v", s)
} else {
    core.Sink(w) // a report is produced here
}
```
Clearly, there is no way that both instructions will be executed, so we shouldn't produce a report.

We already have some tests where the body has more than one block (e.g. some tests in the `dominance` folder), but these new tests focus explicitly on instruction order and block order, and they cover cases not covered in other tests.

The fix I propose is to only visit a `Referrer` instruction if ...
1. It is in the same block as the last visited instruction.
2. It is in a block that is reachable from the last visited block, i.e. it could conceivably be reached in an execution of the function.

This implies that:
1. We need to keep track of the last visited block (trivial)
2. We need to be able to determine whether a block is reachable from another block (in this PR, using DFS on the SSA blocks graph)
## Additional details
For call instructions, we require that the instruction of the call is the highest instruction reached so far in its block (see `source.go:117-126`). For non-call instructions, we need to be more lenient. Indeed, consider the following SSA:
```
1: if.then
	0(*ssa.Alloc          ): t1 = new [1]interface{} (varargs)
	1(*ssa.IndexAddr      ): t2 = &t1[0:int]
	2(*ssa.ChangeInterface): t3 = change interface interface{} <- io.Writer (w)
	3(*ssa.Store          ): *t2 = t3
	4(*ssa.Slice          ): t4 = slice t1[:]
	5(*ssa.Call           ): t5 = example.com/core.Sink(t4...)
	6(*ssa.Jump           ): jump 2
```
In the graph, when visiting from `w` the order of instructions visited is actually: `ChangeInterface --> Store --> IndexAddr --> Alloc --> Slice --> Call`. (I would drop a screenshot, but currently my screenshot tool is broken).

- [x] Tests pass


internal/pkg/source/source.go
```diff
@@ -72,7 +73,7 @@ func (a *Source) dfs(n ssa.Node) {
 	a.marked[n.(ssa.Node)] = true
 
 	if instr, ok := n.(ssa.Instruction); ok {
-		a.recordIndex(instr)
+		a.record(instr)
 	}
 
 	if n.Referrers() != nil {
```
immutableT, line 79: What if we removed referrers that are in the "past" here?
Would this do the trick? Would this read better?

internal/pkg/source/source.go
```diff
@@ -130,10 +132,38 @@ func (a *Source) visitReferrers(referrers *[]ssa.Instruction) {
 				continue
 			}
 		}
+
+		// If the referrer is in a different block from the one we last visited,
+		// and it can't be reached from the block we are visiting, then stop visiting.
+		rb := r.Block()
+		if a.lastBlockVisited != nil && rb != a.lastBlockVisited && !a.canReach(a.lastBlockVisited, rb) {
```
PurelyApplied, line 139: Idioms.

internal/pkg/source/source.go
```diff
@@ -130,10 +132,38 @@ func (a *Source) visitReferrers(referrers *[]ssa.Instruction) {
...
+
 		a.dfs(r.(ssa.Node))
 	}
 }
 
+func (a *Source) canReach(start *ssa.BasicBlock, dest *ssa.BasicBlock) bool {
+	stack := []*ssa.BasicBlock{start}
+	seen := map[*ssa.BasicBlock]bool{start: true}
+	for len(stack) > 0 {
+		current := stack[len(stack)-1]
```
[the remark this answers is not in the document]
mlevesquedion: That's true, but it is defined like so: `type stack []ssa.Node`, so it can't be used here. I could introduce a new type that would be defined like `type stack []*ssa.BasicBlock`. The definitions for its methods would be identical to those of the existing `stack` type, though.

Alternatively, I could write a generic `stack`, but I feel like that wouldn't be a good ROI. We've had this discussion in the past, but I don't think having a dedicated `stack` type is a big win.

internal/pkg/source/source.go
mlevesquedion: I think it could be done, but I'm not sure it would read better. We would need to create a new slice to hold the referrers we want to keep.

In fact, I think we might want to move the `n.Referrers()  != nil` check into `visitReferrers`.

Also, your comment made me realize that the code for skipping referrers that are "in the past" was placed in a weird location in `visitReferrers`, so I moved it. Do you think it's better now?

internal/pkg/source/source.go
mlevesquedion: Sure. The line is already very long though, and for me that makes it a little hard to read. Do you think I should extract part of the conditional?

internal/pkg/source/source.go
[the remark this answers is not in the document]
immutableT: Up to you.
IMHO, implementing generic algorithms (inline) within a function that is clearly domain specific (SSA) increases the cognitive overhead for review. So, I would try to hide this.

internal/pkg/source/source.go
PurelyApplied: That would work.  You can also lean on the operators and space with newlines if you want, e.g.

```golang
if rb := r.Block(); a.lastBlockVisited != nil &&
    rb != a.lastBlockVisited &&
    !a.canReach(a.lastBlockVisited, rb) {
```

internal/pkg/source/source.go
immutableT: Yes, this reads better.

You could expose a property on the `Source` type that returns referrers, and this property will be aware of the "past" and "future" concepts.

internal/pkg/source/source.go
PurelyApplied: Also, I didn't meant to sound glib in the first comment, I just know you can be prone to overlook the `if`-declaration idiom.

mlevesquedion: It's okay.

Yeah you're right, in this case I think it was because of the line length, so thanks for suggesting splitting on the `&&`s.

I've been putting off writing a linter for this, but I guess I'll need to get to it :upside_down_face:

internal/pkg/source/source.go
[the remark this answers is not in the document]
mlevesquedion: SG, I've duplicated the `stack` code and modified `canReach` to use it.

internal/pkg/source/source.go
mlevesquedion: Cool.

I think that is a good idea. I've performed the refactoring, and it definitely reads better. The `visitReferrers` function was doing too many things. The refactoring is pretty big relative to the code that really needed to change in this PR though. Let me know if you want me to extract it to its own PR.

immutableT (approved):

PR #80, opened 2020-09-02
Fix false positives due to propagation through extracts

mlevesquedion:
## The bug
While running the analyzer on kubernetes, I got a false positive report for an error value that was returned from a function that returns two values: a `Source` value and an `error`. Here is an abstract example:
```go
func CreateSource() (core.Source, error) {
	return core.Source{}, nil
}

func TestExtracts() {
	s, err := CreateSource()
	core.Sink(err) // unexpected diagnostic here
	_ = s
}
```
Just because the `error` value is returned next to a `Source`, that doesn't mean it should be tainted.

## The problem

The core of the problem is that we are traversing to nodes that we shouldn't be traversing to. Consider the following piece of an SSA graph:
image

While scanning the instructions, the code in `source.go:sourcesFromBlocks` will find the `Alloc` and create a `Source` from it, which will start a DFS. The DFS reaches the `Extract`, which is used to extract the `Source` from the tuple created that is returned by the `CreateSource` function. The problem is that with the current call, the traversal continues through the `Call`, and into the other `Extract` (not shown in the image), which leads to the call to `core.Sink`, and therefore an unexpected diagnostic.

## The fix

The fix is simply to avoid traversing to a `Call` from an `Extract`. The tainted `Extract` will have the `Source`-producing call as an `Operand`, which just need to avoid it.

## Additional details

The tests cover a variety of failure cases that I ran into while developing this fix.

The last test case documents a logical extension of current behavior, namely that if a function takes a `Source` as argument, we consider its return value(s) to be tainted.

- [x] Tests pass

### DIFF
--- a/internal/pkg/levee/testdata/src/example.com/tests/extracts/tests.go
+++ b/internal/pkg/levee/testdata/src/example.com/tests/extracts/tests.go
@@ -22,7 +22,7 @@ func CreateSource() (core.Source, error) {
 	return core.Source{}, nil
 }
 
-func CreateSourceRetvalsFlipped() (error, core.Source) {
+func CreateSourceFlipped() (error, core.Source) {
 	return nil, core.Source{}
 }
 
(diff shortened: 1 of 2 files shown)


PurelyApplied (changes requested):
Good bug identification.  However, you are currently a bit too general in what traversal steps to skip.

The following tests should be added.  They currently fail.

* Return a reference rather than a concrete source causes the returned sources to be ignored
```golang
func CreateSource() (*core.Source, error) {
	return &core.Source{}, nil
}
//     analysistest.go:399: example.com/tests/extracts/tests.go:35: no diagnostic was reported matching "a source has reached a sink"
//    analysistest.go:399: example.com/tests/extracts/tests.go:42: no diagnostic was reported matching "a source has reached a sink"
FAIL
```

* All entries of a returned tuple should be tainted if called on a source method.
```golang
// core.go
func (s Source) Copy() (Source, error) {
	if time.Now().Weekday() == time.Sunday {
		return Source{}, fmt.Errorf("no copying on Sundays; have the original: %v", s)
	}

	return Source{
		Data: s.Data,
		ID: s.ID,
	}, nil
}

// test.go

func TestReceiverShouldTaintBoth() {
	s := core.Source{Data: "secret"}
	cpy, err := s.Copy()
	core.Sink(cpy) // emits only if the returned type is Source, but not if *Source or interface{}
	core.Sink(err) // does not currently emit
}
```

mlevesquedion:
These are good test cases, but I believe they are caused by a different issue.

The first one is caused by the fact that no `Source` is identified from the instructions. Because the call returns a pointer, no local copy (`Alloc`) is made, so we never catch it. We might've caught it through our `case *ssa.Call:` in `source.go:sourcesFromBlocks`, but because what's returned is a `tuple`, not directly a `Source`, we don't catch it. Here's the SSA graph:

image

The second one is also caused by the previous observation, along with the fact that we don't visit through `Source` methods. So when the returned value is a `Source`, not a `*Source`, we don't DFS through the method call, but there `is` an `Alloc` for the returned `Source`, so we find the sink. If the returned value is a `*Source` though, we don't detect it.

So in both cases, the fix is we need better `Source` identification. I think that belongs to its own PR, along with additional tests. WDYT?

PurelyApplied (approved):

PR #81, opened 2020-09-03
Fix: identify sources in extracts

mlevesquedion:
This PR is a successor to #80. PR #80 addresses false positives caused by traversing from a tainted `Extract` to the other `Extract`s of a `Call`.

This PR addresses `Extract`s not being identified as `Source`s due to being returned by a call that returns a `*Source` instead of a `Source`.

See #80, as well as my comment here: https://github.com/google/go-flow-levee/pull/80#issuecomment-686506245, for additional information.

- [x] Tests pass

### DIFF
--- a/internal/pkg/levee/testdata/src/example.com/core/source.go
+++ b/internal/pkg/levee/testdata/src/example.com/core/source.go
@@ -28,6 +28,14 @@ func (s Source) GetData() string {
 	return s.Data
 }
 
+func (s Source) Copy() (Source, error) {
+	return s, nil
+}
+
+func (s Source) CopyPointer() (*Source, error) {
+	return &s, nil
+}
+
 // Innocuous will _not_ be configured to be a source, even though underlying types are equal.
 type Innocuous struct {
 	Data string
--- a/internal/pkg/levee/testdata/src/example.com/tests/extracts/tests.go
+++ b/internal/pkg/levee/testdata/src/example.com/tests/extracts/tests.go
@@ -22,7 +22,7 @@ func CreateSource() (core.Source, error) {
 	return core.Source{}, nil
 }
 
-func CreateSourceRetvalsFlipped() (error, core.Source) {
+func CreateSourceFlipped() (error, core.Source) {
 	return nil, core.Source{}
 }
 
@@ -42,19 +42,19 @@ func TestOnlySourceExtractIsTaintedInstructionOrder() {
 	core.Sink(s) // want "a source has reached a sink"
 }
 
+func TestOnlySourceExtractIsTaintedFlipped() {
+	err, s := CreateSourceFlipped()
+	core.Sink(s) // want "a source has reached a sink"
+	core.Sink(err)
+}
+
 func TestOnlySourceExtractIsTaintedInstructionOrderFlipped() {
-	err, s := CreateSourceRetvalsFlipped()
-	core.Sink(s) // want "a source has reached a sink"
-	core.Sink(err)
-}
-
-func TestExtractsFlipped() {
-	err, s := CreateSourceRetvalsFlipped()
+	err, s := CreateSourceFlipped()
 	core.Sink(err)
 	core.Sink(s) // want "a source has reached a sink"
 }
 
-func TestExtractsFromCallWithSourceArg(s core.Source) {
+func TestExtractsFromCallWithSourceArgAreTainted(s core.Source) {
 	str, i, e := TakeSource(s)
 	core.Sink(str) // want "a source has reached a sink"
 	core.Sink(i)   // want "a source has reached a sink"
--- a/internal/pkg/levee/testdata/src/example.com/tests/extracts/tests.go
+++ b/internal/pkg/levee/testdata/src/example.com/tests/extracts/tests.go
@@ -60,3 +60,27 @@ func TestExtractsFromCallWithSourceArgAreTainted(s core.Source) {
 	core.Sink(i)   // want "a source has reached a sink"
 	core.Sink(e)   // want "a source has reached a sink"
 }
+
+func NewSource() (*core.Source, error) {
+	return &core.Source{}, nil
+}
+
+func TestNewSource() {
+	s, err := NewSource()
+	core.Sink(s) // want "a source has reached a sink"
+	core.Sink(err)
+}
+
+func TestCopy() {
+	s := core.Source{}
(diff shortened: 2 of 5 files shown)


PurelyApplied (changes requested):
If this is a bug involving source identification, tests should be use the source identifying analyzer, not end-to-end tests.

mlevesquedion:
I agree that there should be tests directed at the `source` analyzer, but I think the e2e tests are valuable as well because they document the cases the analyzer as a whole can handle.

I'll add tests for Extracts.

I will need to add more tests in a future PR since this code has changed significantly in recent times, e.g. #66.

PurelyApplied:
I'm not saying the e2e tests don't have value, but this PR needs tests closer to the bone of the issue.

mlevesquedion:
Agreed. This is going to be interesting to test because it looks like `Extract`s don't have `Pos`itions. I'll let you know when the tests are done.

mlevesquedion:
I've added the tests.

Since `Extract`s don't have a position in the source code, I use the `Extract`'s related `Call` when producing the report.

(Why do `Extract`s not have a position? I don't know. All I know is that `Extract`s embed a `register`, which is what holds the `Pos`. The default value for a `Pos` is `NoPos` (0). This value is never set for extracts, so their `Pos` functions returns `NoPos`.)

mlevesquedion:
I've though about this some more, and since `Extract`s don't have a position, that means the report we produce in levee is going to be rather unhelpful. It would be easy to report the `Call`'s position instead, and I think that is the best we can do with the SSA.

To implement this, I would write a `Pos` method on `Source` that would deal with these details. This would be in a future PR.

Should I add positions to the `want` comments? This could also be in that future PR.

WDYT?

PurelyApplied:
\*cough\*

Extract calls it `Index`.  [Docs](https://godoc.org/golang.org/x/tools/go/ssa#Extract).

mlevesquedion:
Uh... do you mean that for `Extract`s the `Pos` is in their `Index` field? If so, that's incorrect. The `Index` is the position of the extracted value in the tuple that was returned by the associated `Call`.

The `Pos` that I'm talking about is the `token.Pos`.

PurelyApplied:
Ahh, my mistake.  I did think you were lamenting not having the position of the `ssa.Extract`.  I didn't shift package gears with you.

In which case, the `Extract.Value.Pos()` should yield what you want, yes?

mlevesquedion:
Yep, that's what I'm doing. It's not quite the best position for reporting though.

I was saying `Call` to refer to `Tuple`, because in the case that I was considering the `Tuple` is a `Call`, but `Extracts` can be produced by other instructions as well.

PurelyApplied (commented):
To avoid letting this PR languish much longer without a real review...

I am conflicted.  While this `value, err` or `value, ok` pattern is extremely common, I'm not sure how I feel about ubiquitously claiming that neighboring values should or should not be tainted by a nearby source.

In particular, I think
```golang
func TestCopy() {
	s := core.Source{}
	cpy, err := s.Copy()
	core.Sink(cpy) // want "a source has reached a sink"
	core.Sink(err)
}
```

As a method on a receiver of source type, I would expect all arguments and return values should become tainted.  While I agree that taint shouldn't propagate from `cpy`, I would expect taint to propagate from `s`.

Conversely, in a slightly modified
```golang
func TestNewSource() {
	s, err := NewSource("my-secret")
	core.Sink(err)
}
```
I don't know that there's any reason I should believe that the `err` didn't contain the secret passed to the "constructor."

::rubber duck::

Maybe that's what's bothering me.  If the `"my-secret"` portion were identified as a source-input, in the same way that the receiver in `s.Copy()` should be identified as a source-input, then I would expect (a) the source return argument to be marked as a source, and (b) the neighboring `err` to be be tainted by the source-input.

So maybe I would be okay with the `err` being clean in `s, err := NewSource()` case, but not a `cpy, err := s.Copy()` or `cpy, err:= Duplicate(s)` cases.

mlevesquedion:
> While I agree that taint shouldn't propagate from `cpy`, I would expect taint to propagate from `s`.

I'm not sure I understand what you mean here. Did you mean taint shouldn't propagate from `err`? I think `cpy` should be marked as a `Source`. (We would need special handling here in the case where a `Source` is sanitized before being copied).

> As a method on a receiver of source type, I would expect all arguments and return values should become tainted.

This would be easy to implement, but I'm concerned it would lead to lots of false positives?

> If the "my-secret" portion were identified as a source-input ...

I think we could do this with Value specifiers (referring to the 1.0 discussion here). However, again I'm concerned that this might produce lots of false positives.

I think ultimately we will need interprocedural analysis to handle these cases properly. If we analyze `Copy` and `NewSource` and find that the `error` value they return is not tainted, then we don't mark the `error` as a `Source` or propagate taint it when visiting a `Call` to `Copy` or `NewSource`. However, if it is tainted, then we should propagate taint.

My goal with this PR was to identify `Sources` that currently aren't being identified simply because they are produced by a function that has more than one return value. The way we should handle `error`s returned by such functions isn't clear to me right now, and as such I think it should be done in a separate PR.

WDYT?

PurelyApplied:
> I'm not sure I understand what you mean here. Did you mean taint shouldn't propagate from `err`? I think `cpy` should be marked as a `Source`. (We would need special handling here in the case where a `Source` is sanitized before being copied).

```golang
func TestCopy() {
	s := core.Source{}
	cpy, err := s.Copy()
	core.Sink(cpy) // want "a source has reached a sink"
	core.Sink(err)
}
```

I think `cpy` should be correctly identified as a source.  I think taint should propagate _from `s`_ to `err`, and in turn `core.Sink(err)` should emit a Diagnostic.  I see no reason for us to have confidence that `err` is untainted.

> This would be easy to implement, but I'm concerned it would lead to lots of false positives?

I am more greatly concerned with false negatives, personally.  There is no evidence (at present, sans interprocedural analysis) that there is no propagation of taint from any call that accepts a source as a parameter.

Ultimately, there is nothing but syntactic sugar distinguishing `s.Copy()` from an arbitrary `Foo(s)`, and in the latter case, I would expect any returned value to be tainted.  

> My goal with this PR was to identify `Sources` that currently aren't being identified [...]
Good point.  My concern is definitely beyond the scope of this PR.  I'll open an Issue.

PurelyApplied (approved):

PR #82, opened 2020-09-03
Filter referrers before visiting

mlevesquedion:
This PR is a successor to #79.

This PR refactors the code involved in visiting referrers in order to make it easier to read and to extend.

Recent PR's have made this code more complicated, so I think this refactoring is necessary.

- [x] Tests pass


internal/pkg/source/source.go
```diff
@@ -100,47 +98,68 @@ func (a *Source) record(target ssa.Instruction) {
...
+	if n.Referrers() == nil {
+		return
+	}
+	for _, r := range *n.Referrers() {
+		// If the referrer is in a different block from the one we last visited,
+		// and it can't be reached from the block we are visiting, then stop visiting.
+		if rb := r.Block(); a.lastBlockVisited != nil &&
+			rb != a.lastBlockVisited &&
+			!a.canReach(a.lastBlockVisited, rb) {
+			continue
```
PurelyApplied, line 140: This subsumes the check in `visitReferrers`, which is not unreached by test code.  (I can't comment there, though, since the line is unchanged.  Thanks GitHub.)

internal/pkg/source/source.go
```diff
@@ -100,47 +98,68 @@ func (a *Source) record(target ssa.Instruction) {
...
 			}
+		}
 
-			if a.config.IsSanitizer(v) {
-				a.sanitizers = append(a.sanitizers, &sanitizer.Sanitizer{Call: v})
+		a.dfs(r.(ssa.Node))
+	}
+}
+
+func (a *Source) referrersToVisit(n ssa.Node) (referrers []ssa.Instruction) {
```
PurelyApplied, line 130: Consider adding a summary docstring for what we're skipping.

PurelyApplied (changes requested):

internal/pkg/source/source.go
mlevesquedion: Good catch! I've removed the check from `visitReferrers`.

internal/pkg/source/source.go
mlevesquedion: Done.

PurelyApplied (approved):

Issue #96, opened 2020-09-10
Infer from a source type any defined type using the source type as the underlying type


PurelyApplied:
Let's deconstruct that Issue title.  Consider the following:

```golang

type Source struct {
  secret string // `levee:"source"`
}

type Foo Source
```

The type `Source` will be identified as a source due to the field tag.  The type `Foo` has the underlying type `Source`, and should also be identified as a source type without explicit configuration.

In this instance, the underling struct literal contains a field tag, which takes part of type identity.  As such, `Foo.secret` also contains the field tag and will be identified as a source type.

However, consider instead 

```golang

type Source struct {  // Suppose this source type is identified by package path and name regexp
  secret string
}

type Foo Source
```

Because `Source` is matched by regexp, and if that regexp does not also match `Foo`, then `Foo` will not be currently be considered a source.  This should be corrected by this Issue.

Implementation detail: if the underlying struct literal were marked as a source type, rather than the named type `Source`, this would propagate more easily.

mlevesquedion:
> The type `Source` will be identified as a source due to the field tag.

According to my understanding of the `fieldtags` analyzer, currently this is not the case. I now realize that this is in fact a "bug": we need the `Source` holding the tagged `Field` to be marked as a `Source`, because if the `Source` gets `Sink`ed, so will the `Field` (probably). It is my understanding that this would not be difficult to implement, now that the `fieldtags` analyzer is hooked into the `sourcetype` analyzer (as you know, the `sourcetype` analyzer is not yet hooked into the `source` analyzer, though).

PurelyApplied:
Testing just now, it looks like you're right.  I think I tested it before and it worked, but it looks like #49 updated `ast` inspection to filter the underlying type to `*ast.StructType`, where `type Foo Bar` has underlying type `*ast.Ident`.

So that is something that should also be updated by this ticket.

mlevesquedion:
I think we may want to not consider `Foo` an alias in `type Foo Source`. We certainly do want to consider `type Foo = Source` an alias though, and in fact we don't really have a choice because in the SSA form the type of a `Foo` instance will be `Source`. So we are already handling those cases. I have added tests documenting this: #104.

According to the [Go specification](https://golang.org/ref/spec#Type_declarations):
* `type Foo = Source` is a type alias, it binds an identifier to a type. Within the scope of the identifier, it is an alias for the underlying type.
[...]

PR #104, opened 2020-09-14
Add tests for type aliases

mlevesquedion:
This PR adds tests showing how type aliases and type definitions are handled.

We already handle type aliases (`type Foo = Source`), and these new tests document that behavior. The tests also show that we do not (currently) handle type definitions where the underlying type is a `Source`, such as `type Foo Source`.

This PR is related to #96.

- [x] Tests pass


PurelyApplied (approved):
What do you think of keeping `Milestone 1` as a feature-add list?  I worry about cluttering it up with every small bug fix and test-add.

mlevesquedion:
SG! I was using the Milestone to indicate what this work was related to, but I agree it'll be easier to see our progress if we just track issues that are open vs closed.

I've added a link to the relevant issue in the description instead.

PR #107, opened 2020-09-14
Report a position for source extracts

mlevesquedion:
## Context
A recent PR #81 introduced detection of `*Source` values returned by functions that have multiple return values. Such values appear only as `Extract`s in the SSA.

## This PR
The current PR addresses the reporting of such `Source`s. The key point is that `Extract`s do not have a valid `token.Pos`. Their `Pos()` returns `NoPos`, which in reports shows up as: `a source has reached a sink, source: -`. The solution proposed in this PR is to use the position of the `Extract`'s related `Tuple`: `(ssa.Extract).Tuple.Pos()`.

Since it is clearly the `Source`'s responsibility to report its position, I have moved this logic to a method on `Source`.

## Caveats
The main limitation of the proposed solution is that the position that appears in the report is not the position of the variable being declared, which may confuse users a little. That being said, reporting the position of the `Call` (e.g.) that produces the `Source` is certainly better than reporting `-`.

This limitation would become more important if a `Call` were to return more than one `Source`. In such a case it may be difficult for a user to determine precisely which `Source` is involved in the report. However, currently I don't see a way to do it that wouldn't be extremely convoluted. (There may be a way to do it using `Debug` information in the `ssa`, but the `buildssa.Analyzer` does not allow us to do that. The extremely convoluted way to do it would be to identify source-producing calls from the `ast` in advance, then when reporting a `Source` we would find the corresponding call in the `ast` and use the additional information in there, along with knowledge of the function's signature, to determine the position of the `Source` variable being created). 

- [x] Tests pass

### DIFF
--- a/internal/pkg/source/testdata/src/analyzertest/sourcetest/identification.go
+++ b/internal/pkg/source/testdata/src/analyzertest/sourcetest/identification.go
@@ -14,12 +14,6 @@
 
 package sourcetest
 
-// source container
-type Source struct {
-	Data string // source field
-	ID   int    // public
-}
-
 // This function allows us to consume multiple arguments in a single line so this file can compile
 func noop(args ...interface{}) {}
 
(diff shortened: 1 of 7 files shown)


internal/pkg/source/testdata/src/analyzertest/sourcetest/identification.go
```diff
@@ -14,12 +14,6 @@
 
 package sourcetest
 
-// source container
-type Source struct {
-	Data string // source field
-	ID   int    // public
-}
-
```
PurelyApplied, line 22: This file is not particularly large.  Is there a testing reason why you've moved this to a separate test file?

mlevesquedion: I needed the `NewSource` function in two different test files, so I figured it would make more sense to have the `Source` + associated functions in its own file. WDYT?

PurelyApplied: Oh duh.  Works for me.

PurelyApplied (approved):

Issue #93, opened 2020-09-10
Update documentation - promises and goals


PurelyApplied:
The following need to be included or updated in our documentation.

- Explicitly document what taint propagation we do and do not attempt to track.
- Document existing inference, i.e., given a source type `Foo`, what other types do we implicitly consider sources?
- State that we do not handle reflection explicitly, though some cases may be handled by general taint propagation.

mlevesquedion:
From [a comment](https://github.com/google/go-flow-levee/pull/292#pullrequestreview-627255820) by @PurelyApplied on #292:

> We should consider explicitly drawing attention to "you should make sure you trust your errors" when we update documentation surrounding the scope of taint propagation.

PR #108, opened 2020-09-14
Mark type as source if it holds a tagged field

mlevesquedion:
## This PR

The current PR is in relation to issue #96.

My original intent was to fix the inconsistent behavior we have right now where a type that holds a tagged field is not considered to be a Source. As a side effect, the way I implemented it gets us some of the way towards supporting identifying `NamedType` as a source type in `type NamedType Source`.

This PR uses the data collected by the `fieldtags` analyzer in the `sourcetype` analyzer in order to mark a `type` containing a tagged field as a `Source`. To do so, the new code scans the fields of the `types.Struct` underlying the type being declared and if it finds a tagged field, declares the type to be a `Source` type.

## Side effects

A side effect of the current implementation is that in `type NamedType Source`, because `Source` has tagged fields, and `NamedType`'s underlying `types.Struct` is the same as the `types.Struct` underlying `Source`, `NamedType` is identified as a source type. `NamedType` would not be identified as a source type if `Source` did not have tagged fields. This behavior may be surprising to users, since we don't handle other cases of `type NamedType Source`.

## Alternative implementation

An alternative implementation would be to export `type`s that hold a tagged field in the `fieldtags` analyzer, and consume those types in the `sourcetype` analyzer. This would circumvent the above-mentioned side effects.

## Discussion

Given all of the above, WDYT? **Is the current approach, with the "side effects" it brings, "ok" to you or should I use a different approach?** 

- [x] Tests pass

### DIFF
--- a/internal/pkg/sourcetype/analyzer.go
+++ b/internal/pkg/sourcetype/analyzer.go
@@ -96,7 +96,8 @@ func run(pass *analysis.Pass) (interface{}, error) {
 
 	// Members contains all named entities
 	for _, mem := range ssaInput.Pkg.Members {
-		if ssaType, ok := mem.(*ssa.Type); ok && conf.IsSource(ssaType.Type()) {
+		if ssaType, ok := mem.(*ssa.Type); ok &&
+			(conf.IsSource(ssaType.Type()) || hasTaggedField(mem, taggedFields)) {
 			exportSourceFacts(pass, ssaType, conf, taggedFields)
 		}
 	}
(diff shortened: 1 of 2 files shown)


PurelyApplied (approved):
> A side effect of the current implementation is that in type NamedType Source, because Source has tagged fields, and NamedType's underlying types.Struct is the same as the types.Struct underlying Source, NamedType is identified as a source type. 

I think this is an acceptable effect, never mind "side."  While I wouldn't necessarily expect our users to be explicitly familiar with the Golang spec, it is what I would expect as a consequence of type identity.

It is perhaps something that could be made explicit in documentation.  Cross-referencing #93 so that this PR appears in that Issue to represent that.

> NamedType would not be identified as a source type if Source did not have tagged fields. This behavior may be surprising to users, since we don't handle other cases of type NamedType Source.

I consider this a separate issue and was the intent of the scope of #96 when I opened it.  This was partially because I was under the mistaken impression at the time that field tags in underlying types were already propagating to named type, and partially because as mentioned above, I think this is the desired behavior.

Issue #97, opened 2020-09-10
Infer as a source type any type that contains a source type as a field


PurelyApplied:
Consider the following:
```golang

type Source struct {
  secret string // `levee:"source"`
}

type Container struct {
  ID int
  content *Source  // This field should implicitly be marked similarly as a source
}
```

While explicit instantiation `c := Container{ID: 10, content: Source{secret: "my-token"}}` should be currently detected by taint propagation to the struct's field, this context may be lost outside the instantiation of the wrapper.

mlevesquedion:
I assume this includes embedding? E.g. if a struct embeds another struct, implicitly it has the embedded struct as a field.

PurelyApplied:
I would say yes.  The distinction in embedding is just a method-set delegation, and we're generally not concerning ourselves with the interface layer.  The embedded field is still a field and should covered by this issue as well.

PR #112, opened 2020-09-14
Remove dead todo comments in source.go

mlevesquedion:
The first TODO, detecting `Source`s passed by value, is already handled by the code in `sourcesFromBlocks`: a source that is passed by value will have an `Alloc` that points to the parameter.

The second case is unclear, and I believe as a TODO it is unhelpful. We currently have some handling of sources in collections. I believe the remainder of what the TODO could be pointing to is covered by issues #96 and #97.

- [x] Tests pass

### DIFF
--- a/internal/pkg/source/source.go
+++ b/internal/pkg/source/source.go
@@ -293,8 +293,6 @@ func sourcesFromParams(fn *ssa.Function, conf classifier) []*Source {
 			if n, ok := t.Elem().(*types.Named); ok && conf.IsSource(n) {
 				sources = append(sources, New(p, conf))
 			}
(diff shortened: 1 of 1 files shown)


PurelyApplied (approved):
You can open a single PR for multiple cleanups.

mlevesquedion:
SG. Since they have different reasons behind them I thought it would be clearer to have a PR for each.

PurelyApplied:
> SG. Since they have different reasons behind them I thought it would be clearer to have a PR for each.

Yeah, I meant the four TODO removal PRs.  The one that actually touched code, even if it was just removing a constant flag and dedenting, I appreciated as a stand-alone.

👍

PR #116, opened 2020-09-15
Do not visit instructions that are not reachable from source

mlevesquedion:
This PR implements a fix that addresses incorrect reports produced when running `levee` on `kubernetes`.

The incorrect reports were caused by traversing to instructions unreachable from the `Source` from which the traversal originated. Here is an example:
```go
func TestDoesNotReachSinkAfterSourceThroughValueCreatedBeforeSource() {
	k := newKey()

	_ = map[string]core.Source{}[k.name]

	core.Sink(k.Name())
}
```

The incorrect report in this case is caused by the following traversal flow: `map[string]core.Source{}[k.name] --> k.name --> k --> k.Name() --> core.Sink` (the `Source` value is produced by the map lookup). Clearly, the `Source` value in this example cannot taint `k`, and in general we should avoid traversing to instructions that are not reachable from the `Source`.

There are some subtleties involved in the above test case, that are required for the incorrect behavior to emerge. First, `k` has to be a pointer returned from a function. That way, it does not have an `Alloc`. If it did have an `Alloc`, then we would not traverse through it (see `source.go:245`).  Second, for similar reasons, we have to access `k`'s `name` field via `k.Name()`.

The assignment to `_` is simply to avoid compiler warnings about unused variables. Interestingly, even though the value returned from the lookup is unusable because of the blank identifier, it still shows up in the SSA.

I have included a second test case:
```go
func TestDoesNotReachSinkInIfBeforeSourceThroughValueCreatedBeforeSource() {
	k := newKey()

	if true {
		core.Sink(k.Name())
	}

	_ = map[string]core.Source{}[k.name]
}
```
This case more closely mirrors the cases encountered on `kubernetes`. The key nuance here is that the call to `core.Sink` is in a different block from the map lookup. Due to the incorrect traversal order described earlier (which is largely the same here), our usual check for calls that are not reachable from the previous instruction (see `source.go:146`) wasn't stopping the traversal.

All of these subtleties and nuances are making me realize that our traversal code is fairly convoluted. I will seek to address this in future PRs. (For context, at present it is not possible to remove any of the other checks we have in place without causing tests to fail).

- [x] Tests pass

### DIFF
--- a/internal/pkg/levee/testdata/src/example.com/tests/callorder/beforesource.go
+++ b/internal/pkg/levee/testdata/src/example.com/tests/callorder/beforesource.go
@@ -18,6 +18,7 @@ import (
 	"example.com/core"
 )
 
+// This type should *not* be identified as a Source.
 type key struct {
 	name string
 }
@@ -33,6 +34,7 @@ func newKey() *key {
 }
 
 func TestDoesNotReachSinkAfterSourceThroughValueCreatedBeforeSource() {
+	// Taint should not propagate to this value.
 	k := newKey()
 
 	_ = map[string]core.Source{}[k.name]
@@ -41,6 +43,7 @@ func TestDoesNotReachSinkAfterSourceThroughValueCreatedBeforeSource() {
 }
 
 func TestDoesNotReachSinkInIfBeforeSourceThroughValueCreatedBeforeSource() {
+	// Taint should not propagate to this value.
 	k := newKey()
 
 	if true {
--- a/internal/pkg/levee/testdata/src/example.com/tests/callorder/beforesource.go
+++ b/internal/pkg/levee/testdata/src/example.com/tests/callorder/beforesource.go
@@ -28,7 +28,7 @@ func (k *key) Name() string {
 
 func newKey() *key {
 	return &key{
-		name: "source",
+		name: "",
 	}
 }
 
--- a/internal/pkg/levee/testdata/src/example.com/tests/callorder/beforesource.go
+++ b/internal/pkg/levee/testdata/src/example.com/tests/callorder/beforesource.go
@@ -1,0 +1,51 @@
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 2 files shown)


internal/pkg/levee/testdata/src/example.com/tests/callorder/beforesource.go
```diff
@@ -0,0 +1,51 @@
...
+package callorder
+
+import (
+	"example.com/core"
+)
+
+type key struct {
+	name string
+}
+
```
PurelyApplied, line 24: Even with the context of this PR's `Conversation`, I was caught off guard here.  I seemed like the tests were expecting this to be identified as a source.

Consider adding a comment to the `type key` that we do *not* expect this to be a source, nor do we expect taint to through its use below.

internal/pkg/source/source.go
```diff
@@ -75,12 +75,16 @@ func New(in ssa.Node, config classifier) *Source {
...
 func (s *Source) dfs(n ssa.Node) {
-	s.preOrder = append(s.preOrder, n)
-	s.marked[n.(ssa.Node)] = true
+	instr, ok := n.(ssa.Instruction)
+	if ok && !s.reachableFromSource(instr) {
+		return
+	}
 
-	if instr, ok := n.(ssa.Instruction); ok {
+	if ok {
```
PurelyApplied, line 83: Nit: Does this read better?

```golang
		if !s.reachableFromSource(instr) {
			return
		}
		s.record(instr)
	}
```

Maybe, maybe not.

PurelyApplied (approved):

internal/pkg/source/source.go
mlevesquedion: I think it does, thanks! :)

Zenix27 (approved):

internal/pkg/levee/testdata/src/example.com/tests/callorder/beforesource.go
mlevesquedion: SG. I've added a comment to the `type` declaration that it shouldn't be identified as a source.

For the `do not expect taint` part I added comments in each test, in case we add other cases later were we *do* expect the `key` to be tainted. (Also I think having the comments in the cases makes it easier to understand the intent when reading just the test cases).

PR #117, opened 2020-09-15
Filter FieldAddr in referrersToVisit

mlevesquedion:
This should have been done in a previous refactoring: #82, the intent of which was to separate the code related to visiting a value's referrers in two parts: 1. select the instructions that should be visited; 2. visit them.

- [x] Tests pass


PurelyApplied (approved):
This may warrant an update to the filter's docstring.  Consider something more abstract so that it doesn't require update on every change.

mlevesquedion:
Agreed! I've made the docstring more focused on the overall behavior.

PR #118, opened 2020-09-17
Fix traversal bugs

mlevesquedion:
## Context

This PR implements fixes for bugs that I have found when running the tool on `kubernetes`. I am done investigating the reports so there will be no more bug fixes originating from this investigation.

## The bugs

This PR fixes 2 distinct bugs related to traversal:
1. Visiting unreachable `Operands`.
2. Using global state to track `maxInstructionReached` and `lastVisitedBlock`.

## 1. Visiting unreachable `Operands`

When considering which referrers to visit, we avoid referrers that are not reachable from the current instruction. We do not do this for operands, however, which can lead to incorrect reports on code such as this:
```go
func TestTaintedColocatedArgumentDoesNotReachSinkThatPrecedesColocation() {
	s := core.Source{}
	i := newInnocuous()
	if err := fail(i); err != nil {
		core.Sink(err) // A REPORT IS INCORRECTLY PRODUCED HERE
	}
	taintColocated(s, i)
}
```
The report in the above case is caused by the following taint flow: `s → taintColocated(...) → i → fail(...) → err → core.Sink(...)`. Taint propagates to `i` because it is an `Operand` of the call to `taintColocated`.

 To fix this bug, I have moved the reachability check into `source.go:dfs` so that it will be done for every `ssa.Node`, which includes `Operands`, not just `Referrers`.

## 2. Using global state to track `maxInstructionReached` and `lastVisitedBlock`

These values need to evolve independently for each traversal path in the SSA graph. Using global values means that what happens in one traversal path can influence what happens in a different traversal path, which is clearly wrong. I have not been able to write a test case demonstrating the incorrect behavior. I believe that writing such a test is not a simple task, and the resulting test is likely to be difficult to understand. Here is a (slightly) simplified example of a real error case from `kubernetes`, in `cmd/kubermark/hollow_node.go`:

```go
func run(config *hollowNodeConfig) {
	clientConfig, err := config.createClientConfigFromFile()
	if err != nil {
		klog.Fatalf("Failed to create a ClientConfig: %v. Exiting.", err)
	}

	client, err := clientset.NewForConfig(clientConfig)
	if err != nil {
		klog.Fatalf("Failed to create a ClientSet: %v. Exiting.", err) // even though this error does not contain sensitive information, we want a report here because currently levee has no way to determine that
	}

	if config.Morph == "kubelet" {
		f, c := kubemark.GetHollowKubeletConfig(config.createHollowKubeletOptions())

		heartbeatClientConfig := *clientConfig
		heartbeatClient, err := clientset.NewForConfig(&heartbeatClientConfig)
		if err != nil {
			klog.Fatalf("Failed to create a ClientSet: %v. Exiting.", err) // even though this error does not contain sensitive information, we want a report here because currently levee has no way to determine that
		}

		endpoint, err := fakeremote.GenerateEndpoint()
		if err != nil {
			klog.Fatalf("Failed to generate fake endpoint %v.", err) // WE DO NOT WANT A REPORT HERE, ERR SHOULD NOT BE TAINTED
		}
		fakeRemoteRuntime := fakeremote.NewFakeRemoteRuntime()
		if err = fakeRemoteRuntime.Start(endpoint); err != nil {
			klog.Fatalf("Failed to start fake runtime %v.", err)
		}
		defer fakeRemoteRuntime.Stop()
		runtimeService, err := remote.NewRemoteRuntimeService(endpoint, 15*time.Second)
		if err != nil {
			klog.Fatalf("Failed to init runtime service %v.", err)  // WE DO NOT WANT A REPORT HERE, ERR SHOULD NOT BE TAINTED
		}

		hollowKubelet := kubemark.NewHollowKubelet(
			f, c,
			client,
			heartbeatClient,
			nil,
			fakeRemoteRuntime.ImageService,
			runtimeService,
			nil,
		)
		hollowKubelet.Run()
	}
}
```

The two lines producing incorrect reports are indicated by comments. The incorrect reports are due to taint incorrectly propagating from colocation in the call to `kubemark.NewHollowKubelet`.

Fixing that bug (propagating to `Operands` we shouldn't propagate to) causes the *correct* reports to disappear. The precise sequence of steps that causes this is not clear, but tracking state per-traversal-path rather than globally (per-source)  fixes it. In fact, just tracking the `lastVisitedBlock` properly fixes the bug, but in the interest of preventing future issues I have also fixed the code involving `maxInstructionReached`.

## The commits

Since this PR fixes 2 bugs, it should perhaps be 2 PR's. However, I believe that the 2 bugs are closely related, and both fixes are required to exhibit correct behavior on the above-mentioned case that prompted this PR, so having just 1 PR makes sense. (If you disagree, please let me know and I will split the PR).

In the interest of making the PR easier to review, I have split the changes into 3 commits:
1. add tests
2. fix operands bug
3. fix global state bug

- [x] Tests pass

### DIFF
--- a/internal/pkg/source/source.go
+++ b/internal/pkg/source/source.go
@@ -79,6 +79,13 @@ func (s *Source) dfs(n ssa.Node) {
 		if !s.reachableFromSource(instr) {
 			return
 		}
+		// If the referrer is in a different block from the one we last visited,
+		// and it can't be reached from the block we are visiting, then stop visiting.
+		if ib := instr.Block(); s.lastBlockVisited != nil &&
+			ib != s.lastBlockVisited &&
+			!s.canReach(s.lastBlockVisited, ib) {
+			return
+		}
 		s.record(instr)
 	}
 	s.preOrder = append(s.preOrder, n)
@@ -131,14 +138,6 @@ func (s *Source) referrersToVisit(n ssa.Node) (referrers []ssa.Instruction) {
 		return
 	}
 	for _, r := range *n.Referrers() {
-		// If the referrer is in a different block from the one we last visited,
-		// and it can't be reached from the block we are visiting, then stop visiting.
-		if rb := r.Block(); s.lastBlockVisited != nil &&
-			rb != s.lastBlockVisited &&
-			!s.canReach(s.lastBlockVisited, rb) {
-			continue
-		}
-
 		if c, ok := r.(*ssa.Call); ok {
 			// This is to avoid attaching calls where the source is the receiver, ex:
 			// core.Sinkf("Source id: %v", wrapper.Source.GetID())
--- a/internal/pkg/source/source.go
+++ b/internal/pkg/source/source.go
@@ -91,7 +91,12 @@ func (s *Source) dfs(n ssa.Node, maxInstrReached map[*ssa.BasicBlock]int, lastBl
 			!s.canReach(lastBlockVisited, ib) {
 			return
 		}
-		s.record(instr, mirCopy, &lastBlockVisited)
+
+		b := instr.Block()
+		if i, ok := indexInBlock(instr); ok && mirCopy[b] < i {
+			mirCopy[b] = i
+		}
+		lastBlockVisited = b
 	}
 	s.preOrder = append(s.preOrder, n)
 	s.marked[n] = true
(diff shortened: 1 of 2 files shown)


internal/pkg/source/source.go
```diff
@@ -62,54 +60,64 @@ func (s *Source) Pos() token.Pos {
...
-		s.visitOperands(n, operands)
+		s.visitOperands(n, operands, mirCopy, lastBlockVisited)
 	}
 }
 
-func (s *Source) record(target ssa.Instruction) {
+func (s *Source) record(target ssa.Instruction, maxInstrReached map[*ssa.BasicBlock]int, lastBlockVisited **ssa.BasicBlock) {
 	b := target.Block()
-	s.lastBlockVisited = b
+	*lastBlockVisited = b
```
PurelyApplied, line 109: Updating via a mutable `**ssa.BasicBlock` is a severe code smell to me.  Consider returning `(instrBlock *ssa.BasicBlock)` instead, or extracting `thisBlock = instr.Block()` to the calling function.

PurelyApplied (approved):
I'm glad these have become parameters.

Please consider the below before merging.

internal/pkg/source/source.go
mlevesquedion: Hmm. I agree that this smells pretty bad.

In fact, re-reading the code now, it seems to me that this does not need to be a method, and the code can be simplified enough to be inlined:
```go
b := instr.Block()
if i, ok := indexInBlock(instr); ok && mirCopy[b] < i {
	mirCopy[b] = i
}
lastBlockVisited = b
```

PR #119, opened 2020-09-17
Refactor: Check for marked nodes in dfs

mlevesquedion:
Since this check needs to be done for every `Node`, and it is currently being done separately for `Referrers` and `Operands`, this refactoring simplifies the code a little by only doing the check in one place.

- [x] Tests pass


internal/pkg/source/source.go
```diff
@@ -221,7 +221,7 @@ func (s *Source) visitOperands(n ssa.Node, operands []*ssa.Value) {
 
 	for _, o := range operands {
 		n, ok := (*o).(ssa.Node)
```
mlevesquedion, line 223: (We need `n` to be out of the if for a call to `dfs` later in the function.)

PurelyApplied (approved):

PR #121, opened 2020-09-17
Fix: skip to next Operand instead of skipping all remaining Operands

mlevesquedion:
This PR fixes a bug in the way visiting a `Call`'s `Operands` works. Currently we `return` when we meet an `Operand` that is an `Alloc` that we don't want to visit. What we actually want to do is `continue` to the next `Operand`.

- [x] Tests pass


mlevesquedion:
(The following is unrelated to this bug fix.)

In the test case introduced by this PR, if we were to add a call to `core.Sink(i)`, it would not get a report. This is because we do not traverse to `Alloc` operands. This is related to the failing test in this WIP PR: #60.

PurelyApplied (approved):

PR #123, opened 2020-09-22
Debug: Render closures

mlevesquedion:
The current (debugging) code for rendering the `ssa` of a function does not render closures. Rendering closures is important because they can contain sources reaching sinks.

I'm removing the `_` (blank identifiers) from `b.Write(Byte|String)` calls because I find they make the code less readable.

- [x] Tests pass


PurelyApplied (approved):
The debug packages remain entirely untested.  If you plan to continue iterating in it, please add testing soon.

mlevesquedion:
SG. I don't have other iterations in mind right now, but I will provide tests either soon, or before I make new changes.

Issue #90, opened 2020-09-10
Add benchmarks


PurelyApplied:
Particularly as we examine cross-function and whole-program analysis, it will be useful to be able to benchmark our overall runtime.  In particular, anecdotal evidence suggests that the construction of the SSA graph by `buildssa.Analyzer` constitutes the majority of our runtime.  It would be good to be able to put actual data to such claims, identify any significant performance regressions, and evaluate the runtime costs of future work.

mlevesquedion:
I think we might want to start with a very high level benchmark, e.g. how long does it take to analyze a large codebase such as `kubernetes`. If this runtime stays stable over time, then we will likely not need any further benchmarks. Currently the runtime is not significantly different from running a standard `go vet`. Such a benchmark could live outside the codebase in some kind of recurring job.

PR #124, opened 2020-09-22
Replace traversal with interpretation

mlevesquedion:
## Why?

The current approach to taint propagation is based on traversing the implicit graph of Referrer/Operand relationships. Because it is not always desirable to taint all of the Referrers/Operands of a given Source, while traversing it is necessary to perform additional checks to avoid propagating to things we shouldn't, e.g. propagating from a tainting instruction to a sink call that occurred before it. Most of these checks have been introduced to fix bugs discovered by running the analyzer on kubernetes: #118, #116, #80, #79. Debugging these cases was fairly complicated, even using the debugging aids (visualizations of the ssa graph and dumps of the ssa code). I believe debugging was difficult because the code itself is complicated, and it has gotten more complicated as a result of fixing the bugs.

## What

This PR replaces the current approach to taint propagation with a new approach that is based on "interpreting" the SSA code. Information about ssa `Values` is maintained in a map that contains the state of each `Value`. Propagation is performed by stepping through the instructions, introducing taint when a `Source` value is encountered and propagating it when reading from a tainted `Value`. This approach intrinsically takes instruction order into account, is more explicit, and is (I believe) easier to understand.

I also believe that this approach will be easier to extend, and it handles some test cases that the current approach does not. In all of the following cases, the current approach fails to produce a report:

```go
func TestIncorrectSanitizationByValue(s core.Source) {
	core.Sanitize(s)
	core.Sink(s) // TODO want "a source has reached a sink"
}
```

```go
func TestDoublePointer(s **core.Source) {
	core.Sink(s)   // TODO want "a source has reached a sink"
	core.Sink(*s)  // TODO want "a source has reached a sink"
	core.Sink(**s) // TODO want "a source has reached a sink"
}
```

```go
func TestTaintIsPropagatedToColocatedPointerArguments(s core.Source, ip *core.Innocuous) {
	i := core.Innocuous{}
	taintColocated(s, &i, ip)
	core.Sink(s)  // want "a source has reached a sink"
	core.Sink(i)  // TODO want "a source has reached a sink"
	core.Sink(ip) // want "a source has reached a sink"
}
```

```go
func TestStructHoldingSourceAndInnocIsTainted(s core.Source, i core.Innocuous) {
	h := Holder{
		s,
		i,
	}
	core.Sink(h) // TODO want "a source has reached a sink"
}
```

## Related PRs

This PR builds on #125, which introduces the above-mentioned tests.

## Additional information

I have run both "versions" of the analyzer on `kubernetes` and they produce the same reports. This provides some evidence that we would not be missing cases by moving to the new approach. The run time is not significantly different.

## Future changes

* Introduce further debugging aids. I have not included it in this PR, but I have written a debugger that allows a developer to step through each instruction in a function to see how the state evolves during interpretation. I think it would also be useful to render the graph of blocks (the control flow graph).
* Cover ssa `Instructions` that are currently not covered, and add tests for them.

- [x] Tests pass


vinayakankugoyal:
One way to make change could be to add the new package with its test, get that reviewed and checked in. Then in subsequent PRs update the code to use the new package.

internal/pkg/interp/interpreter.go
```diff
@@ -0,0 +1,425 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package interp defines an analyzer that performs taint analysis on
+// functions by interpreting their ssa code.
+package interp
```
vinayakankugoyal, line 17: Are there unit tests specifically testing this?

mlevesquedion: This is a good question. The answer is no.

The code in `internal/pkg/interp` is tested via the "end-to-end" tests under `internal/pkg/levee`.

The main reason I didn't write unit tests is that for the purposes of "trying out" this new approach I felt like the end-to-end tests were sufficient. I do think this new approach is preferable to the current approach, and it should probably have unit tests before (and if) it is merged.

The other reason why there are currently no unit tests is that I am not sure what those tests should look like. Ideally, I would like to test that processing each type of instruction has the intended effect, e.g. updating the `state`, appending a `result`, etc. I think this should be doable using `analysistest` and `want` comments, at least to some extent, but I am not sure what the cost/benefit would be, and how strongly the tests would have to be coupled to the implementation.

mlevesquedion:
> One way to make change could be to add the new package with its test, get that reviewed and checked in. Then in subsequent PRs update the code to use the new package.

I agree that that would be a cleaner way to do it.

My intent in writing this PR was to determine whether this new approach I had in mind would make sense and to quickly get an opinion on it. I feel that a big part of that is showing that it can be swapped in without failing any tests (and even making some additional tests pass). I wanted to avoid spending too much time on this code in the event that we decided not to integrate it.

internal/pkg/interp/interpreter.go
```diff
@@ -0,0 +1,425 @@
...
+		results = append(results, analyzeFunction(conf, fieldPropagators, f)...)
+	}
+
+	return results, nil
+}
+
+// analyzeFunction performs taint analysis on the given function and
+// returns a slice of Results.
+func analyzeFunction(conf *config.Config, fieldpropagators fieldpropagator.ResultType, f *ssa.Function) []Result {
+	// nothing to analyze
```
vinayakankugoyal, line 103: https://github.com/golang/go/wiki/CodeReviewComments#comment-sentences

here and other places.

vinayakankugoyal:
As I understand this change does not improve the performance but improves our ability to debug issues in the analyzer, correct? Could we compare the 2 approaches in terms of performance? Are we gaining or loosing anything significant by switching over to this approach?

internal/pkg/interp/interpreter.go
mlevesquedion: That link seems to say that comments for *declarations* should be full sentences, which I believe they are (e.g. in this snippet I think the doc comment for `analyzeFunction` is in good form). I don't have the link, but I remember reading elsewhere that if a comment is just a couple words and not a complete sentence, then it is ok to omit capitalizing the first word + ending with a period.

mlevesquedion:
That is correct. I did not emphasize this in the PR description, but I believe that this change will also make it easier for us to extend the analyzer, in large part because 1. the code is more straightforward, 2. we are explicitly tracking how state changes as we step through the program.

Concerning performance: When I ran this "version" of the analyzer on kubernetes, I got a runtime of just under 3 minutes. That is in the same ballpark as the "current" version, which in my experience typically takes 3-4 minutes to run.

> Are we gaining or loosing anything significant by switching over to this approach?

I think the most important thing that we are gaining is an analyzer whose core logic is easier to understand and to modify. The recent changes that I've had to make where rather painful, and overall it feels like the current approach is backwards. I think stepping through the instructions is simpler and can give more precise results.

internal/pkg/interp/interpreter.go
vinayakankugoyal: I re-read internal google style guide on this. You are correct that this does not need to be capitalized or punctuated. Thanks!

PurelyApplied (commented):
Some high-level thoughts:

In this project's pre-OSS infancy, we were more concerned with performance.  I believe that was part of the reason we started with source identification and referrers traversal -- to "cut out the middle man" as it were and only look at those Values and Instructions that were relevant.  While this project is now more concerned with the results produced than a necessarily quick runtime, I am concerned about how these different approaches would scale.  In general, I would expect the places where a Source could reach to be a very small portion of a user's overall codebase.  Do you have an anecdotal runtime comparison on how this approach vs the existing approach run against our favorite for-instance codebase Kubernbetes?  Quasi-related, we may wish to push on #90 to quantify some of these differences.

I'm not sure I would want to delete the `source` analyzer.  It seems to me that it would be a natural input to this analyzer, rather than constant delegation to `config`.  I stand by the more modular design of having separate identification passes for "root" sources, sanitizer, and sink calls, and to have `levee` (or this in its place) act as a culminating analysis with those identifications provided as input.  This feeds into vinayakankugoyal's thread / our recurring conversations re: testing.  Also re: paragraph one, it could be used as an initial guide to avoid analyzing in depth functions which never touch a source.  This may require some improvement in the returned result from the `source` analyzer, but I think that would be worthwhile in any case.

Do you want implementation-detail comments to this PR?  I'm trying to keep it high-level at the moment to talk about approach and to avoid splitting out too many threads, but can provide lower-level comments if you're interested.

I haven't looked at any of the "moved" tests, becuase `git` seems to have gotten very confused by the diffs set.  I'll wait to review any of these until after the pending test PRs have merged and those changes have folded back into this branch.

mlevesquedion:
I think it's a good idea to stay high level for now! I've converted the PR to Draft mode. I think this is clearer than putting "WIP" in the title, so in future I will use Drafts instead.

I do have an anecdotal runtime comparison (from a previous comment to Vinayak):
> Concerning performance: When I ran this "version" of the analyzer on kubernetes, I got a runtime of just under 3 minutes. That is in the same ballpark as the "current" version, which in my experience typically takes 3-4 minutes to run.

There may be weird cases out there were performance would break down, but in general I wouldn't expect the performance to be very different from what we have currently. In my understanding, the interpretation approach would be slower on large multi-block functions where there are multiple paths to try, with lots of instructions on each path. In fact if you were to create a "tree" of if/else statements the runtime would be exponential in the number of "levels" in the tree. That is unlikely to occur in practice, I would hope that would be caught in code review!

Concerning the `source` analyzer, I was thinking of consuming the `sourcetype` analyzer in the near future :thinking:. Given the above anecdotal performance comparison (which you didn't have at the time of writing your comment), I think there would be little performance benefit to having a separate analyzer prune the list of functions we want to analyze.

> I haven't looked at any of the "moved" tests, becuase git seems to have gotten very confused by the diffs set. I'll wait to review any of these until after the pending test PRs have merged and those changes have folded back into this branch.

Ack.

mlevesquedion:
After rebasing #125, I had to make a small change to track sanitization status explicitly. This means that we need to maintain the following invariant: `val.source == nil` iff `val.sanitized`.

Why do we need to track sanitization status explicitly?

Well, on the one hand, we want to create a Source when we have this situation:
```go
func Foo(thisCouldBeAnything interface{}) {
    s := thisCouldBeAnything.(Source)
}
```

On the other hand, we don't want to create a Source when we have this situation:
```go
func Foo(s core.Source) {
    safe := sanitize(s).(Source) // "sanitize" returns interface{}
}
```

Besides tracking sanitization explicitly, I don't see a way to handle both cases.

mlevesquedion:
Closing because of staleness. This will likely be revisited in the future, but right now isn't a priority.

PR #125, opened 2020-09-22
Add test targets

mlevesquedion:
This PR reworks existing tests and adds new tests. The new tests include cases that the current approach (graph traversal) does not handle, but that the approach proposed in #124 does handle.

- [x] Tests pass

### DIFF
--- a/internal/pkg/levee/testdata/src/example.com/tests/colocation/tests.go
+++ b/internal/pkg/levee/testdata/src/example.com/tests/colocation/tests.go
@@ -18,16 +18,14 @@ import (
 	"example.com/core"
 )
 
-func TestTaintIsPropagatedToColocatedPointerArguments(s core.Source, ip *core.Innocuous) {
-	i := core.Innocuous{}
+func TestTaintIsPropagatedToColocatedPointerArguments(s core.Source, i core.Innocuous, ip *core.Innocuous) {
 	taintColocated(s, &i, ip)
 	core.Sink(s)  // want "a source has reached a sink"
 	core.Sink(i)  // TODO want "a source has reached a sink"
 	core.Sink(ip) // want "a source has reached a sink"
 }
 
-func TestTaintIsPropagatedToColocatedPointerArgumentsThroughEface(s core.Source, ip *core.Innocuous) {
-	i := core.Innocuous{}
+func TestTaintIsPropagatedToColocatedPointerArgumentsThroughEface(s core.Source, i core.Innocuous, ip *core.Innocuous) {
 	taintColocatedEface(s, &i, ip)
 	core.Sink(s)  // want "a source has reached a sink"
 	core.Sink(i)  // TODO want "a source has reached a sink"
--- a/internal/pkg/levee/testdata/src/example.com/tests/colocation/tests.go
+++ b/internal/pkg/levee/testdata/src/example.com/tests/colocation/tests.go
@@ -18,12 +18,24 @@ import (
 	"example.com/core"
 )
 
-func TestTaintIsPropagatedToOperandAfterNonArrayAlloc(s core.Source, ip *core.Innocuous) {
+func TestTaintIsPropagatedToColocatedPointerArguments(s core.Source, ip *core.Innocuous) {
 	i := core.Innocuous{}
 	taintColocated(s, &i, ip)
+	core.Sink(s)  // want "a source has reached a sink"
+	core.Sink(i)  // want "a source has reached a sink"
 	core.Sink(ip) // want "a source has reached a sink"
 }
 
-func taintColocated(a interface{}, i *core.Innocuous, c interface{}) interface{} {
-	return nil
+func TestTaintIsPropagatedToColocatedPointerArgumentsThroughEface(s core.Source, ip *core.Innocuous) {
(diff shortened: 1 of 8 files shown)


internal/pkg/levee/testdata/src/example.com/tests/structlit/tests.go
```diff
@@ -0,0 +1,62 @@
...
+}
+
+func TestStructHoldingSourceAndInnocIsTainted(s core.Source, i core.Innocuous) {
+	h := Holder{
+		s,
+		i,
+	}
+	core.Sink(h) // TODO want "a source has reached a sink"
+}
+
```
PurelyApplied, line 39: Consider adding the tests for an unnamed struct, e.g.
```golang
	h := struct{
		s *core.Source
		i *core.Innocuous
	}{
		s,
		i,
	}
	core.Sink(h) // TODO want "a source has reached a sink"
```

internal/pkg/levee/testdata/src/example.com/tests/typeassert/tests.go
```diff
@@ -0,0 +1,29 @@
...
+
+func TestSourceAssertedFromTaintedEface(s core.Source) {
+	var e interface{} = s
+	core.Sink(e.(core.Source)) // want "a source has reached a sink"
+}
+
+func TestSourcePointerAssertedFromTaintedEface(s *core.Source) {
+	var e interface{} = s
+	core.Sink(e.(*core.Source)) // want "a source has reached a sink"
+}
```
PurelyApplied, line 29: To distinguish between source detection and propagation detection here, consider adding

```golang
func Test(iface interface{}) {
  s := iface.(*core.Source)
  core.Sink(s) // want ...
}
```

PurelyApplied (changes requested):
🎉 Love to see wider testing.  I've suggested a few more.

internal/pkg/levee/testdata/src/example.com/tests/structlit/tests.go
PurelyApplied: I didn't notice until I saw the other PR, but this is an example of a `struct literal`.  I'm not sure what this file was meant to be testing given the name, since it does not test struct literals (beyond the underlying literals to the named type declarations).

internal/pkg/levee/testdata/src/example.com/tests/pointers/tests.go
```diff
@@ -0,0 +1,30 @@
...
+func TestNew() {
+	s := new(core.Source)
+	core.Sink(s) // want "a source has reached a sink"
+}
+
+func TestDoublePointer(s **core.Source) {
+	core.Sink(s)   // TODO want "a source has reached a sink"
+	core.Sink(*s)  // TODO want "a source has reached a sink"
+	core.Sink(**s) // TODO want "a source has reached a sink"
+}
```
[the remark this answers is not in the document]
mlevesquedion: Good point! To the best of my knowledge those cases don't exist. Added.

internal/pkg/levee/testdata/src/example.com/tests/structlit/tests.go
mlevesquedion: I think that is a very good test case to add! Done.

To my knowledge, all of them are struct literals: https://tour.golang.org/moretypes/5, so I think the name `structlit` is ok.

internal/pkg/levee/testdata/src/example.com/tests/typeassert/tests.go
mlevesquedion: Very good idea.

This test will actually fail in the `interp` PR, because right it doesn't create new `Source`s from `TypeAssert`s, it only propagates taint through `typeassert`s. I'll go and fix that.

I've added an extra test that uses the very common `, ok` form.

(Interestingly, the `, ok` test passes, but the one you suggested doesn't. This is because in the `, ok` case we create a Source from the `Extract`, which we handle, while in the case you suggested we can only detect the source from the `TypeAssert` instruction, which we don't handle.

internal/pkg/levee/testdata/src/example.com/tests/colocation/tests.go
```diff
@@ -18,12 +18,24 @@ import (
 	"example.com/core"
 )
 
-func TestTaintIsPropagatedToOperandAfterNonArrayAlloc(s core.Source, ip *core.Innocuous) {
+func TestTaintIsPropagatedToColocatedPointerArguments(s core.Source, ip *core.Innocuous) {
```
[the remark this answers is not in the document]
mlevesquedion: I've moved them into the parameters. The test is now more readable because more consistent. (The behavior is unchanged.)

internal/pkg/levee/testdata/src/example.com/tests/structlit/tests.go
PurelyApplied: I've been one step meta too long and was conflating the decl with the literal.

An int literal: `5`

A struct literal: `Foo{1, 2, 3}`.

`self.duh()`.

PurelyApplied (approved):

Issue #89, opened 2020-09-10
Refine identification of sources, sinks, and sanitizers


PurelyApplied:
A user may wish to configure identification of sources, sinks, and sanitizers based on many criteria beyond package, type, and method names.

See [previous discussion](https://github.com/google/go-flow-levee/pull/62/files?short_path=efa6200) for context.

> ### Identification of sources, sinks, and sanitizers
> 
> Much of our work is directly with `ssa.Value`s and `ssa.Instruction`s.
> Whether those values or instructions represent one element or another,
>  the configuration to identify these members should be as uniform as possible.
> In that light, specification should follow:
> 
> #### ValueSpecifier
> A ValueSpecifier marks an `ssa.Value`, e.g., for identification as a source or as the safe > result of a sanitizer.
> Values should be specifiable via any or all of:
> 
> * Specifier name: For identification during reporting
> * Type / Field: Package path and type name, optional field name
> * Field tags
> * Scope: local variable, free variable, or global
> * Context: within a specific package or function call.  See CallSpecifier below.
> * Is a reference
> * Is "reference-like," e.g., slices, maps, chans.
> * Const value (e.g., `"PASSWORD"`)
> * Argument position (for use below)
> 
> #### CallSpecifier
> A CallSpecifier marks an `*ssa.Call`, e.g., as a sink or sanitizer.
> Calls should be specifiable via any or all of:
> 
> * Specifier name: For identification during reporting
> * Symbol: package path, function name, optional receiver name.
> * Context: A CallSpecifier indicating scope, such as package or specific functions.
> * Based on argument value and position (ValueSpecifier).
> * Based on return values (ValueSpecifier)

PR #126, opened 2020-09-22
Make field tags configurable

### LINKED ISSUE #115: Make identifying field tags configurable
Currently, we only examine for the tag `levee:"source"`.  The source tag should allow this to be user-configurable.

mlevesquedion:
Closes #115.

- [x] Tests pass

### DIFF
--- a/configuration/README.md
+++ b/configuration/README.md
@@ -35,13 +35,13 @@ type Example struct {
 }
 ```
 
-The tag `levee:"source"` is built-in. Additional tags may be identified via regexp:
+The tag `levee:"source"` is built-in. Additional tags may be identified via explicit string literals (not regexps). The following example shows how the `levee:"source"` tag could be defined if it weren't built-in:
 ```json
 {
 	"FieldTags": [
 		{
-			"Key": "<key regexp>",
-			"Val": "<value regexp>"
+			"Key": "levee",
+			"Val": "source"
 		}
 	]
 }
--- a/configuration/README.md
+++ b/configuration/README.md
@@ -28,6 +28,25 @@ Sources are identified via regexp according to package, type, and field names.
 }
 ```
 
+Sources may also be identified via field tags:
+```go
+type Example struct {
+	fieldName fieldType `levee:"source"` // this field will be considered a Source
+}
+```
+
+The tag `levee:"source"` is built-in. Additional tags may be identified via regexp:
+```json
+{
+	"FieldTags": [
+		{
+			"Key": "<key regexp>",
+			"Val": "<value regexp>"
+		}
+	]
+}
+```
+
 Sinks and sanitizers are identified via regexp according to package, method, and (optional) receiver name.
 
 ```json
--- a/internal/pkg/config/config.go
+++ b/internal/pkg/config/config.go
@@ -45,8 +45,8 @@ type Config struct {
 }
 
 type fieldTagMatcher struct {
-	KeyRE regexp.Regexp
-	ValRE regexp.Regexp
+	Key string
+	Val string
 }
 
 // IsSourceFieldTag determines whether a field tag made up of a key and value
@@ -66,7 +66,7 @@ func (c Config) IsSourceFieldTag(key, val string) bool {
 }
 
 func (ftm fieldTagMatcher) matches(key, val string) bool {
-	return ftm.KeyRE.MatchString(key) && ftm.ValRE.MatchString(val)
+	return ftm.Key == key && ftm.Val == val
 }
 
 func (c Config) IsSinkCall(call *ssa.Call) bool {
--- a/internal/pkg/config/config.go
+++ b/internal/pkg/config/config.go
@@ -45,8 +45,8 @@ type Config struct {
 }
 
 type fieldTagMatcher struct {
-	Key regexp.Regexp
-	Val regexp.Regexp
+	KeyRE regexp.Regexp
+	ValRE regexp.Regexp
 }
 
 // IsSourceFieldTag determines whether a field tag made up of a key and value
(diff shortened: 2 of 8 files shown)


internal/pkg/fieldtags/analyzer.go
```diff
@@ -71,73 +71,81 @@ func run(pass *analysis.Pass) (interface{}, error) {
...
-			if tag[j] == '\\' {
+			if wholeTag[j] == '\\' {
 				j++
 			}
 			j++
 		}
-		value := tag[i:j]
+		value := wholeTag[i:j]
+		// remove trailing escaped quote if present
+		value = strings.TrimSuffix(value, `\"`)
```
mlevesquedion, line 134: Highlighting this change with a comment since it is lost in a sea of renamings.

configuration/README.md
```diff
@@ -28,6 +28,25 @@ Sources are identified via regexp according to package, type, and field names.
...
+}
+```
+
+The tag `levee:"source"` is built-in. Additional tags may be identified via regexp:
+```json
+{
+	"FieldTags": [
+		{
+			"Key": "<key regexp>",
+			"Val": "<value regexp>"
```
PurelyApplied, line 44: Do we want to do regexp here rather than explicit string matching?  I feel like tags are less volatile, and tags a user will want to match is likely to be a minimal set.  I could go either way on it, though.

internal/pkg/config/config.go
```diff
@@ -41,6 +41,32 @@ type Config struct {
 	Sources    []sourceMatcher
 	Sinks      []funcMatcher
 	Sanitizers []funcMatcher
+	FieldTags  []fieldTagMatcher
+}
+
+type fieldTagMatcher struct {
+	Key regexp.Regexp
+	Val regexp.Regexp
```
PurelyApplied, line 49: These fields do not need to be exported.

internal/pkg/fieldtags/testdata/src/tests/test.go
```diff
@@ -18,6 +18,7 @@ type Person struct {
 	password, creds      string      `levee:"source"`               // want "tagged field: password, creds"
 	secret               string      `json:"secret" levee:"source"` // want "tagged field: secret"
 	another              interface{} "levee:\"source\""             // want "tagged field: another"
+	hasCustomFieldTag    string      `example:"sensitive"`          // want "tagged field: hasCustomFieldTag"
```
PurelyApplied, line 21: I'd like to also see a test that includes something like
```golang
`example:"sensitive,custom,additionaltag"`
```

internal/pkg/fieldtags/analyzer.go
PurelyApplied: I think I'm going to play the regexp golf for this method after this gets submitted.

internal/pkg/fieldtags/testdata/src/tests/test.go
PurelyApplied, line 21: I don't know if we care, but this file currently fails `go vet -structtag`.  It's against convention / `reflect.StructTag` to tag unexported fields.

PurelyApplied (changes requested):
The only real requested change is for that additional test, just to cover bases.  👍

internal/pkg/fieldtags/analyzer.go
mlevesquedion: Sounds like fun, curious to see what you come up with :smile:

configuration/README.md
mlevesquedion: I feel the same.

I was essentially going for consistency with the rest of the config. Since everything else is a regexp, users may expect this to be as well. Regexp also allow users more flexibility if they need it.

I've added "RE" to the end of the JSON keys to make it more explicit that these are regexp.

WDYT?

internal/pkg/config/config.go
mlevesquedion: I think they have to be exported for the JSON Unmarshaling to work :thinking:.

There may be a way around that. If that sounds like something we want to do, I will fix the other matches in another PR (their fields are also exported).

internal/pkg/fieldtags/testdata/src/tests/test.go
mlevesquedion: Interesting. Do you have a link for that? I would be curious to read more. This convention sounds strange to me, I thought struct tags were intended for adding metadata in general.

While looking for a link I looked at `reflect.StructTag` and I think it might be a good idea to rely on that instead of duplicating the code. I don't remember why I didn't use that initially.

mlevesquedion: I think this is outside the scope of the current PR, but reasonable as a drive-by. Done.

configuration/README.md
PurelyApplied: Either way.  We're going to end up revisiting it when config is redesigned in #89 et al in any case.  Something I'll keep in mind then.

internal/pkg/config/config.go
PurelyApplied: Hmmm... That could maybe make sense... Maybe I'm getting my antipatterns backwards, I just thought it seemed strange to export fields of an unexported type.

I suppose of marshalling infers the underlying struct literal, it wouldn't need the type exported for `reflect` to do its job.  But it would need the fields exported in order to set those values.

Weird.  But okay.

internal/pkg/fieldtags/testdata/src/tests/test.go
PurelyApplied: I dunno.  Implementation and the testing contract changed.  I think that puts it within the scope of this PR.  `</hair_splitting>`

PurelyApplied: Oh, I just ran it locally, and [the go vet docs](https://golang.org/cmd/vet/) claim

> structtag    check that struct field tags conform to reflect.StructTag.Get

I only bring up `StructTag` because it seems to have defined the community standard.

mlevesquedion: Fair enough. This should already have been tested anyway.

mlevesquedion:
Thanks for your review!

I've added the test, let me know if you want me to do other changes.

PurelyApplied (approved):
Now cross-over with #129, where you commented maybe literals would be better since it would allow us to use `StructTag.Get`.  Your call, I don't have a horse in the race either way.

mlevesquedion:
I think I care more about being able to use `StructTag`, so I've changed the regexps to literals.

PR #129, opened 2020-09-23
Use reflect.StructTag.Get in fieldtag logic.

PurelyApplied:
It looks like `StructTag` is just a named `string` and we can co-opt `StructTag.Get`.

While `go vet -structtag` doesn't complain about it, it does interact strangely with escaped tags like `"levee:\"source\""`.  But it is much cleaner to use it as the community standard truth.

Opening this PR as a draft until conflicting PRs are merged.

### DIFF
--- a/internal/pkg/config/config.go
+++ b/internal/pkg/config/config.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"go/types"
 	"io/ioutil"
+	"reflect"
 	"strings"
 	"sync"
 
@@ -50,21 +51,28 @@ type fieldTagMatcher struct {
 	Val string
 }
 
-func (ftm fieldTagMatcher) matches(key, val string) bool {
-	return ftm.Key == key && ftm.Val == val
-}
-
 // IsSourceFieldTag determines whether a field tag made up of a key and value
 // is a Source.
-func (c Config) IsSourceFieldTag(key, val string) bool {
+func (c Config) IsSourceFieldTag(tag string) bool {
+	// tag is the entire, quote-wrapped string, e.g. "`levee:...`"
+	// Trim for use by reflect.StructTag
+	if len(tag) < 2 {
+		return false
+	}
+
+	st := reflect.StructTag(tag[1 : len(tag)-1])
+
 	// built in
-	if key == "levee" && val == "source" {
+	if st.Get("levee") == "source" {
 		return true
 	}
 	// configured
 	for _, ft := range c.FieldTags {
-		if ft.matches(key, val) {
-			return true
+		val := st.Get(ft.Key)
+		for _, v := range strings.Split(val, ",") {
+			if v == ft.Val {
+				return true
+			}
 		}
 	}
 	return false
--- a/internal/pkg/config/config.go
+++ b/internal/pkg/config/config.go
@@ -21,6 +21,7 @@ import (
 	"go/types"
 	"io/ioutil"
 	"reflect"
+	"strconv"
 	"strings"
 	"sync"
 
@@ -54,13 +55,10 @@ type fieldTagMatcher struct {
 // IsSourceFieldTag determines whether a field tag made up of a key and value
 // is a Source.
 func (c Config) IsSourceFieldTag(tag string) bool {
-	// tag is the entire, quote-wrapped string, e.g. "`levee:...`"
-	// Trim for use by reflect.StructTag
-	if len(tag) < 2 {
-		return false
+	if unq, err := strconv.Unquote(tag); err == nil {
+		tag = unq
 	}
-
-	st := reflect.StructTag(tag[1 : len(tag)-1])
+	st := reflect.StructTag(tag)
 
 	// built in
 	if st.Get("levee") == "source" {
--- a/internal/pkg/config/fieldtags_test.go
+++ b/internal/pkg/config/fieldtags_test.go
@@ -30,41 +30,36 @@ func TestFieldTagsIdentification(t *testing.T) {
 
 	cases := []struct {
 		desc string
-		key  string
-		val  string
+		tag  string
 		want bool
 	}{
 		{
 			"built-in field tag",
-			"levee",
-			"source",
+			"`levee:\"source\"`",
 			true,
 		},
 		{
 			"custom field tag",
-			"example",
-			"sensitive",
+			"`example:\"sensitive\"`",
 			true,
 		},
 		{
 			"different tag key",
-			"notexample",
-			"sensitive",
+			"`notexample:\"sensitive\"`",
 			false,
 		},
 		{
 			"different tag value",
-			"example",
-			"notsensitive",
+			"`example:\"notsensitive\"`",
 			false,
 		},
 	}
 
 	for _, tt := range cases {
 		t.Run(tt.desc, func(t *testing.T) {
-			got := config.IsSourceFieldTag(tt.key, tt.val)
+			got := config.IsSourceFieldTag(tt.tag)
 			if got != tt.want {
-				t.Errorf("config.IsSourceFieldTag(%q, %q) == %v, want %v", tt.key, tt.val, got, tt.want)
+				t.Errorf("config.IsSourceFieldTag(%q) == %v, want %v", tt.tag, got, tt.want)
 			}
 		})
 	}
--- a/internal/pkg/config/fieldtags_test.go
+++ b/internal/pkg/config/fieldtags_test.go
@@ -53,6 +53,11 @@ func TestFieldTagsIdentification(t *testing.T) {
 			"`example:\"notsensitive\"`",
 			false,
 		},
+		{
+			"escaped tag accepted",
+			`"levee:\"source\""`,
+			true,
+		},
 	}
 
 	for _, tt := range cases {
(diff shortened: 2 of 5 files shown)


internal/pkg/fieldtags/testdata/src/tests/test.go
```diff
@@ -15,10 +15,9 @@
...
 type Person struct {
-	password, creds      string      `levee:"source"`               // want "tagged field: password, creds"
-	secret               string      `json:"secret" levee:"source"` // want "tagged field: secret"
-	another              interface{} "levee:\"source\""             // want "tagged field: another"
-	name                 string      `some_key:"non_secret"`
-	spaceAfterFinalQuote string      `key:"value" `
+	Password, Creds      string      `levee:"source"`               // want "tagged field: Password, Creds"
+	Secret               string      `json:"secret" levee:"source"` // want "tagged field: Secret"
+	Name                 string      `some_key:"non_secret"`
+	SpaceAfterFinalQuote string      `key:"value" `
```
PurelyApplied, line 21: Exported fields to satisfy `go vet -structtag`.

internal/pkg/fieldtags/analyzer.go
```diff
@@ -91,53 +91,13 @@ func (sp *sourcePatterns) isSource(field *ast.Field) bool {
...
-		}
-
-		i = j + 2 // skip closing quote and space
-		j = i
+	if len(tag) <= 2 {
+		return false
 	}
 
-	return false
+	// Tags are a string literal, beginning and ending with '`' or '"'
```
mlevesquedion, line 98: I think it would be good to have that comment higher up in the method, since if I understand correctly it explains why we want to return if `len(tag) <= 2`.

internal/pkg/fieldtags/analyzer.go
```diff
@@ -91,53 +91,13 @@ func (sp *sourcePatterns) isSource(field *ast.Field) bool {
...
+	tag = tag[1:len(tag)-1]
+	return reflect.StructTag(tag).Get("levee") == "source"
```
mlevesquedion, line 100: This is very nice. Glad we're getting rid of the mess above!

I am just now realizing that this does lock us in to using explicit strings, though, at least for the `Key`. I think that is fine, just something we will need to indicate clearly in the documentation.

internal/pkg/fieldtags/testdata/src/tests/test.go
```diff
@@ -15,10 +15,9 @@
 package fieldtags
 
 type Person struct {
-	password, creds      string      `levee:"source"`               // want "tagged field: password, creds"
-	secret               string      `json:"secret" levee:"source"` // want "tagged field: secret"
-	another              interface{} "levee:\"source\""             // want "tagged field: another"
```
mlevesquedion, line 20: I think we should still support this case, since `structtag` doesn't complain about it and `StructTag.Get` supports it, e.g. see the example program below. If we use `strconv.Unquote` we can convert the string literal from the `ast` into something `StructTag.Get` can handle, e.g. `"levee:\"source\""` becomes `levee:"source"` with `Unquote`.

Example program showing how StructTag handles the case above (`f.Tag` is a `StructTag`):
```go
package main

import (
	"fmt"
	"reflect"
)

func main() {
	f, _ := reflect.TypeOf(myStruct{}).FieldByName("myField")
	fmt.Println(f.Tag)
}

type myStruct struct {
	myField string "key:\"value\""
}
```
(this prints `key:"value"`)

mlevesquedion (changes requested):
I think using `StructTag` makes sense. Good stuff!

internal/pkg/fieldtags/testdata/src/tests/test.go
```diff
@@ -15,12 +15,11 @@
...
-	secret                   string      `json:"secret" levee:"source"` // want "tagged field: secret"
-	another                  interface{} "levee:\"source\""             // want "tagged field: another"
-	hasCustomFieldTag        string      `example:"sensitive"`          // want "tagged field: hasCustomFieldTag"
-	hasTagWithMultipleValues string      `levee:"value0,source,value2"` // want "tagged field: hasTagWithMultipleValues"
-	name                     string      `some_key:"non_secret"`
-	spaceAfterFinalQuote     string      `key:"value" `
+	password, creds          string `levee:"source"`               // want "tagged field: password, creds"
+	secret                   string `json:"secret" levee:"source"` // want "tagged field: secret"
+	hasCustomFieldTag        string `example:"sensitive"`          // want "tagged field: hasCustomFieldTag"
+	hasTagWithMultipleValues string `example:"sensitive,verbose"`  // want "tagged field: hasTagWithMultipleValues"
```
PurelyApplied, line 21: This case changed because we do not support multiple `levee:` values.  We could, but it would need to be accounted for above.

internal/pkg/fieldtags/testdata/src/tests/test.go
```diff
@@ -15,12 +15,11 @@
 package fieldtags
 
 type Person struct {
-	password, creds          string      `levee:"source"`               // want "tagged field: password, creds"
-	secret                   string      `json:"secret" levee:"source"` // want "tagged field: secret"
-	another                  interface{} "levee:\"source\""             // want "tagged field: another"
```
PurelyApplied, line 20: This PR yields no longer supporting these escaped-quote style of tags.  We could jump through some hoops to do so if necessary, but if the point was to keep the lookup simple, I'd rather wait until someone complains that try to preempt the issue.

PurelyApplied:
Sorry for the forced push, but the changes from the initial were sweeping enough that an entire review is rather necessary at this point.

internal/pkg/fieldtags/testdata/src/tests/test.go
PurelyApplied: I didn't know about `strconv.Unquote`!  That keeps things nice and tidy then.

internal/pkg/fieldtags/testdata/src/tests/test.go
mlevesquedion: I think I'd prefer >2 values with the target value in the middle or at the end (to test that we're actually looping), but this is fine.

mlevesquedion (commented):
I think this looks good. I like how you made the Config take the whole tag, I think that makes sense.

Waiting to see how it looks with `strconv.Unquote` before I approve. I think it'll look cleaner + allow us to handle the case that we would be dropping otherwise, but I may be wrong.

PurelyApplied:
> I'll get to it tomorrow.
What a hopeful person I was two weeks ago.

internal/pkg/config/config.go
```diff
@@ -50,21 +52,25 @@ type fieldTagMatcher struct {
...
-func (ftm fieldTagMatcher) matches(key, val string) bool {
-	return ftm.Key == key && ftm.Val == val
-}
-
 // IsSourceFieldTag determines whether a field tag made up of a key and value
 // is a Source.
-func (c Config) IsSourceFieldTag(key, val string) bool {
+func (c Config) IsSourceFieldTag(tag string) bool {
+	if unq, err := strconv.Unquote(tag); err == nil {
+		tag = unq
```
mlevesquedion, line 59: Never occurred to me to do something like that (writing to a parameter). Reads well, I like it.

internal/pkg/config/fieldtags_test.go
```diff
@@ -30,41 +30,41 @@ func TestFieldTagsIdentification(t *testing.T) {
...
-			"notsensitive",
+			"`example:\"notsensitive\"`",
 			false,
 		},
+		{
+			"escaped tag accepted",
+			`"levee:\"source\""`,
+			true,
+		},
 	}
```
mlevesquedion, line 61: I think it would be good to have a test case that covers the `for _, v := range strings.Split(val, ",") {` code, now that the logic for multi-value tags is in config.

mlevesquedion:
Haha, you couldn't have known. It's good to be optimistic.

I think this looks very good!

One last thing: I think we need a test for a multi-valued tag in `config/fieldtags_test.go`.

internal/pkg/config/fieldtags_test.go
```diff
@@ -30,41 +30,71 @@ func TestFieldTagsIdentification(t *testing.T) {
...
+			"`example:\"foo,bar,baz\"`",
+			false,
+		},
+		{
+			"multiple values, with target",
+			"`example:\"foo,sensitive,bar\"`",
+			true,
+		},
+		{
+			"multiple key value sets, no target",
```
mlevesquedion, line 72: The description does not seem to match the data in the test case. The data is the same as the next case, though the description is different.

mlevesquedion:
Good extra cases. LGTM, expect for one case where the description does not seem to match the data. The description says "no target" but there is a target and the expected result is `true`.

mlevesquedion (approved):
Approved, assuming you will fix the test case that has a typo.

internal/pkg/config/fieldtags_test.go
PurelyApplied: Also, the same test case twice in a row.

Copy-Paste, my old nemesis...

Issue #92, opened 2020-09-10
Detect field assignment to zero-value as sanitization


PurelyApplied:
Consider the following:

```golang
type Source struct {
  secret string // `levee:"source"`
}

func Bar(s Source) {
  s.secret = ""
  Sink(s)  // This should not emit a Diagnostic.
}
```

Explicitly setting to zero-values all fields containing source data should be recognized as sanitization.

PurelyApplied:
Optionally, a user may wish to provide sanctioned non-zero values, e.g. `s.secret = "REDACTED"`.

PurelyApplied:
Cross-referencing #97, a wrapping type should be able to likewise sanitize via either of `Container.content = nil` or `Container.content.secret = ""`.

mlevesquedion:
I think we could handle some of these cases without too much headache, but in general this looks like it's going to be very tricky.

If we do a good job with it though, eventually we may be able to identify sanitizers automatically.

PR #131, opened 2020-09-24
Add tests for embedded structs + fields

mlevesquedion:
This PR adds tests for embedded structs and fields.

Structs that embed Source structs are not currently considered Sources themselves. This is why 2 of the tests have `TODO`s. This is in relation to issue #97.

The other tests are made to pass by a small change to `source.go`.

- [x] Tests pass


PurelyApplied (approved):
You might add some trivial "actual" source data, since detecting zero-values as implicitly safe is on the roadmap in #92.  Then again, I'm sure there are plenty of tests that will need to be modified and added as part of that, so it's not a big deal.

mlevesquedion:
Gotcha. I think I'm going to put that off for now. I agree that these should not actually produce reports, it will be a nice surprise to have those tests fail once we implement zero-values.

PR #133, opened 2020-09-29
Debugging: improve SSA rendering and add tests

mlevesquedion:
This PR improves the SSA rendering by outputting the function's signature and aligning instruction indexes across a function.

This PR adds tests to cover existing + new functionality. The tests use "golden" files, defining expectations for the output of `render.SSA` for each function in `testdata/tests.go`.

- [x] Tests pass


PurelyApplied (approved):
Nice to see tests in this corner.  I have mixed feelings about golden files in general, but they're not so bad here I think.

PR #136, opened 2020-10-05
Example config: use log instead of k?log$

mlevesquedion:
The current config is slightly outdated. I think we should use just plain "log", as the substring "log" should occur in all logging packages (we may care about more than `klog`) and is unlikely to occur outside of logging packages.

- [x] Tests pass

### DIFF
--- a/configuration/example-config.json
+++ b/configuration/example-config.json
@@ -23,7 +23,7 @@
   ],
   "Sinks": [
     {
-      "PackageRE": "k?log$",
+      "PackageRE": "log",
       "ReceiverRE": "",
       "MethodRE": "Info|Warning|Error|Fatal|Exit"
     }


configuration/example-config.json
```diff
@@ -23,7 +23,7 @@
   ],
   "Sinks": [
     {
-      "PackageRE": "k?log$",
```
PurelyApplied, line 26: Did we want that terminal anchor?  Will that not overlook, e.g., `klog/v1`?

mlevesquedion: I think we indeed do not want the `$`, since it does overlook `klog/v1`.

PurelyApplied: Oh.  Monday mornings.  I was reading this diff absolutely backwards, like you were adding the anchor.

PurelyApplied (approved):

Issue #99, opened 2020-09-10
Adopt some manner of cross-function labeling to identify interprocedural propagation into non-source types.


PurelyApplied:
Similar to #98, consider the following:

```golang
func Sinker(st string) {
  Sink(st)
}

func Foo(s Source){
  Sinker(s.String())
}
```

During analysis of `Sinker`, the concrete type of `st` is not concerning to reach a sink.  However, without also configuring `Sinker` as a sink, the taint propagation from `s -> st := s.String() -> Sink(st)` is lost.

Some preliminary work has been done in #57, #58, #60.

PR #137, opened 2020-10-05
Consider source collections to be sources

mlevesquedion:
## Context

This PR is inspired by a case I stumbled upon recently that the analyzer currently isn't catching:

```go
func Oops() {
  ss := GetSources()
  core.Sink(ss) // TODO: want "a source has reached a sink" 
}

func GetSources() []Source {
  return nil // omitted for simplicity
}
```

Since we don't know what the `GetSources` call is doing, we should assume that it is returning a slice containing at least 1 unsanitized `Source`. This may not be the case: the slice may be empty, or it may contain only unsanitized sources. However, without interprocedural analysis (#99) and granular taint tracking for collections (#47), and likely other features, in general we should assume that collections that hold Sources are themselves Sources, such that 

## The PR

This PR adds support for identification of Source-holding collections as Sources. This entails:
* Adding tests for the `levee` and `source` analyzers. (As a "side-effect", we get a few currently failing tests to pass, as well.)
* Changes to `source.go`. Most of these, especially the `isSourceType` function, should be in the `sourcetype` analyzer, but it is not yet consumed anywhere, so currently I think this is the best place to make those changes. (Incidentally, I think having this functionality makes a good case for updating and consuming the `sourcetype` analyzer, so I am planning on doing that next.)

- [x] Tests pass

### DIFF
--- a/internal/pkg/source/source.go
+++ b/internal/pkg/source/source.go
@@ -380,7 +380,7 @@ func sourcesFromBlocks(fn *ssa.Function, conf classifier) []*Source {
 			// source obtained through a field or an index operation
 			case *ssa.Field, *ssa.FieldAddr, *ssa.IndexAddr, *ssa.Lookup:
 
-			// source map (arrays and slices have regular Allocs)
+			// source chan or map (arrays and slices have regular Allocs)
 			case *ssa.MakeMap, *ssa.MakeChan:
 			}
 
--- a/internal/pkg/source/source.go
+++ b/internal/pkg/source/source.go
@@ -314,11 +314,8 @@ func identify(conf classifier, ssaInput *buildssa.SSA) map[*ssa.Function][]*Sour
 func sourcesFromParams(fn *ssa.Function, conf classifier) []*Source {
 	var sources []*Source
 	for _, p := range fn.Params {
-		switch t := p.Type().(type) {
-		case *types.Pointer:
-			if n, ok := t.Elem().(*types.Named); ok && conf.IsSource(n) {
-				sources = append(sources, New(p, conf))
-			}
+		if isSourceType(conf, p.Type()) {
+			sources = append(sources, New(p, conf))
 		}
 	}
 	return sources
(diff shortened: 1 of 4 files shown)


internal/pkg/source/source.go
```diff
@@ -382,17 +379,39 @@ func sourcesFromBlocks(fn *ssa.Function, conf classifier) []*Source {
...
+		return isSourceType(c, tt.Elem())
+	case *types.Chan:
+		return isSourceType(c, tt.Elem())
+	case *types.Map:
+		key := isSourceType(c, tt.Key())
+		elem := isSourceType(c, tt.Elem())
+		return key || elem
+	default:
+		u := tt.Underlying()
+		return c.IsSource(tt) || (u != tt && isSourceType(c, u))
```
PurelyApplied, line 411: I might prefer to be explicit in which types do or don't get passed to `isSourceType`, rather than rely on the implementation detail of which are self-underlying.

```golang
	case *types.Basic, *types.Interface, *types.Tuple, *types.Struct:
		return false
	default:
		u := tt.Underlying()
		fmt.Printf("%v, underlying=%v\n", tt, u)
		return c.IsSource(tt) || isSourceType(c, u)
	}
```

At the very least, the non-obvious `u != tt` condition deserves a comment.

PurelyApplied (changes requested):
I would've sworn we had an Issue open for this, but apparently not?  Maybe I'm conflating it with one of the other `inference`-tagged items.

mlevesquedion:
Same, I thought we had an issue and I meant to mention that in the description but I forgot.

internal/pkg/source/source.go
mlevesquedion: Agreed, I think that reads better.

Is the `Printf` as a suggestion to make sure we weren't missing anything? (I gave it a shot and it looks good to me). Or are you suggesting that we should include that in the code?

PurelyApplied (approved):

Issue #94, opened 2020-09-10
Update documentation - contributing / reporting false positives


PurelyApplied:
Document in README that users are welcome to open an Issue when they experience unexpected behavior, false positives or false negatives, etc.

mlevesquedion:
If we really want to maximize our chances of getting reports, when producing a report we might even want to give users a link that  directly creates a new issue on GitHub.

PR #138, opened 2020-10-05
Introduce issue template for false positive reports

mlevesquedion:
In relation to #94. I intend to add something to the README with a link that points to a new issue using this template.

### DIFF
--- a/.github/ISSUE_TEMPLATE/false-positive.md
+++ b/.github/ISSUE_TEMPLATE/false-positive.md
@@ -1,0 +1,20 @@
+---
+name: False positive
+about: This template is used to describe false positive reports.
+title: "[FALSE POSITIVE]: "
+labels: false positive
+assignees: mlevesquedion
+
+---
+
+## False positive report
+
+Use this issue template to describe a false positive report produced by the analyzer. A false positive report is a report produced by the analyzer on code that you consider to be safe. For example, if the analyzer were to produce a report on the following piece of code, then that would be a false positive:
+
+```go
+func NothingWrongHere() {
+    Sink(nil) // we don't want a report here, since the value that reached the sink is not sensitive
+}
+```
+
+Please make it as easy as possible for us to reproduce what you observed. If possible, provide the exact configuration and code on which the analyzer produced a report. If you cannot do that (e.g., the code is closed source), please provide a simplified example and confirm that it also yields the false positive report.
--- a/.github/ISSUE_TEMPLATE/false-positive.md
+++ b/.github/ISSUE_TEMPLATE/false-positive.md
@@ -1,6 +1,6 @@
 ---
 name: False positive
-about: This template is used to describe false positive reports.
+about: Report a false positive.
 title: "[FALSE POSITIVE]: "
 labels: false positive
 assignees: mlevesquedion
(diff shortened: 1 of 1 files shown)


PurelyApplied (commented):
Expanding templates is a good idea!

Would it be better to have a more generic `User Issue` template, rather than (or in addition to) specifically false positives?

.github/ISSUE_TEMPLATE/false-positive.md
```diff
@@ -0,0 +1,21 @@
+---
+name: False positive
+about: This template is used to describe false positive reports.
+title: "[FALSE POSITIVE]: "
+labels: false positive
+assignees: mlevesquedion
```
PurelyApplied, line 6: I'd like to see consistency across the templates.  If you want label `false positive` here, use `false negative` in the other.  Alternatively, both are bugs.

Do you mean for these to autoassign to you?

mlevesquedion:
Good point. My aim with this series of PR's was to tackle #94, so originally I was going to do just a false positives template. But then it occurred to me a that template for false negatives would also be useful. I'll create a template for general issues shortly.

.github/ISSUE_TEMPLATE/false-positive.md
mlevesquedion: Sounds good. I think more granularity could be useful, so I'll create a false negative tag.

> Do you mean for these to autoassign to you?
Yes. I figured it would be good to have it autoassign to _someone_ so we would know about issues as soon as possible. I put my username as a default. Since we can have more than one autoassignee, would you like to be autoassigned as well? Or some other method?

mlevesquedion:
This PR is a precursor to #147.

.github/ISSUE_TEMPLATE/false-positive.md
PurelyApplied: It's something we could iterate on.  We could start more coarse, and if we find ourselves with a large influx of reports, add granularity down the line.  Hopefully, they're not tags that are going to need much use.

PurelyApplied: Cross-referencing the other PRs again, they're also fine as is.  We're not overburdened with tags right now, either.

PurelyApplied: (Don't you love how a single carriage return doesn't break the quote formatting?  I banner-blind'd that block.)

I've set a `Watch` on the repo, so I think I should get a notification anytime an Issue is opened.  My concern is mostly a logistical one.  Since GitHub doesn't have a way to distinguish between `Assigned to` and `Actively worked on by`, there wouldn't be a clear "this is a ticket that could be started on by anyone" state of having no assignee.  So while it's fine to auto-assign to you while it's just us working in this space, since I will just ping you and ask if I should pick up an Issue assigned to you.

\shrug

mlevesquedion: Oh wow sorry about the quote block. Fixed.

I think for now I'll set up a `Watch` too and remove the auto-assign since that will allow us to spot which issues we haven't looked into yet.

PurelyApplied (approved):

PR #142, opened 2020-10-07
Do not traverse to boolean from comma-ok map lookup

mlevesquedion:
Changes from 2 recent PR's (#134 and #137) caused a test failure to land on master due to an unforeseen interaction.

#134 added the following test:
```go
func TestOnlySourceExtractIsTaintedFromLookup() {
	s, ok := map[string]core.Source{}[""]
	core.Sink(ok)
}
```
Which made sense given that the new functionality in that PR was to avoid traversing from an `Extract` to any of its `Operands`, e.g. not just `Calls` (what used to be the case), but also `Lookups` and any other instruction.

#137 made collections containing sources be recognized as sources themselves, so e.g. in the test above, `map[string]core.Source{}` is a source. Sadly, this meant the test was now failing, because that opened a path to get from the `map` to the `core.Sink` call via the `ok` value.

This PR fixes the interaction by preventing traversal to the `ok` value in a map lookup.

Relevant graph to help understand why the test was passing in #134 but failing once #137 was merged:
image

- [x] Tests pass

### DIFF
--- a/internal/pkg/source/source.go
+++ b/internal/pkg/source/source.go
@@ -136,6 +136,8 @@ func (s *Source) visitReferrers(n ssa.Node, maxInstrReached map[*ssa.BasicBlock]
 // Specifically, we want to avoid referrers that shouldn't be visited, e.g.
 // because they would not be reachable in an actual execution of the program.
 func (s *Source) referrersToVisit(n ssa.Node, maxInstrReached map[*ssa.BasicBlock]int) (referrers []ssa.Instruction) {
+	_, visitingFromLookup := n.(*ssa.Lookup)
+
 	if n.Referrers() == nil {
 		return
 	}
@@ -156,6 +158,11 @@ func (s *Source) referrersToVisit(n ssa.Node, maxInstrReached map[*ssa.BasicBloc
 			if i < maxInstrReached[r.Block()] {
 				continue
 			}
+		}
+
+		// do not taint the boolean produced by a comma-ok map lookup
+		if ex, ok := r.(*ssa.Extract); ok && ex.Index == 1 && visitingFromLookup {
+			continue
 		}
 
 		if fa, ok := r.(*ssa.FieldAddr); ok && !s.config.IsSourceFieldAddr(fa) {


PurelyApplied (changes requested):
This feels very special-case, brittle, and likely to cover more than intended.  Nothing in the code indicates a check for boolean, for instance, though the comments specify `foo, ok := bar()`.

I'd rather just revert the offending change and correcting the issue in that diff set.  I've opened that at #143 .

mlevesquedion:
I agree that it's a special case, but it seems to me that that is what the `referrersToVisit` function is for.

This only covers the case where we are doing a `Lookup` with a `, ok` pattern. From the `ssa` package docs:
```go
// The Lookup instruction yields element Index of collection X, a map
// or string.  Index is an integer expression if X is a string or the
// appropriate key type if X is a map.
//
// If CommaOk, the result is a 2-tuple of the value above and a
// boolean indicating the result of a map membership test for the key.
// The components of the tuple are accessed using Extract.
//
// Pos() returns the ast.IndexExpr.Lbrack, if explicit in the source.
//
// Example printed form:
// 	t2 = t0[t1]
// 	t5 = t3[t4],ok
//
```

I don't think we would ever want to propagate to the boolean.

mlevesquedion:
As proposed in #143, reverting the changes would mean introducing a bug, namely tainting booleans in lookups.

I don't have a hard preference, but I think the current PR should be preferred, since I think it is an appropriate fix for the issue.

PurelyApplied:
There will always be bugs.  What we want is to unblock `master`, and a revert is the fastest path to that goal.  This code needs review and better testing, and we should unblock `master` in the interm.

I agree that we shouldn't propagate to a boolean.  I'm saying that this introduces logic that goes beyond that supposition.  At a glance, it makes me thing that `foo, bar, baz = := func() (Source, string, string)` isn't going to work correctly, since you're hardcoding an index and assuming it's a bool.  That needs a test.  There are likely other tests necessary.  But that longer discussion best belongs to a revisit of #134 after `master` is unblocked.

mlevesquedion:
SG, thanks for explaining that.

mlevesquedion:
Closed in favor of #144.

PR #146, opened 2020-10-07
Do not traverse to booleans

mlevesquedion:
This PR builds on #145.

I am opening this PR to fix a bug that we currently have: when there is a comma-ok form, e.g. `s, ok := map[string]Source{}[""]`, we are currently propagating from the `Lookup` instruction to the `boolean` `Extract` (the `ok` value), which does not make sense: the boolean isn't tainted. This also occurs with `chan` receives, which appear as `UnOp` instruction. To my knowledge, no other `UnOps` can use the comma-ok form.

Since booleans can't meaningfully be tainted, this PR modifies the traversal to avoid booleans altogether.

- [x] Tests pass

### DIFF
--- a/internal/pkg/source/source.go
+++ b/internal/pkg/source/source.go
@@ -218,20 +218,20 @@ func (s *Source) canReach(start *ssa.BasicBlock, dest *ssa.BasicBlock) bool {
 }
 
 func (s *Source) visitOperands(n ssa.Node, operands []*ssa.Value, maxInstrReached map[*ssa.BasicBlock]int, lastBlockVisited *ssa.BasicBlock) {
-	_, visitingFromExtract := n.(*ssa.Extract)
+	// Do not visit Operands if the current node is an Extract.
+	// This is to avoid incorrectly tainting non-Source values that are
+	// produced by an Instruction that has a Source among the values it
+	// produces, e.g. a call to a function with a signature like:
+	// func NewSource() (*core.Source, error)
+	// Which leads to a flow like:
+	// Extract (*core.Source) --> Call (NewSource) --> error
+	if _, ok := n.(*ssa.Extract); ok {
+		return
+	}
 
 	for _, o := range operands {
 		n, ok := (*o).(ssa.Node)
 		if !ok {
-			continue
-		}
-
-		// Do not visit a Call if the current node is an Extract.
-		// This is to avoid incorrectly tainting non-Source values that are
-		// returned from a call that has a Source among its return values,
-		// e.g. a call to a function with a signature like:
-		// func CreateSource() (core.Source, error)
-		if _, ok := (*o).(*ssa.Call); visitingFromExtract && ok {
 			continue
 		}
 
--- a/internal/pkg/source/source.go
+++ b/internal/pkg/source/source.go
@@ -86,6 +86,13 @@ func (s *Source) dfs(n ssa.Node, maxInstrReached map[*ssa.BasicBlock]int, lastBl
 	if s.marked[n] {
 		return
 	}
+	// booleans can't meaningfully be tainted
+	if v, ok := n.(ssa.Value); ok {
+		if basic, ok := v.Type().(*types.Basic); ok && basic.Info() == types.IsBoolean {
+			return
+		}
+	}
+
 	mirCopy := map[*ssa.BasicBlock]int{}
 	for m, i := range maxInstrReached {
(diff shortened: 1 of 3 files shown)


internal/pkg/source/source.go
```diff
@@ -162,11 +162,35 @@ func (s *Source) referrersToVisit(n ssa.Node, maxInstrReached map[*ssa.BasicBloc
...
+	// the boolean in a comma-ok extract is always at the 1st index
+	if e.Index != 1 {
+		return false
+	}
+	switch n.(type) {
+	// Lookup is for a map lookup, e.g.
+	// maybeValue, ok := myMap[myKey]
+	// UnOp is for a chan recv, e.g.
+	// maybeValue, ok := <-myChan
+	case *ssa.Lookup, *ssa.UnOp:
```
PurelyApplied, line 187: The following `Instruction` types need to be unpacked by an `Extract`.

* `ssa.TypeAssert` appears to be the only other type that stores a `CommaOk bool` field.  Should we add that to this set?

* `Call` and `Return` bundle return types for an arbitrary function call.  We shouldn't make a value judgement on any particular element based on return position.  We might include them in the case switch if we were being explicit in our enumeration.  Either way.

* `ssa.Next` reutrns a 3-tuple, where we should also not propagate to the `bool`.  This could be a part of a future PR / deeper future examination.

* `ssa.Select` contains a much more complex `n+2` tuple, but I'm not sure we have any `select` tests yet.  That probably puts it as a candidate for a future PR.

internal/pkg/source/source.go
```diff
@@ -162,11 +162,35 @@ func (s *Source) referrersToVisit(n ssa.Node, maxInstrReached map[*ssa.BasicBloc
...
+		}
+
 		referrers = append(referrers, r)
 	}
 	return referrers
 }
 
+// shouldNotVisitExtract determines if the extract is extracting a boolean from
+// a comma-ok map lookup or chan recv, in which case we should not visit it,
+// because a boolean cannot meaningfully be tainted.
```
PurelyApplied, line 176: > because a boolean cannot meaningfully be tainted.

If this is our motivation, then do we care about the `Extract`?  Could we not just check for `bool` type and skip that irrespective of context?  I.e., considering this case:

```golang
func Foo(s Source) {
  var ok bool = s.IsReady()
  Sinkf("Source is ready: %v", ok)  // We would currently produce a report here, right?
  // ...
}
```

Even without an `Extract`, if we assert that a `bool` shouldn't carry taint, then we're missing this sort of case.

PurelyApplied (changes requested):
We've talked about it before, but we're edging into the territory of many special cases that would be better served by a large and close examination of each `Instruction` as to whether or not we want to consider taint propagation.  While a second "special case" in our code here isn't that bad, I feel like there probably shouldn't be a third in this style.

internal/pkg/source/source.go
mlevesquedion: Good point on `TypeAssert`, I figured it was alright because we're already avoiding propagating *from* an `Extract` to the instruction it is extracting from. So that covers e.g. `s, ok := (interface{}).(Source)`. However, if the interface value is tainted, then a different logic applies. I was missing a test case.

internal/pkg/source/source.go
mlevesquedion: I agree, I think the implementation for that will be simpler and it'll cover more cases.

My motivation for *not* banning all booleans was that it might allow us to discover more bugs in the future. That's a pretty weak motivation though.

I implemented the boolean check and added a few tests.

internal/pkg/source/source.go
```diff
@@ -400,6 +405,15 @@ func sourcesFromBlocks(fn *ssa.Function, conf classifier) []*Source {
 	return sources
 }
 
+func isBoolean(n ssa.Node) bool {
+	if v, ok := n.(ssa.Value); ok {
+		if basic, ok := v.Type().(*types.Basic); ok && basic.Info() == types.IsBoolean {
```
PurelyApplied, line 410: \<grumbles quietly about how a monad sure would be nice to have in Go.\>

PurelyApplied (approved):

mlevesquedion:
It wasn't in the PR template when this PR was opened, but I just want to mention that I ran this change on the `kubernetes/kubernetes` repo and it ran fine.

PR #140, opened 2020-10-05
Create false negative issue template

mlevesquedion:
I plan to reference this template in the documentation for #94.

PR for false positives template: #138.

PR #147, opened 2020-10-08
Update documentation: bug reports/false negatives/false positives

### LINKED ISSUE #94: Update documentation - contributing / reporting false positives

mlevesquedion:
Closes #94.

Depends on #138, #140 and #148 (those have been merged).

- [x] Tests pass

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,12 +32,14 @@ See [design/](design/README.md).
 
 See [configuration/](configuration/README.md) for configuration details.
 
-### False positives
+## Reporting bugs
 
 Static taint propagation analysis is a hard problem. In fact, it is [undecidable](https://en.wikipedia.org/wiki/Rice%27s_theorem). Concretely, this means two things:
 * False negatives: the analyzer may fail to recognize that a piece of code is unsafe.
 * False positives: the analyzer may incorrectly claim that a safe piece of code is unsafe. 
-Since our aim is to avoid credentials leaks, we care more deeply about false negatives. If you discover unsafe code that the analyzer is not recognizing as unsafe, please file a bug [here](http://www.google.com). We still care about false positives though, since in general they will waste someone's time. If the analyzer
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -32,6 +32,15 @@ See [design/](design/README.md).
 
 See [configuration/](configuration/README.md) for configuration details.
 
+## Reporting bugs
+
+Static taint propagation analysis is a hard problem. In fact, it is [undecidable](https://en.wikipedia.org/wiki/Rice%27s_theorem). Concretely, this means two things:
+* False negatives: the analyzer may fail to recognize that a piece of code is unsafe.
+* False positives: the analyzer may incorrectly claim that a safe piece of code is unsafe. 
+Since our aim is to avoid credentials leaks, we care more deeply about false negatives. If you discover unsafe code that the analyzer is not recognizing as unsafe, please open an issue [here](https://github.com/google/go-flow-levee/issues/new?template=false-negative.md). We still care about false positives though, since in general they will waste someone's time. If the analyzer produces a report for code that you consider to be safe, please open an issue [here](https://github.com/google/go-flow-levee/issues/new?template=false-positive.md).
```
PurelyApplied, line 40: > Since our aim is to avoid credentials leaks,

While this is the only at-scale application of `levee` we've done so far, I'd still prefer to keep documentation as application-agnostic as possible.  Consider a more general `Since taint propagation is often used as a security safeguard`?

PurelyApplied: >  still care about false positives though, since in general they will waste someone's time

Nit: this feels a bit passively voiced.  Consider `Conversely, false positives waste developer time and should also be addressed.`  Though I also tend to overly-academic writing, so maybe it's fine as is.

PurelyApplied (approved):
Minor rewording requested, but good to keep the docs fresh.

README.md
mlevesquedion: I think those are both great suggestions!

I've included them both. Thanks! :+1:

PR #150, opened 2020-10-09
Debug: render blocks using domination preorder

mlevesquedion:
While investigating a multi-block test case (#149), it occurred to me that the rendered SSA would be easier to read if the blocks appeared roughly in the same order as they do in the source code. Using `DomPreorder` gets us closer to that ideal.

- [x] Tests pass

### DIFF
--- a/internal/pkg/debug/render/ssa_test.go
+++ b/internal/pkg/debug/render/ssa_test.go
@@ -47,7 +47,7 @@ func TestSSA(t *testing.T) {
 		got := SSA(f)
 
 		if diff := cmp.Diff(want, got); diff != "" {
-			t.Errorf("render.SSA diff (-want +got):\n%s", diff)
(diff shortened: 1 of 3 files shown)


internal/pkg/debug/render/ssa_test.go
```diff
@@ -47,7 +47,7 @@ func TestSSA(t *testing.T) {
 		got := SSA(f)
 
 		if diff := cmp.Diff(want, got); diff != "" {
-			t.Errorf("render.SSA diff (-want +got):\n%s", diff)
+			t.Errorf("render.SSA(%s) diff (-want +got):\n%s", f.Name(), diff)
```
mlevesquedion, line 50: While fixing the test case, it occurred to me that this error message was missing something important...

PurelyApplied (approved):

PR #152, opened 2020-10-14
Add tests for conditionals in loops

mlevesquedion:
This PR adds tests in order to improve our coverage of some potentially tricky cases.

Consider, e.g. : 

```go
func TestTaintPropagationOverMultipleIterations() {
	var e1 interface{}
	var e2 interface{}
	for true {
		if true {
			e1 = core.Source{}
		} else {
			e2 = e1
		}
	}
	core.Sink(e1) // want "a source has reached a sink"
	core.Sink(e2) // want "a source has reached a sink"
}
```

In this example, `e2` can be tainted if the loop executes at least 2 iterations, with the `if`'s `then` branch being chosen first, then the `if`'s `else` branch. (In this case the `else` branch will never be taken because of the simplified condition, but the `ssa` module does not know that.)

The reason this works is that `e1` and `e2` in the above example will be related via `Phi` nodes, so the taint will be propagated from `e1` to `e2` while traversing.

## Future work

While thinking of test cases for this PR, I discovered what I consider to be a bug involving sanitization. Consider:

```go
func TestSanitizedBeforeExit() {
	var e interface{}
	for false {
		if true {
			e = nil
		} else {
			e = core.Source{}
		}
		e = core.Sanitize(e)[0]
	}
	core.Sink(e)
}
```

Clearly, it is impossible for an unsanitized value to reach the sink, since `e` is only tainted inside the loop, and each loop iteration ends by sanitizing `e`. However, currently we do produce a report for `core.Sink(e)`. The reason we produce a report here is that we traverse through sanitizing calls, which gives a path to `core.Sink(e)`. Furthermore, since the block in which sanitization occurs _does not dominate_ the block in which the call to `core.Sink` occurs, we do not consider `e` to have been properly sanitized, so we produce a report.

I believe I have found an appropriate fix and I intend to write a PR for it tomorrow.

- [x] Tests pass


PurelyApplied (approved):
I was vaguely under the impression that the SSA we traversed would be impacted by compiler optimization.  I'm surprised that many of these `if true` blocks don't appear to be simplified.  Although I might be thinking of the `go vet` case?  Does this ring any bells for you?

Either way, always good to see improved testing for gnarly cases.

mlevesquedion:
Can you remind me what the `go vet` case is? Doesn't ring any bells right now. 

I was a little surprised by the behavior involving `if true` as well. Somewhat related: if I use a purely syntactic endless loop (`for {`), without any `break` instructions in the body, the `ssa` module recognizes that the instructions beyond are unreachable and does not bother creating a block for them.

PurelyApplied:
> Can you remind me what the `go vet` case is? Doesn't ring any bells right now.

`go vet` is the more general analysis framework for, well, vetting Go code.  We can run analysis on a bit of code using `/path/to/levee -config foo.json package/to/analyze`, and that should work.  But if a consumer were to run several different kinds of analysis, `go vet` offers a way to merge effort and resolve conflicts in shared option names, or simply as a standard wrapper for any `Analyzer`.

So rather than directly invoking the built binary, we could use `go vet -vettool=/path/to/levee -config foo.json package/to/analyze`.  Notably, `go vet` also accepts build args, so one could conceivably even specify the compiler to use, etc.  I'm not sure if the `analysis` package would honor that request, but I would hope it does.  All this to say, I'm not sure if our tests are capturing behavior in that case.

I think within the scope of this PR, these tests are still a good addition.  If you wanted to remain robust, you could replace all the `if true` with `if flipCoin()` with some one-liner utility that dodges optimization, but honestly, I'm bet optimization would break a lot of our tests.

> if I use a purely syntactic endless loop [...], the ssa module [...] does not bother creating a block for them.

That's interesting.  I'll have to go compare some `ssadump`s.

mlevesquedion:
I see, thank you for explaining! I didn't know/never realized that you could give build args to `go vet`. It would be interesting to see where and how those get applied. I'll take a look later today. I think the results of `buildssa` should be the same no matter the build args, since it looks like `buildssa` ingests file `AST`s.

> I'm bet optimization would break a lot of our tests.

Yep. `rg "if true" | wc -l` yields 25 matches.

I think your `flipCoin()` suggestion would make the tests more readable. It is reasonable to expect that `if true` would lead to some optimization, and indeed we have both had that expectation at some point.

I agree that for this PR, this is not a big concern, but I'll keep it in mind for a future refactoring.

PR #156, opened 2020-10-19
Add tests for named return values

mlevesquedion:
This PR adds tests for named return values.

In this case:
```go
func TestNamedReturnValue() (s core.Source) {
	return
}
```
`s` will get an `Alloc`, as though it had been declared in the body. In the function's `ssa`, `s` is listed as a local variable.

If we change `core.Source` to `*core.Source`, we no longer expect a report. Indeed, the code becomes roughly equivalent to:
```go
func Test() {
	var s *core.Source
	_ = s
}
```
In this case, `s` does not get an `Alloc` (because nothing needs to be allocated), and in fact `s` is not even listed as a local variable (because it is simply nil).

If a value were assigned to the pointer return value, then it would be detected by other means.

- [x] Tests pass

### DIFF
--- a/internal/pkg/levee/testdata/src/example.com/tests/namedreturn/tests.go
+++ b/internal/pkg/levee/testdata/src/example.com/tests/namedreturn/tests.go
@@ -24,7 +24,7 @@ func TestNamedReturnValue() (s core.Source) {
 }
 
 func TestNamedReturnValuePointer() (s *core.Source) {
-	core.Sink(s) // we do not want a report here, because no Source value has been created
+	core.Sink(s) // we do not report here, because no Source value has been created, so s must be nil
 	return
 }
 
--- a/internal/pkg/levee/testdata/src/example.com/tests/namedreturn/tests.go
+++ b/internal/pkg/levee/testdata/src/example.com/tests/namedreturn/tests.go
@@ -27,3 +27,9 @@ func TestNamedReturnValuePointer() (s *core.Source) {
 	core.Sink(s) // we do not want a report here, because no Source value has been created
 	return
 }
+
+func TestNamedReturnValuePointerWithInitialization() (s *core.Source) {
+	s = &core.Source{}
+	core.Sink(s) // want "a source has reached a sink"
+	return
+}
--- a/internal/pkg/levee/testdata/src/example.com/tests/namedreturn/tests.go
+++ b/internal/pkg/levee/testdata/src/example.com/tests/namedreturn/tests.go
@@ -1,0 +1,29 @@
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package namedreturn
+
+import (
+	"example.com/core"
+)
+
+func TestNamedReturnValue() (s core.Source) {
+	core.Sink(s) // want "a source has reached a sink"
+	return
+}
+
+func TestNamedReturnValuePointer() (s *core.Source) {
+	core.Sink(s) // we do not want a report here, because no Source value is being created
+	return
+}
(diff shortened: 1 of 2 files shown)


internal/pkg/levee/testdata/src/example.com/tests/namedreturn/tests.go
```diff
@@ -0,0 +1,29 @@
...
+	"example.com/core"
+)
+
+func TestNamedReturnValue() (s core.Source) {
+	core.Sink(s) // want "a source has reached a sink"
+	return
+}
+
+func TestNamedReturnValuePointer() (s *core.Source) {
+	core.Sink(s) // we do not want a report here, because no Source value has been created
```
PurelyApplied, line 27: I agree that we're not expecting a report under current implementation, but I'd consider this a bug in our reliance on `Alloc` rather than a statement of intent.  It's unclear from this comment and this PR's description whether or not you consider this working as intended.

I'd be much happier with a `// TODO want "a source has reached a sink"`, and optionally an explanation as to why we currently don't.

mlevesquedion: I see. Let me try to clarify, then.

Given just this code:
```go
func TestNamedReturnValuePointer() (s *core.Source) {
	core.Sink(s) // we do not want a report here, because no Source value has been created
}
```
There is nothing wrong here, so we don't want a report. Indeed, `s` is `nil`, so it can't be tainted. For `s` to be non-nil, it will have to be assigned a value, i.e. something obtained via a parameter or created locally, in which case we will catch it through those other means. The same logic applies in the following example:
```go
func Test() {
	var s *core.Source
	core.Sink(s) // we don't want a report here either
}
```
We don't want a report here either, because `s` is `nil`. It has never held and will not hold a reference to an actual `core.Source` value.

Does that make sense?

PurelyApplied: I'm sorry.  I really have to stop doing reviews first thin in the morning, before I'm properly caffeinated.  I read through the nilary arguments and was reading `Foo(s *core.Source)`.  I was surprised that you seemed unconcerned that an argument passed by referenced seemed to not be issuing a report.

Now that I'm reading what you actually wrote, yes, I see that you're saying we report on a `var s *Source = nil`.

I think this case might be one that I'm okay with the false positive, at least for now.  So similar to the comment on your other test PR: I'd open an issue about reporting on explicitly nil arguments of source reference type, and add a `// TODO We don't actually want` before the `// want "foo"` that would suppress this test.

Related, would this also report on, say, an empty or `nil` instance of `[]core.Source` or `map[string]*core.Source`?

mlevesquedion: > Now that I'm reading what you actually wrote, yes, I see that you're saying we report on a var s *Source = nil.

We actually *don't* produce a report there, which I think is correct behavior.

Same for `[]core.Source` and `map[string]*core.Source`.

mlevesquedion: My phrasing may have led you onto the wrong path. When I said "we do not want a report here", I didn't mean "we are currently producing a report here, but we shouldn't be".

PurelyApplied: I've now become absolutely and entirely confused...

If we are producing a report here, but shouldn't be, why is Jenkins testing passing?  There's not `// want` there to satisfy the report.

mlevesquedion:
I added a test that shows a case where we *would* want a report.

I investigated whether a `*Source` could be initialized by a function, e.g.
```go
func InitializeSource(s *Source) {
    s.Data = os.Getenv("secret")
}
```
This can't happen unless the `*Source` is assigned in some way, either via a parameter or a local variable, in which case we will catch it. Otherwise, the argument passed to `InitializeSource` will be `nil`, so it won't be able to do anything with it.

Adding a level of indirection, e.g. `**Source` does not change the reasoning in the previous paragraph: the function will receive `nil`, so it won't be able to do anything.

internal/pkg/levee/testdata/src/example.com/tests/namedreturn/tests.go
PurelyApplied: To close a long thread of me confusing myself, I'd prefer this comment not use the word `want` due to its strong association elsewhere in our testing.  `// we do not report here, because ...`

PurelyApplied (approved):
Mentioned at the tail of the long thread, but my easily-confused self might prefer a rewording on that comment to avoid overloading "want."

internal/pkg/levee/testdata/src/example.com/tests/namedreturn/tests.go
mlevesquedion: My bad, forgot to make the update after we discussed it. Thanks for the reminder! Done.

Issue #103, opened 2020-09-14
Allow customization of reporting messages


mlevesquedion:
It would likely be useful for users to be able to customize all or part of the report message. Currently, the report message is generic: `<sink position>: a source has reached a sink: <source position>`. This is an adequate message in general, but users may wish to add something to it to make it more actionable, e.g. by specifying what a developer who sees the message should do about it: (this is just an example) `The sensitive value at <source position> must be sanitized before reaching the call at <sink position>`.

Issue #91, opened 2020-09-10
Add configuration validation / sanity-checks


PurelyApplied:
When #87, #88, #89 are implemented, a user may be able to configure, e.g., a sink belonging to a package that is listed outside the configured scope.  Some level of validation should safeguard against user error.

mlevesquedion:
Some of these checks would likely be easy to implement and I think they could help users tremendously, since proper configuration is necessary for our analysis to produce useful results.

We might also want to provide a way for users to check that their configuration targets what they are expecting by allowing them to run the tool in a mode where it simply reports `Sources` and `Sinks` it finds, without doing any propagation.

Issue #88, opened 2020-09-10
Enable specification of analysis scope


PurelyApplied:
Currently, analysis is executed on all provided packages, as well as the transitive closure of all Go package dependencies.

A user may consider third-party packages, generated code to be out-of-scope, or specifically sanctioned packages out of scope for analysis.  While upstream analysis on these may be necessary (e.g. `buildssa`), we should bypass our analysis of such specified packages as much as possible.

Ideally, this bypass should occur prior to entry into the various `analyzer.Analyzer`s to reduce the size of the transitive dependency graph.

mlevesquedion:
For performance reasons I agree that we will likely want to avoid analyzing anything we don't have to, but if for whatever reason we *have* to analyze everything, we could just avoid _producing reports_ for violations that are out of scope.

mlevesquedion:
I think with the `Exclude` field we now have adequate support for this feature.

Issue #87, opened 2020-09-10
Provide a mechanism by which false positives may be suppressed.


PurelyApplied:
If analysis is integrated to become a blocking test, users will be dismayed if a false positive blocks progress.  It should be possible to bypass these errors.

This could be done via an explicit allow-list (inclusive-)or in-code comment in the `// NOLINT` style.  The former should perhaps be favored for the possibility of out-of-scope findings, e.g. a finding in a user's third-party dependencies or in generated code.

mlevesquedion:
I'm having a hard time seeing exactly how users would define an allowlist. Would users write source code positions in a file? It sounds like that would probably lead to a fair amount of toil for users. Did you have a different idea in mind?

To me code comments make the most sense because I think they are simpler for users to introduce and to maintain. I think out-of-scope findings will be avoided via #88. I'm not sure how false positives in generated code would be handled though. Perhaps we could provide users with a way to exclude files via some naming pattern.

mlevesquedion:
Whatever mechanism we choose, I'm concerned that allowing users to suppress false positives may lead to some issues going undetected, e.g. if `levee` reports a false positive on a given line of code, the user suppresses it through e.g. a comment, and later on the line of code is changed and becomes a true positive, but because of the suppression `levee` does not report it.

mlevesquedion:
A built-in way for users to suppress false positives would be to use a sanitizer. We may choose to advertise this or not. Some pros and cons to using sanitizers to suppress false positives:

Pros:
* Requires no extra work from us, since this is already supported.
* Simple for users, since they don't need to learn a new concept/mechanism.

Cons:
* Suppressing false positives is not what sanitizers are intended for. Users may create a "do nothing" sanitizer that does not actually do any sanitization and is intended merely to suppress false positives, which could lead to true positives being missed if it is used improperly. (We cannot prevent users from doing this, but we could warn against doing this in the documentation).
* We don't currently have a way to report that using a sanitizer in a given piece of code was not necessary, and I think it is unlikely that we will develop such a feature. Over time, a code base could accumulate unnecessary sanitizing calls.

mlevesquedion:
The `staticcheck` tool uses code comments to suppress false positives. Details [here](https://staticcheck.io/docs/#ignoring-problems). Summary:
* False positives on a line can be suppressed via a code comment on the preceding line. Unnecessary suppressions are reported.
* False positives in an entire file can similarly be suppressed via a code comment.

[...]

PR #157, opened 2020-10-20
Rework configuration to allow YAML and future configuration feature-add.

PurelyApplied:
This refactor is intended to enable future resolution of #87, #88, #89, #91, #103, i.e. our `label:configuration` issues.

- [x] Tests pass

===

### Where should the heavy lifting occur?

Pros and cons of iterating towards `package config` being a simple data container, with actual matching logic being owned by the relevant packages.

* Pro:
  * Keep `ssa` logic where `ssa` is being used, rather than `config` requiring global implementation knowledge
  * Keep `config` "low level," avoiding need to expose excessive number of methods as we add additional configuration features.
* Con:
  * Each package will need to "unpack" configuration, e.g. `source` will need to `for i, s := range cfg.Source { s.Package.Match(mySsa.pkg) // ... }` etc.

I could go either way on it, personally.  A lot of `ssa` logic currently lives in `config`.

I think maybe the ideal option would be to split the difference and to rewrite the current method-set to only take basic types  E.g.
`func (c Config) IsSourceField(packagePath, typeName, fieldName) bool`.  Or maybe to return the set of matchers that satisfy the given constraint.  I'm not sure, and I think this is something that's only going to evince itself as we try to migrate to the new struct.

### What do you think of the example YAML?

@mlevesquedion and I have had some discussion offline about what a decent implementation might look like.  Now having iterated on it, I'd appreciate your and/or other's input on this spec style.  Some key questions:

* Do you want to maintain the name `packageRE`, or have `package` implicitly be a regexp?  Or accept either, using `packageRE` as a regexp patch and `package` as an explicit string match?
* Does the `fieldtags` format of explicitly listing key-value pairs read well?  We could accept a list of tag strings, e.g., `foo:"bar"` as the input, but I worry how that would read / parse in YAML.

### DIFF
--- a/internal/pkg/config/specifiers.go
+++ b/internal/pkg/config/specifiers.go
@@ -1,16 +1,56 @@
 package config
 
 import (
-	"fmt"
-
 	"github.com/google/go-flow-levee/internal/pkg/config/regexp"
 )
 
 // This type marks intended future work
 type NotImplemented = interface{}
 
+// ConfigV2 is a more generic config.
+type ConfigV2 struct {
+	ApiVersion string `yaml:"apiVersion"`
+	Kind       string
+
+	Metadata metaSpec
+
+	Source    []specifier
+	Sink      []specifier
+	Sanitizer []specifier
+	Allowlist []specifier
+}
+
+type metaSpec struct {
+	ValidateConfig NotImplemented `yaml:"validateConfig"`
+	Scope          NotImplemented
+}
+
+type valueSpec struct {
+	Id            string           // Identifier for this spec
+	TypeSpec      `yaml:",inline"` // Match source type by path and type name
+	FieldSpec     `yaml:",inline"` // Match field of above type by name or tags
+	Scope         NotImplemented   // Match by local/param/global
+	IsReference   NotImplemented   // Match by pointer/slice/map
+	MatchConstant NotImplemented   // Match explicit value, e.g. "PASSWORD"
+	Context       NotImplemented   // Match invocation context
+	Unless        []valueSpec      // Exclusion matchers
+}
+
+type callSpec struct {
+	Id        string           // Identifier for this spec
+	TypeSpec  `yaml:",inline"` // Match by package and optional receiver
+	Function  string           // Match function/method by name
+	Arguments NotImplemented   // Match function invocation by arguments
+	Context   NotImplemented   // Match invocation context
+	Unless    []callSpec       // Exclusion matchers
+}
+
+type specifier struct {
+	Value *valueSpec `yaml:",omitempty"`
+	Call  *callSpec  `yaml:",omitempty"`
+}
+
 type FieldSpec struct {
-	typeSpec  `yaml:",inline"`
 	Field     regexp.Regexp
 	Fieldtags []fieldTagMatcher
 }
@@ -28,60 +68,11 @@ func (fs FieldSpec) MatchTag(key, value string) bool {
 	return false
 }
 
-type typeSpec struct {
+type TypeSpec struct {
 	Package regexp.Regexp
 	Type    regexp.Regexp
 }
 
-func (ts typeSpec) Match(pkg, typ string) bool {
+func (ts TypeSpec) Match(pkg, typ string) bool {
 	return ts.Package.MatchString(pkg) && ts.Type.MatchString(typ)
-}
-
-type valueSpec struct {
-	FieldSpec     `yaml:",inline"` // Match according to field name or tags
-	Id            string
-	Unless        []valueSpec
-	Scope         NotImplemented
-	IsReference   NotImplemented
-	MatchConstant NotImplemented
-}
-
-type callSpec struct {
-	typeSpec     // patch package and optional receiver
-	Id           string
-	FunctionName string
-	Arguments    NotImplemented
-}
-
-type metaSpec struct {
-	ValidateConfig NotImplemented `yaml:"validate-config"`
-	Scope          NotImplemented
-}
-
-type Specifier struct {
-	Value *valueSpec `yaml:"value,omitempty"`
-	Call  *callSpec  `yaml:"call,omitempty"`
-}
-
-func (s Specifier) validate() error {
-	switch {
-	case s.Value != nil && s.Call != nil:
-		return fmt.Errorf("a specifier should include only a value or call specification")
-	case s.Value == nil && s.Call == nil:
-		return fmt.Errorf("specifier includes neither a value or call specification")
-	}
-	return nil
-}
-
-// ConfigV2 is a more generic config
-type ConfigV2 struct {
-	Apiversion string
-	Kind       string
-
-	Metadata metaSpec
-
-	Source    []Specifier
-	Sink      []Specifier
-	Sanitizer []Specifier
-	Allowlist []Specifier
 }
--- a/internal/pkg/config/specifiers.go
+++ b/internal/pkg/config/specifiers.go
@@ -1,0 +1,87 @@
+package config
+
+import (
+	"fmt"
+
+	"github.com/google/go-flow-levee/internal/pkg/config/regexp"
+)
+
+// This type marks intended future work
+type NotImplemented = interface{}
+
+type FieldSpec struct {
+	typeSpec `yaml:",inline"`
+	Field    regexp.Regexp
+	Tags     []fieldTagMatcher
+}
+
+func (fs FieldSpec) MatchName(name string) bool {
+	return fs.Field.MatchString(name)
+}
+
+func (fs FieldSpec) MatchTag(key, value string) bool {
+	for _, ftm := range fs.Tags {
+		if ftm.Key == key && ftm.Val == value {
+			return true
+		}
+	}
+	return false
+}
+
+type typeSpec struct {
+	Package regexp.Regexp
+	Type    regexp.Regexp
+}
+
+func (ts typeSpec) Match(pkg, typ string) bool {
+	return ts.Package.MatchString(pkg) && ts.Type.MatchString(typ)
+}
+
+type valueSpec struct {
+	FieldSpec     `yaml:",inline"` // Match according to field name or tags
+	Id            string
+	Unless        []valueSpec
+	Scope         NotImplemented
+	IsReference   NotImplemented
+	MatchConstant NotImplemented
+}
+
+type callSpec struct {
+	typeSpec     // patch package and optional receiver
+	Id           string
+	FunctionName string
+	Arguments    NotImplemented
+}
+
+type metaSpec struct {
+	ValidateConfig NotImplemented `yaml:"validate-config"`
+	Scope          NotImplemented
+}
+
+type Specifier struct {
+	Value *valueSpec `yaml:"value,omitempty"`
+	Call  *callSpec  `yaml:"call,omitempty"`
+}
+
+func (s Specifier) validate() error {
+	switch {
+	case s.Value != nil && s.Call != nil:
+		return fmt.Errorf("a specifier should include only a value or call specification")
+	case s.Value == nil && s.Call == nil:
+		return fmt.Errorf("specifier includes neither a value or call specification")
+	}
+	return nil
+}
+
+// ConfigV2 is a more generic config
+type ConfigV2 struct {
+	Apiversion string
+	Kind       string
+
+	Metadata metaSpec
+
+	Source    []Specifier
+	Sink      []Specifier
+	Sanitizer []Specifier
+	Allowlist []Specifier
+}
--- a/internal/pkg/config/specifiers.go
+++ b/internal/pkg/config/specifiers.go
@@ -1,3 +1,17 @@
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
 package config
 
 import (
--- a/internal/pkg/config/specifiers.go
+++ b/internal/pkg/config/specifiers.go
@@ -28,10 +28,10 @@ type ConfigV2 struct {
 
 	Metadata metaSpec
 
-	Source    []specifier
-	Sink      []specifier
-	Sanitizer []specifier
-	Allowlist []specifier
+	Sources    []specifier
+	Sinks      []specifier
+	Sanitizers []specifier
+	Allowlist  []specifier
 }
 
 type metaSpec struct {
--- a/internal/pkg/config/specifiers.go
+++ b/internal/pkg/config/specifiers.go
@@ -10,9 +10,9 @@ import (
 type NotImplemented = interface{}
 
 type FieldSpec struct {
-	typeSpec `yaml:",inline"`
-	Field    regexp.Regexp
-	Tags     []fieldTagMatcher
+	typeSpec  `yaml:",inline"`
+	Field     regexp.Regexp
+	Fieldtags []fieldTagMatcher
 }
 
 func (fs FieldSpec) MatchName(name string) bool {
@@ -20,7 +20,7 @@ func (fs FieldSpec) MatchName(name string) bool {
 }
 
 func (fs FieldSpec) MatchTag(key, value string) bool {
-	for _, ftm := range fs.Tags {
+	for _, ftm := range fs.Fieldtags {
 		if ftm.Key == key && ftm.Val == value {
 			return true
 		}
--- a/internal/pkg/config/testdata/src/configv2/discussion/source_examples.go
+++ b/internal/pkg/config/testdata/src/configv2/discussion/source_examples.go
@@ -1,0 +1,57 @@
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// package discussion demonstrates some examples of configuration intended by the new config style
+package discussion
+
+import "os"
+
+// All Tokens are sources, even though the underlying type is a basic kind.
+type Token string
+
+// Since configuration does not specify any field as a source, then any TokenBundle is identified as a source.
+// (Implicitly, every field is a source field.)
+type TokenBundle struct {
+	bundle                     []string
+	masterToken                string
+	internalCommunicationToken string
+}
+
+// Configuration specifically identifies a field as a source in this type.
+// When a type has a source field, direct access of the non-source fields should not propagate taint.
+type NamedToken struct {
+	token string
+	name  string
+}
+
+// Configuration specifically identifies a field as a source in this type via field tag.
+// As above, direct access of non-source fields should not propagate taint.
+type TaggedToken struct {
+	Token string `myTag:"source"`
+	name  string
+}
+
+func Foo() {
+	// The string in this environment variable is sensitive.
+	password := os.Getenv("CFSSL_CA_PK_PASSWORD")
+
+	// The string in this environment variable is not.
+	userDir := os.Getenv("USER_DIR")
+
+	_, _ = password, userDir
+}
+
+func SaveSecret(secret interface{}) {
+	// The parameter secret, though received by interface{}, is still identified as a source.
+}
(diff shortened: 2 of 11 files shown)


mlevesquedion:
> Where should the heavy lifting occur?

I don't have a strong preference here either, but I agree that the config should not know about the `ssa` package. I think having methods that take basic types would probably work. We certainly don't want other packages asking the config for a list of e.g. sources and doing the work themselves. Mostly though, I'm drawn to the same conclusion as you: I think this will become clearer as the implementation progresses.

> Do you want to maintain the name packageRE, or have package implicitly be a regexp? Or accept either, using packageRE as a regexp patch and package as an explicit string match?

IMO if we use regexps, having RE in the name is important. Even *with* RE in the name right now though, I often forget that they are regexp's. I don't think there's a big benefit to allowing RE's, but they *can* lead to unexpected behavior, so **I think we should use explicit strings.** What was the original reasoning in adopting regexps? One advantage I see is if that if you have a vague idea of what your e.g. source types are, you can say "A source is something that has 'token' or 'secret' somewhere in it." ... but I feel like taking a guess here is unlikely to give good results. Rather, it is likely to lead to false confidence and/or false positives.

> Does the fieldtags format of explicitly listing key-value pairs read well? We could accept a list of tag strings, e.g., foo:"bar" as the input, but I worry how that would read / parse in YAML.

I think it reads quite well. I think something like `- mytag:"source"` would be more confusing.

internal/pkg/config/testdata/test-config-v2.yaml
```diff
@@ -0,0 +1,58 @@
...
+sink:
+- call:  &println-sink
+    id: "don't print"
+    package: fmt
+    function: Println
+    unless:
+      - context:
+          package: my/sanctioned/package
+- value:
+    id: "Sink function member"
```
mlevesquedion, line 41: I'm not sure using a value specifier here is entirely intuitive. I think a dedicated `funcField` matcher would be easier to understand.

internal/pkg/config/specifiers.go
```diff
@@ -0,0 +1,78 @@
...
+	"github.com/google/go-flow-levee/internal/pkg/config/regexp"
+)
+
+// This type marks intended future work
+type NotImplemented = interface{}
+
+// ConfigV2 is a more generic config.
+type ConfigV2 struct {
+	ApiVersion string `yaml:"apiVersion"`
+	Kind       string
```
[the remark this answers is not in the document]
PurelyApplied: I think both the `ApiVersion` and `Kind` headers were just a knee-jerk reaction to them being part of the various Kubernetes configs.  But those conventions don't apply here.

I'll drop the `Kind`.  I think even the `ApiVersion` could be pruned until if/when we get to a point of supporting more than one version of config.  But that feels like a post-1.0 sort of world that we're not a part of yet.

internal/pkg/config/specifiers.go
```diff
@@ -0,0 +1,78 @@
...
+	IsReference   NotImplemented   // Match by pointer/slice/map
+	MatchConstant NotImplemented   // Match explicit value, e.g. "PASSWORD"
+	Context       NotImplemented   // Match invocation context
+	Unless        []valueSpec      // Exclusion matchers
+}
+
+type callSpec struct {
+	Id        string           // Identifier for this spec
+	TypeSpec  `yaml:",inline"` // Match by package and optional receiver
+	Function  string           // Match function/method by name
```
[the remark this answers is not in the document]
PurelyApplied: Ha!  That's what I had been using, but I shortened it to be consistent with the current JSON config.  Happy to change it back, though.

internal/pkg/config/testdata/test-config-v2.yaml
```diff
@@ -0,0 +1,58 @@
...
+    fieldtags:
+      - myTag: source
+      - myTag: otherSource
+- call:
+    id: "sensitive environment information"
+    package: os
+    function: getEnv
+    arguments:
+    - value:
+        position: 0
```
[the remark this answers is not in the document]
PurelyApplied: This is more what I had in mind with something like `validateConfig` flag above - not a data-validation, but more of a sanity-check sort of thing.  The naming, implementation details, etc would be a future concern; I was throwing it out there as a "for instance" in this draft.  (I should've been more clear here with liberal comments in the Yaml.)

internal/pkg/config/testdata/test-config-v2.yaml
```diff
@@ -0,0 +1,58 @@
+apiVersion: config/v2
+kind: config
+
+metadata:
+  validateConfig: true
```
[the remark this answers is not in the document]
PurelyApplied: See comment below.

internal/pkg/config/specifiers.go
```diff
@@ -0,0 +1,78 @@
...
+	Scope          NotImplemented
+}
+
+type valueSpec struct {
+	Id            string           // Identifier for this spec
+	TypeSpec      `yaml:",inline"` // Match source type by path and type name
+	FieldSpec     `yaml:",inline"` // Match field of above type by name or tags
+	Scope         NotImplemented   // Match by local/param/global
+	IsReference   NotImplemented   // Match by pointer/slice/map
+	MatchConstant NotImplemented   // Match explicit value, e.g. "PASSWORD"
```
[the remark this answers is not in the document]
PurelyApplied: Could.  I suspect that would massage some implementation details.  For instance, `ssa.Member`s doesn't implement `ssa.Node`, so the interface that will be required to check a source by type is likely going to be different than that used to check an `ssa.Value`.

internal/pkg/config/specifiers.go
```diff
@@ -0,0 +1,78 @@
...
+	ValidateConfig NotImplemented `yaml:"validateConfig"`
+	Scope          NotImplemented
+}
+
+type valueSpec struct {
+	Id            string           // Identifier for this spec
+	TypeSpec      `yaml:",inline"` // Match source type by path and type name
+	FieldSpec     `yaml:",inline"` // Match field of above type by name or tags
+	Scope         NotImplemented   // Match by local/param/global
+	IsReference   NotImplemented   // Match by pointer/slice/map
```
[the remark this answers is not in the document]
PurelyApplied: I was mostly just pulling from #89 and the associated discussion.  The only use I see at present would be as a condition on a call argument.  I might only be concerned if a function is called with a mutable argument, for instance.

I'm not sure `map[string]string` or some other built-in type would parse from a YAML string well, although I suppose syntax we'd need to parse is pretty limited - `map[foo]bar`, `[]slice`, `[k]array`, `chan baz`.  I'd worry about something more complex, though, e.g. `map[foo]map[bar]baz`.

internal/pkg/config/specifiers.go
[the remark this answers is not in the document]
PurelyApplied: Counter-point, I'm trying to keep the config a bit implementation-agnostic and accessible.  If we split out `type`, `field`, `value`, and `call`, it begins to feel a bit precious.

```yaml
source:
- type:   # identify all fizz.buzz as sources
    package: fizz
    struct: buzz
- field:   # Type foo.bar contains field baz, which is a source
    package:  foo
    struct: bar
    fieldName: baz
- field:    # Any type containing a specific tag should be marked a source
    fieldTag:
      - qux: thud
- call:  # The string returned by os.getEnv("CFSSL_CA_PK_PASSWORD") is sensitive
    package: os
    function: getEnv
    arguments:
    - value:
        position: 0
        const: CFSSL_CA_PK_PASSWORD
```

That feels a bit redundant between the `field` and `type` spec.

Their inclusion in this draft implementation was just an attempt to extract shared components, not necessarily to expose that as part of the configuration api.

internal/pkg/config/testdata/test-config-v2.yaml
PurelyApplied: I don't know.  On the one hand, yes, functions feel separate from values.  But on the other hand, we're already assuming a good bit of familiarity with Go from anyone building a config.  That functions are first-class objects is part of that familiarity.

I feel like this ends up being a documentation issue, where we should be explicit that any struct field would need to be identified as a `value`.

I'll come up with some more concrete examples for discussion, and see how it feels then.

internal/pkg/config/testdata/test-config-v2.yaml
```diff
@@ -0,0 +1,58 @@
...
+  scope:
+    skip:
+    - imports: testing
+    - package: my/special/pkg
+
+source:
```
vinayakankugoyal, line 11: should this be called sources?

internal/pkg/config/testdata/test-config-v2.yaml
```diff
@@ -0,0 +1,58 @@
...
+- call:
+    id: "sensitive environment information"
+    package: os
+    function: getEnv
+    arguments:
+    - value:
+        position: 0
+        const: CFSSL_CA_PK_PASSWORD
+
+sink:
```
vinayakankugoyal, line 32: should this be called sinks?

internal/pkg/config/specifiers.go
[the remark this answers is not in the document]
mlevesquedion: SG. I wasn't really thinking about the implementation TBH.

To me, being able to know up front the *kind* of value that is being defined makes it easier to understand *what* is being defined. Reading the example in your comment, it is easy for me to say "this is a *type*", "this is a *field*", and then when I read the `key: value` pairs I can fill in the details of what specifically is being defined. Whereas if I only have `value`, I have to look at the `key: value` pairs to deduce what kind of thing is being defined. In the case of `field` vs `type`, until I get to `fieldName`, I have no idea what kind of thing is being defined.

For structs we may want to have a list of sensitive field names, in case there are many sensitive fields.

I feel like this discussion is getting beyond the current scope of the PR, we can defer if you want.

PurelyApplied:
> What was the original reasoning in adopting regexps? 

To my memory, it was mostly out of convenience.  We were working with a log library at the time with distinct `log.Info`, `log.Warning`, `log.Error`, `log.Fatal`, `log.Exit` levels, each with the `Info`, `Infof`, `Infoln`, `InfoDepth`, and `InfoDepthf` variants.  It was much cleaner to do a `Info|Warning|Error|Fatal|Exit` matcher over explicitly enumerating thirty different possible sinks.

mlevesquedion:
That's fair. In that case we could support alternation without supporting the full set of regexp capabilities.

PurelyApplied:
> That's fair. In that case we could support alternation without supporting the full set of regexp capabilities.

I'm not sure how you mean that.  It would be pretty easy to allow both, e.g.

```yaml
- id: "exact"
  package: my/explicit/package/name
  type: ExactTypeName
- id: "relaxed"
  packageRegexp: "(my|your|daves)/package"
  typeRegexp: "Secret$"
- id: "mix-and-match"
  package: "this/exact/package"
  typeRegexp: "^(?:Other)?Secret$"
```
and to error marshalling on any that included both an explicit and a regexp matcher.

mlevesquedion:
I think allowing both might be the best way to go.

I was viewing this through the "what if we didn't support regexps" assumption.

PurelyApplied:
I'll work dual support in then in a future commit.

There's a change between the newest example config and the first one.  We could either go with:

```yaml
sources:
  - some value spec
  - another value spec
  - some type spec
  - another type spec
  - some call spec
```

versus

```yaml
sources:
  type:
    - some type spec
    - another type spec
  value:
    - some value spec
    - another value spec
  call:
    - some call spec
```

mlevesquedion:
I prefer the flat version (no subcategories), but I don't have a strong preference. I think it'll be slightly easier for users to manage a list that doesn't have subcategories.

internal/pkg/config/testdata/src/configv2/discussion/test-config-v2.yaml
```diff
@@ -0,0 +1,66 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+source:
+  value:
+    - id: "Tokens are secret"
```
mlevesquedion, line 17: I think `description` may be a better key name. It seems to be how you're using it, and I suspect that's how users will use it as well.

internal/pkg/config/testdata/src/configv2/discussion/test-config-v2.yaml
```diff
@@ -0,0 +1,66 @@
...
+      type: TokenBundle
+    - id: "NamedToken.token is secret, NamedToken.name is not"
+      package: configv2/discussion
+      type: NamedToken
+      fieldName: token
+    - id: "Tagged fields are secret, neighboring untagged fields are not"
+      # empty package and type fields match anything
+      fieldTag:
+      - myTag: source
+    - id: "In context, the parameter of SaveSecret(secret interface{}) is a source"
```
mlevesquedion, line 31: I'm a little worried that such cases could be very frequent in some code bases and that they would lead to an explosion in the size of the config, which could make it hard for users to manage.

It seems to me that this is something we should handle via IPA (**I**nter**P**rocedural **A**nalysis).

internal/pkg/config/testdata/src/configv2/discussion/test-config-v2.yaml
```diff
@@ -0,0 +1,66 @@
...
+      # this could use additional details and perhaps a better name than "scope"
+      scope: parameter
```
mlevesquedion, line 33: What if there's more than one parameter? What if it's a local variable? What if it's a global variable?

internal/pkg/config/testdata/src/configv2/discussion/test-config-v2.yaml
```diff
@@ -0,0 +1,66 @@
...
+      context:
+        package: configv2/discussion
+        functionName: SaveSecret
+  call:
+    - id: "sensitive environment information"
+      package: os
+      function: getEnv
```
mlevesquedion, line 40: Nit: `Getenv`

internal/pkg/config/testdata/src/configv2/discussion/test-config-v2.yaml
```diff
@@ -0,0 +1,66 @@
...
+      arguments:
+      - position: 0
+        const: CFSSL_CA_PK_PASSWORD
```
mlevesquedion, line 43: This feels rather brittle. What if the position changes? (Not applicable in this case, but let's assume the function has more than 1 parameter.) What if it's a variable instead of a constant (same value)?

mlevesquedion: (In both of those cases, hopefully the analysis validation would produce a report saying that this no longer matches.)

But I'm legitimately worried about the case where an argument is a variable. What are the other possibilities for what "arguments" could be?

internal/pkg/config/testdata/src/configv2/discussion/source_examples.go
```diff
@@ -0,0 +1,57 @@
...
+// package discussion demonstrates some examples of configuration intended by the new config style
+package discussion
+
+import "os"
+
+// All Tokens are sources, even though the underlying type is a basic kind.
+type Token string
+
+// Since configuration does not specify any field as a source, then any TokenBundle is identified as a source.
+// (Implicitly, every field is a source field.)
```
mlevesquedion, line 24: I think this makes sense, but it will need to be clearly documented. As will everything else, but I feel like this could be a gotcha. Most source types contain only a few sensitive fields.

It may be helpful to include a flag that prints what the config will match.

internal/pkg/config/testdata/src/configv2/discussion/test-config-v2.yaml
PurelyApplied: Not to discount the value of interprocedural work, but this would cover the case of, e.g., a `[]byte` stream that is expected to contain a secret.  Interprocedural would cover the case that this byte stream were either `type SecretStream []byte` or produced out of scope by some `var secret interface{} = os.Getenv("PASSWORD") ; SaveSecret(secret)`.

I agree that, by itself, it is insufficient to cover our concern about interface obfuscation.

internal/pkg/config/testdata/src/configv2/discussion/source_examples.go
PurelyApplied: > It may be helpful to include a flag that prints what the config will match.

This was the intent in a design of subordinate analyzers.  The `sourceType` analyzer emits exactly what source type and fields are matched by config.  The `source` analyzer emits what `*ssa.Value`s are identified as sources, currently based on type and in future possibly based on identified function return values (e.g. `os.Getenv`).

For complete information, we could produce similar analyzers for `sanitizer` and `sink` function declarations and invocations.  I have long held that my ideal world would be one where `levee` performs taint propagation, with all `ssa.Node`s having already been painted with the relevant information.  In that stance, a `--print-identified=source-types` option would just start `levee` analysis by printing the input it has received.

internal/pkg/config/testdata/src/configv2/discussion/test-config-v2.yaml
PurelyApplied: > This feels rather brittle. What if the position changes? (Not applicable in this case, but let's assume the function has more than 1 parameter.) What if it's a variable instead of a constant (same value)?

I don't disagree on any particular point.  Again, I've mostly been trying to reproduce #89 and the associated discussion.  I agree it's brittle to refactoring and anything other that constant arguments.  But this particular case is [TOB-K8S-005](https://github.com/kubernetes/kubernetes/issues/81117), and I would love to defend against such things.

> But I'm legitimately worried about the case where an argument is a variable. What are the other possibilities for what "arguments" could be?

Here's is a gnarly edge-case I've run into, though.  Supposing my sinks are log calls, if I have a sanctified log wrapper `SafeLogf` that cleans all its arguments, I wouldn't want to report any issue if I pass sources to that.  But I might want to report an issue if the format string itself is tainted.

```golang
func Foo(s Source) {
  tainted := fmt.Sprintf("something something %v blah blah", s.Data)
  SafeLogf(tainted + " and something else: %v", s)  // The s will be sanitized, but the format string has the source data hardcoded into it.
}
```

I think the most compelling \*ahem\* argument for argument-level specification is being able to specify which arguments in a sink get sunk, which arguments in a sanitizer get sanitized, etc.  The above example, for instance, could be configured as a sink that only sinks the first argument.  I only have a vague idea about how I might approach that as an implementation, and I agree that maybe this initial draft doesn't cover well how that configuration might look.  But I was trying to leave room to be extensible in that direction.

A lot of these `NotImplemented` portions fit that bill.  It's granularity that isn't an immediate goal, but I'm trying to accommodate extensibility in that direction.  Since they're essentially dead code, I would be happy to prune much of the `NotImplemented` portions before this PR (or another superseding it) were to be merged.  This PR was mostly about fleshing out these discussions.

internal/pkg/config/testdata/src/configv2/discussion/test-config-v2.yaml
mlevesquedion: I think in short, to me it seems more robust to match calls than to match functions, because a call matcher can potentially be used in many places. It's also easier for users to just say "this function produces a source" than to look for the potentially many places where a value tainted from the result of a source-producing function is used.

In your example, I'm thinking that `[]byte` must have come from somewhere, and in general I think it would be preferable to look for the location where it is produced.

Being able to match a specific parameter of a specific function may be useful if other options don't work, though.

internal/pkg/config/testdata/src/configv2/discussion/test-config-v2.yaml
mlevesquedion: > I don't disagree on any particular point. Again, I've mostly been trying to reproduce #89 and the associated discussion. I agree it's brittle to refactoring and anything other that constant arguments. But this particular case is TOB-K8S-005, and I would love to defend against such things.

Agreed. (Weirdly, I don't recall ever seeing the string "TOB-K8S-005" before. To me it's just been "finding 9 in the k8s security audit". I'll remember it going forward, though.)

> I think the most compelling *ahem* argument for argument-level specification is being able to specify which arguments in a sink get sunk, which arguments in a sanitizer get sanitized, etc.

I agree, that does sound valuable.

I think it would be straightforward to implement since in the `ssa` it's easy to find which argument is in which position in a `Call`.

Going back to the TOB-K8S-005 example, I think one thing that could give us more robustness would be to track variable values, but I think that's only possible under an interpretation approach, and there are certainly cases where it would break down.

PurelyApplied:
I think I bit off a bit much in this PR, but the conversation was valuable.  From this, I'm taking the following action items as future PRs:

In the short term:
* YAML only really buys us comments (which I really want) and anchors (which could be useful, but aren't mission critical).  Although (a) the YAML marshal pattern might be easier to work with than the JSON marshal, and (b) YAML is a superset of JSON so we wouldn't have to change existing config.  As a consequence, I might be inclined to do this sooner rather than later.
* Refactor `config`'s interface to take basic arguments rather than `ssa` arguments.
* Permit string matchers to be either exact or regexp.  Distinguish these by name, e.g. `package` matches exactly vs `packageRE`.
* Add `description` and `exclude` fields to matchers.

The above should be doable without significantly changing how we interact with the config, call-sites notwithstanding

In the medium term:
* Distinguish value, call, and type matchers.
* Remove distinction implicit in our current separation of source by type | taint by value | sink by call.
* This unblocks identification of, e.g., `os.Getenv(...)` as source specification.

PurelyApplied:
Most of the above items have been merged.  I'll open an Issue about adding `description` fields.  Closing this PR.

PR #159, opened 2020-10-21
Add sanity check to automatic detection of test directories.

PurelyApplied:
There is an awkward interaction between `git` and the implicit building of packages for testing from `testdata` directories.

Because `git` only tracks files, not directories, it is possible after changing branches to have an empty directory after `git` has moved or deleted files.  Because we assume that every folder in `testdata` can be converted into an associated Go package, testing fails when it finds an empty directory in testdata.

This PR examines test directories for Go files and produces a more helpful error message if any are empty.

Test locally with `mkdir .../go-flow-levee/internal/pkg/levee/testdata/src/example.com/tests/empty`.  

- [x] Tests pass

### DIFF
--- a/internal/pkg/levee/levee_test.go
+++ b/internal/pkg/levee/levee_test.go
@@ -70,6 +70,6 @@ func checkForGoFiles(path string) error {
 			return nil
 		}
 	}
-	return fmt.Errorf("found no Go files in test directory (%s).  You may need to run `git clean`.", path)
+	return fmt.Errorf("found no Go files in test directory (%s)", path)
 
 }
--- a/internal/pkg/levee/levee_test.go
+++ b/internal/pkg/levee/levee_test.go
@@ -16,8 +16,10 @@ package internal
 
 import (
 	"flag"
+	"fmt"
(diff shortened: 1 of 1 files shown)


mlevesquedion:
This has happened to me in the past and my fix was to delete the directory.

I don't think this check is needed, but I also don't think including it hurts us in any way, so if you think this check can be helpful, I'm happy to approve it.

I've tested it locally and can confirm that it works.

(My "clean code" sense is telling me I don't think our testing code should know about our VCS. However, it's highly unlikely that we'll ever use anything other than Git, so this concern is largely irrelevant.)

mlevesquedion (approved):

PurelyApplied:
I'll reword the error to be VCS agnostic.  But yeah, the opaque error message caught me off guard this morning, and I threw away some time wondering why `master` was suddenly failing locally but not on Jenkins.  It didn't occur to me that `dominance` tests had been renamed, so I went straight for a `git bisect` instead of, you know, examining the test failures.  I'm sure I'll forget, so I'd prefer a more meaningful error message where available.

PR #160, opened 2020-10-21
Introduce infer analyzer

### LINKED ISSUE #96: Infer from a source type any defined type using the source type as the underlying type

### LINKED ISSUE #97: Infer as a source type any type that contains a source type as a field

mlevesquedion:
This PR introduces an analyzer that identifies *named* types as sources from their underlying types and their fields, implementing the core logic needed for #96 and #97. **In the future, the output of this analyzer could be used to identify sources by type.** Therefore, it will integrate either with the `source` or the `sourcetype` analyzer.

The core of the analysis consists of building a "type graph" that indicates which types use which other types in their definition. Traversing the graph in reverse topological order allows "sourceyness" to be propagated from configured sources to types that refer to them.

When a type definition like the below is encountered, an edge is added from `B` to `A`. If `B` is a configured source, or an inferred source, `A` will also be identified as an inferred source when traversing the graph.
```go
type A B
```
**This information is obtained using the `ast`, since the fact that `A`'s underlying type is `B` is lost in the `ssa`.**

When a struct definition like the below is encountered, an edge is added from `B` to `A`.
```go
type A struct {
    b B
}
```
**This information is obtained using the `ssa`, since the `ssa` provides a convenient API for this purpose.``

In practice, things are a little bit more complicated: a type may contain multiple types, some of them named, some of them not named, e.g.:
```go
map[string][]map[Source][]string
```

Cycles can occur due to corecursive type definitions, e.g.:
```go
type A struct {
    b *B
}
type B struct {
    a *A
} 
```
In the topological sorting, nodes within a cycle may be in any order. This does not pose any particular issues to the analysis, and it does not impact the order for nodes outside of the cycle.

- [x] Tests pass
- [x] Running against a large codebase such as [Kubernetes](https://github.com/kubernetes/kubernetes) does not error out. (N/A, not exposed and not yet integrated with the main analyzer.)
- [x] Appropriate changes to README are included in PR (N/A)


internal/pkg/sourceinfer/analyzer.go
```diff
@@ -0,0 +1,336 @@
...
+}
+
+type selectorFinder struct {
+	foundSelectors   []*ast.SelectorExpr
+	foundIdentifiers map[*ast.Ident]bool
+}
+
+func (sf *selectorFinder) Visit(n ast.Node) ast.Visitor {
+	sel, ok := n.(*ast.SelectorExpr)
+	if !ok {
```
mlevesquedion, line 174: I considered using `if sel, ok := ...; ok {`, but with the other `if` below I found it produced too much nesting.

internal/pkg/sourceinfer/analyzer.go
```diff
@@ -0,0 +1,336 @@
...
+		}
+		order = append(order, o)
+	}
+	for o := range graph {
+		visit(o)
+	}
+	return order
+}
+
+func isSourceType(c *config.Config, t types.Type) bool {
```
mlevesquedion, line 310: This is duplicated from `source.go`. It is my intent that the `source` analyzer will not need this function once it consumes this analyzer, either directly or via the `sourcetype` analyzer.

In the meantime, I could move it to `utils` if you think that would make more sense.

mlevesquedion: Alternatively, this could easily be re-implemented in terms of `findNamedTypes` (defined above).

internal/pkg/sourceinfer/testdata/src/example.com/tests/core/corecursive.go
```diff
@@ -0,0 +1,45 @@
...
+	}
+
+	D struct { // want D:"inferred source"
+		e *E
+	}
+
+	E struct { // want E:"inferred source"
+		c *C
+	}
+)
```
PurelyApplied, line 45: Your graph approach probably covers this case, but for maximum edge-casing, I'd throw in the following test case:

```golang
type Foo struct {
  neighbors map[string]Foo
}

type Bar struct {
  neighbors map[string]Bar
  s Source
}
```

This sort of recursion forms a tautology, where `Foo` is a source type only if `map[string]Foo` is a source type, only if `Foo` is a source.  That ambiguity should default to "not a source", but should be broken by ties like in `Bar`.

I guess you could probably get the same result with just a `type A struct { *A }`.  So some directly-recursive tests might be good.  Negative co-recursive cases would also be good.

PurelyApplied (commented):
I think it would be good to be explicitly clear on what you are and aren't trying to do here, and how it is intended to be consumed.  My understanding is as followed, but being less familiar with the `ast` analysis, it could be off.

* Scan the ast for _named_ type declarations.
* Build a graph connecting types
* Bleed out source labeling from "primary" source types to those which consume them in type or field declarations
* In the future, source identification by type will delegate / call to this set of identified types

Commented below, but the limitation to named types could be a concern, depending on intended consumption.

It also seems odd to be consuming the `ssa` and coercing it back into the `ast`.  Is there a reason to not just use the `ssa` types?  If I were to approach this problem, I would probably do the following:

* Examine `ssa.Member`s, filter to `ssa.Type`s, identify primary sources.
* Traverse the `ssa.Node` graph.  Build a type graph similar to yours here of `types.Type` from `*ssa.Value.Type()`.
* Seed at primary sources, bfs to mark all types seen.

The only key difference between that `ssa` approach and the `ast` approach you've presented here is that the returned collection would be `types.Type` rather than `types.Object`, but again, I'm rather concerned about missing unnamed types.

mlevesquedion:
> My understanding is as follows: ...

Your understanding is correct. I'll update the description to be more explicit. The only thing I would amend is that the *ssa* is also scanned for field-to-named type relationships.

> Is there a reason to not just use the ssa types?

This is a good question, and I should probably have included something in the description in anticipation of it. The issue is that with a piece of code like `type A B`, the fact that `A` is derived from `B` is not available in the SSA. When you ask for the `Type()` of the relevant member, you get `A`, and when you ask for the `Underlying` type, you get the `struct` definition of `B`, but `B`'s name doesn't appear anywhere.

internal/pkg/sourceinfer/analyzer.go
```diff
@@ -0,0 +1,336 @@
...
+
+			typeBeingDefined := pass.TypesInfo.ObjectOf(ts.Name)
+
+			selectorFinder := &selectorFinder{nil, map[*ast.Ident]bool{}}
+			ast.Walk(selectorFinder, (ast.Node)(ts.Type))
+			for _, sel := range selectorFinder.foundSelectors {
+				named, ok := pass.TypesInfo.TypeOf(sel).(*types.Named)
+				if !ok {
+					continue
+				}
```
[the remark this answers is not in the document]
mlevesquedion: You are correct. The current approach is restricted to named types. For a case like this, I think instead of returning a slice of `types.Object`, we may want to return a `IsSource` function, which in this case would scan the fields for named types and determine if any of them was identified as a source during the analysis.

To me, it would make more sense to add that functionality in a future PR, since this PR is already quite big .

PurelyApplied:
(I'm still not saying that `ast` is a bad approach, even if I might personally prefer it.)

It looks like `Type().Underlying()` distinguishes separate equivalent structs.  So you can detect `type A Source` with `a.Type().Underlying() == s.Type().Underlying()`, while not matching `NotSource` that shares type equality.

```golang
// added to testcase source.go
type A Source

// In a hack rework of sourceType analyzer
// ...
a, s, n := // *ssa.Type for A, Source, and NotSource respectively

au, su, nu := a.Type().Underlying(), s.Type().Underlying(), n.Type().Underlying()

fmt.Printf("au == su: %v\n", au == su)  // true
fmt.Printf("nu == su: %v\n", nu == su)  // false
fmt.Printf("au == nu: %v\n", au == nu)  // false
```

You can pull purelyapplied:infer-sources if you want to reprod yourself.

internal/pkg/sourceinfer/analyzer.go
```diff
@@ -0,0 +1,336 @@
...
+	}
+
+	ins.Preorder(genDeclFilter, func(n ast.Node) {
+		gd, ok := n.(*ast.GenDecl)
+		if !ok {
+			return
+		}
+		for _, spec := range gd.Specs {
+			// type definition
+			ts, ok := spec.(*ast.TypeSpec)
```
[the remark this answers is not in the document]
mlevesquedion: Good catch! Done.

mlevesquedion:
Very interesting! I hadn't thought to compare the `Underlying()` values directly, for some reason. The other approach I had in mind was to produce some kind of descriptor containing the names and types of the fields. But that would have failed to distinguish equivalent structs.

Thank you for looking into this. I'll give this approach a shot and see how it reads. The `ast` code is not particularly complicated, but the `ssa` code will likely be more direct, and using only one of those would certainly simplify the code.

internal/pkg/sourceinfer/testdata/src/example.com/tests/core/corecursive.go
mlevesquedion: These are very good cases. I've added negative corecursive cases, as well as positive and negative self-recursive cases.

mlevesquedion:
I took a look at using `Underlying()`. I think it would be possible to make it work, but it would be more complicated than using the `ast`.

With the `ast`, it is easy to create an edge from `A` to `B` because we can directly look at the type definition `type B A`.

With the `ssa`, we can't add an edge directly, because we don't have a way to determine that `B` was defined from `A`. Using `Underlying()`, we can find types that have the same `Underlying()` type, but we don't know which type is defined from which. So _a priori_, we have to add edges in both directions. If we have a group of types that all have the same `Underlying()`, then we need to add the full set of edges. This is undesirable from a time/space complexity standpoint, but in practice I doubt it would really be an issue. More importantly, it is less obvious than adding an edge based on a type definition.

I think the `ast` approach is simpler and should be preferred. Of course, this is under the assumption that we want to use a graph traversal as the core "source inference" algorithm. I think this problem lends itself well to a graph-based approach, modelling the relationships between types as a graph feels natural and allows us to use standard algorithms.

mlevesquedion:
I've added tests for types defined in functions and simplified the code by removing any reliance on the `ssa` package.

Unnamed types will be handled in a future PR.

internal/pkg/sourceinfer/analyzer.go
```diff
@@ -0,0 +1,285 @@
...
+	"golang.org/x/tools/go/analysis"
+	"golang.org/x/tools/go/analysis/passes/buildssa"
+	"golang.org/x/tools/go/analysis/passes/inspect"
+	"golang.org/x/tools/go/ast/inspector"
+)
+
+// ResultType is a set of types.Object that are inferred Sources.
+type ResultType map[types.Object]bool
+
+type inferredSource struct{}
```
PurelyApplied, line 38: Consider renaming to `inferredSourceFact` to clarify the type is not referring to the inferred source objects themselves.

internal/pkg/sourceinfer/analyzer.go
```diff
@@ -0,0 +1,285 @@
...
+
+Similarly, the following type definition would cause Baz to be identified
+as a source, because it holds a field of type Foo:
+type Baz struct {
+  f Foo
+}
+Indeed, if Foo contains sensitive data, and Bar contains Foo, Bar also
+contains that data, via Foo.
+
+The analysis is recursive: types identified as sources during analysis
```
PurelyApplied, line 66: Nit: given the graph you build out of types, this algorithm is not recursive, is it?  You maybe meant it as the closure suggested above?

internal/pkg/sourceinfer/analyzer.go
PurelyApplied: I think I mentioned it offline, but if the boundary for this analyzer is to identify named source types, it likely should replace the current `sourcetypes` analyzer.  It makes sense, too, for it to occur early in the analyzer dependency chain, with `ast` cutting lower level than `ssa`.

That should be a future PR, because I would like to maintain the ability to see identified sources and fields that are currently presented by `sourcetypes`.  Having this analyzer absorb that feature can be part of the PR that replaces the `sourcetypes` analyzer.

internal/pkg/sourceinfer/analyzer.go
```diff
@@ -0,0 +1,285 @@
...
+will be used to evaluate the sourceyness of types that depend on them.
+For example, in the below type definitions, both Qux and Quux will be
+identified as sources:
+type Qux Foo
+type Quux Qux
+`,
+	Run: run,
+	Requires: []*analysis.Analyzer{
+		inspect.Analyzer,
+		buildssa.Analyzer,
```
PurelyApplied, line 76: `buildssa.analyzer` is no longer required.

PurelyApplied (approved):
Mentioned in a comment thread, but to reiterate because I think it's important: I think it's good to draw that explicit boundary at identifying `types.Object` source types and to position this analyzer early in our chain.  This analyzer should replace the current `sourcetypes` analyzer, possibly expanding identification to distinguish primary sources, inferred sources.

internal/pkg/sourceinfer/analyzer.go
mlevesquedion: That makes sense to me!

internal/pkg/sourceinfer/analyzer.go
mlevesquedion: Good point. Done.

internal/pkg/sourceinfer/analyzer.go
mlevesquedion: In my understanding, the *analysis* is recursive, since inferring that a type is a source leads the analysis to try to infer further sources from that type. Actually, in this case I guess *iterative* could also work...

Whether the *algorithm* itself is recursive is debatable, since it's basically DFS. There is a recursive DFS in there (toposort) and a non-recursive DFS as well (the sourceyness propagation proper).

What I'm hearing is that the term "recursive" here isn't very helpful, so I've removed that part altogether. I think the resulting description is clearer.

mlevesquedion:
Thank you for reiterating. That comment thread actually slipped past me originally, I only caught it while re-reviewing everything.

I think differentiating configured sources from inferred sources could be useful. We may event want to be able to give the inference path, which should be easy using the graph.

PR #162, opened 2020-10-29
Use funcmatcher for exclusion matchers.

PurelyApplied:
* Add additional tests for exclusion of methods with receivers.

- [x] Tests pass

------

Minor refactor.  I was concerned about excluding from exclusion any method attached to a receiver, since we were manually constructing the path for a given function.  Associated tests added.

### DIFF
--- a/configuration/README.md
+++ b/configuration/README.md
@@ -75,12 +75,14 @@ To match only methods without any receiver (i.e., a top-level function), use the
 
 ### Restricting analysis scope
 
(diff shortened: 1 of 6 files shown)


PurelyApplied:
As an aside, good call on adding the reminder to check the docs in the template.  I thought about it going in, did the changes, and then completely forgot to go update the docs.

mlevesquedion (approved):
LGTM

PR #164, opened 2020-10-30
Change config interface to take basic types.

PurelyApplied:
* Add utils methods to massage consumption of `string`-focused arguments.

- [x] Tests pass

----

Coming from discussion in #157, this is a refactor in the direction of better modularity - SSA logic should be owned by analyzers and string-matching logic should belong to `config`.


internal/pkg/fieldtags/analyzer.go
```diff
@@ -93,7 +93,7 @@ func (r ResultType) IsSourceFieldAddr(fa *ssa.FieldAddr) bool {
 	return r.IsSource(field)
 }
 
-// IsSource determines whether a types.Var is a source, that is whether it refers to a field previously identified as a source.
+// IsSourceType determines whether a types.Var is a source, that is whether it refers to a field previously identified as a source.
 func (r ResultType) IsSource(field *types.Var) bool {
```
mlevesquedion, line 97: The function's name no longer matches the doc comment. I think the name `IsSourceField` would be more appropriate given the function's behavior and parameter name.

mlevesquedion: More importantly though, to me it looks like this change belongs in a different PR.

internal/pkg/source/source_test.go
```diff
@@ -34,33 +30,27 @@ type testConfig struct {
...
-
-	match, _ := regexp.MatchString(c.sourcePattern, d.String())
+func (c *testConfig) IsSourceType(path, typename string) bool {
+	match, _ := regexp.MatchString(c.sourcePattern, typename)
 	return match
 }
 
-func (c *testConfig) IsSanitizer(call *ssa.Call) bool {
-	match, _ := regexp.MatchString(c.sanitizerPattern, call.String())
+func (c *testConfig) IsSanitizer(p, r, n string) bool {
```
mlevesquedion, line 38:
```suggestion
func (c *testConfig) IsSanitizer(path, recv, name string) bool {
```
These names feel a bit too concise, and they are inconsistent with the other functions' signatures.

internal/pkg/utils/utils.go
```diff
@@ -45,3 +46,44 @@ func FieldName(fa *ssa.FieldAddr) string {
 	}
 	return st.Field(fa.Field).Name()
 }
+
+// DecompoeType returns the path, typename, and indicators for if the Type is Named or an Interface
```
mlevesquedion, line 50:
```suggestion
// DecomposeType returns the path and name of a Named type.
```

internal/pkg/utils/utils.go
```diff
@@ -45,3 +46,44 @@ func FieldName(fa *ssa.FieldAddr) string {
...
+	dotPos := strings.LastIndexByte(packageQualifiedName, '.')
+	if dotPos == -1 {
+		return packageQualifiedName
+	}
+	return packageQualifiedName[dotPos+1:]
+}
+
+// DecomposeFunction returns the path, receiver, and name strings of a ssa.Function.
+// For functions that have no receiver, returns an empty string for recv.
+// If f is nil, returns empty strings for all return values.
```
mlevesquedion, line 76: When would `f` be `nil`?

internal/pkg/utils/utils.go
```diff
@@ -45,3 +46,44 @@ func FieldName(fa *ssa.FieldAddr) string {
...
+	}
+
+	if pkg := n.Obj().Pkg(); pkg != nil {
+		path = pkg.Path()
+	}
+
+	return path, n.Obj().Name()
+}
+
+func UnqualifiedName(v *types.Var) string {
```
mlevesquedion, line 65: This isn't used outside this package, so it should not be exported.

internal/pkg/utils/utils.go
```diff
@@ -45,3 +46,44 @@ func FieldName(fa *ssa.FieldAddr) string {
...
+
+// DecomposeFunction returns the path, receiver, and name strings of a ssa.Function.
+// For functions that have no receiver, returns an empty string for recv.
+// If f is nil, returns empty strings for all return values.
+func DecomposeFunction(f *ssa.Function) (path, recv, name string) {
+	if f == nil {
+		return
+	}
+
+	path = f.Pkg.Pkg.Path()
```
mlevesquedion, line 82: This looks potentially dangerous to me, given this comment in `ssa.go`, line 306:
```go
	Pkg       *Package      // enclosing package; nil for shared funcs (wrappers and error.Error)
```
(in the definition for `ssa.Function`)

internal/pkg/utils/utils.go
```diff
@@ -45,3 +46,44 @@ func FieldName(fa *ssa.FieldAddr) string {
...
+	name = f.Name()
+	recvVar := f.Signature.Recv()
+	if recvVar != nil {
```
mlevesquedion, line 85: Consider bringing `recvVar` into the `if`'s head.

internal/pkg/config/config_test.go
```diff
@@ -40,15 +41,15 @@ func runTest(pass *analysis.Pass) (interface{}, error) {
 	}
 
 	for _, f := range in.SrcFuncs {
-		if conf.IsSinkFunction(f) {
+		if conf.IsSink(utils.DecomposeFunction(f)) {
```
mlevesquedion, line 44: I think this inlined call looks quite nice.

mlevesquedion: On the other hand, I'm not sure I like the naming. A few thoughts:
1. I think the name `utils` has overstayed its welcome, and it would help to have a more specific name. (I think it might make sense to have two packages, `ssautils` and `typeutils`, but that wouldn't be a huge gain).
2. The name `DecomposeFunction` sounds quite nice to me, but I think it is too general to accurately describe the behavior. Really, this is just grabbing the name, recv and pkg of the function, but to me `Decompose` hints that something deeper is going on. This comment also applies to `DecomposeType`. (A suggestion: maybe it would read better to have a function named `DescribeFunction` that would produce a `functionDescriptor` struct with the relevant fields, and have the config take that instead.)

mlevesquedion (commented):
I think these changes are doing a good job of isolating the config from the details of `ssa` and `type`. However, I feel like the name `utils` is no longer appropriate. I also think tests should be added to cover the new functions in the `utils` package.

internal/pkg/fieldpropagator/analyzer.go
```diff
@@ -86,6 +87,9 @@ func run(pass *analysis.Pass) (interface{}, error) {
 }
 
 func analyzeBlocks(pass *analysis.Pass, conf *config.Config, tf fieldtags.ResultType, meth *ssa.Function) {
+	if meth == nil || meth.Signature == nil {
```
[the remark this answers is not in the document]
PurelyApplied: It was a habitual nil-safety check for the following `meth.Signature.Results()`, but yeah, it shouldn't be necessary.  Removed.

internal/pkg/fieldtags/analyzer.go
PurelyApplied: I think it was accidentally swept up by my IDE during iteration / a renaming.  Reverted.

internal/pkg/utils/utils.go
PurelyApplied: I've got to get better about reviewing my docstrings...

internal/pkg/utils/utils.go
PurelyApplied: Safety check added.

internal/pkg/utils/utils.go
PurelyApplied: For instance,

```golang
// levee.go:62
				case conf.IsSink(utils.DecomposeFunction(v.Call.StaticCallee())):
```

`StaticCallee()` may be nil for builtins and dynamic-dispatch calls.  See [the `CallCommon` docs](https://godoc.org/golang.org/x/tools/go/ssa#CallCommon) for the particulars.

internal/pkg/config/config_test.go
PurelyApplied: I agree that `Decompose*` is not a great name, and that `utils` is also a bit too ambiguous.  Will update soon.

I agree with making a more generic interface to the `type Config` itself, but that feels like a future PR.

internal/pkg/utils/utils.go
mlevesquedion: Oh, right! Thanks for explaining.

I feel like handling that is the caller's responsibility. To me, `DecomposeFunction`'s job is to take a `Function` and decompose it into `path`, `recv` and `name`. It shouldn't be called with a `nil` value. The caller shouldn't be trying to decompose a non-existent `StaticCallee()`, so if `StaticCallee()` is `nil`, it shouldn't call `DecomposeFunction` at all. 

I think it would be clearer to have the `nil` check in the caller. When I see `v.Call.StaticCallee()`, I can easily see why that might be `nil`. When I see `if f == nil {`, it's not obvious to me why `f` could be nil, and in fact I don't think it should ever be.

internal/pkg/config/config_test.go
PurelyApplied: I've been staring at this for a short while now, and for the life of me, I can't come up with a better name than `Decompose`.  `FunctionDetails` seems equally generic.  `PathTypeAndNameStrings(f)` seems cumbersome.  `IDStrings`?  I don't know.

As for `utils`, I'm just going to end up pulling three functions into their own sub-utils.  Since they're intended to massage the `config` interface, I suppose they could live in `config`.  That's where there were during iteration, but it felt like violating the "`config` shouldn't know about `ssa`" ideal.

Do you have any suggestions for names?

mlevesquedion: I've thought about the naming some more and sadly I don't have a very solid suggestion to make. I think `Decompose*` is fine. Having a `functionDescriptor` struct could be a nice approach, but maybe not.

internal/pkg/utils/utils.go
```diff
@@ -45,3 +46,42 @@ func FieldName(fa *ssa.FieldAddr) string {
 	}
 	return st.Field(fa.Field).Name()
 }
+
+// DecomposeType returns the path and name of a Named type
+// Returns empty strings if the type is not *types.Named
+func DecomposeType(t types.Type) (path, name string) {
```
mlevesquedion, line 52: Similarly to `DecomposeFunction`, should this take a `types.Named` directly?

PurelyApplied: Could, but bubbling up a type assertion is going to be messier than bubbling up a nil-check.  As it is, we take a `Type`, and we know that some `Type`s don't have names.  Either way, you can pass what you get to a matcher, and they are very unlikely to match anything, so it all shakes out.

Alternatively, something relatively simple here

```golang
		ssaType, ok := mem.(*ssa.Type)
		if !ok || !conf.IsSourceType(utils.DecomposeType(ssaType.Type())) {
			continue
		}
```

needs a second layer of assertion
```golang
		ssaType, ok := mem.(*ssa.Type)
		if !ok {
			continue
		}
		namedType, ok := ssaType.Type().(*types.Named)
		if ok && !conf.IsSourceType(utils.DecomposeType(namedType)) {
			continue
		}
```

Really, it comes down to the fact that the most common argument into `DecomposeType` is what comes out of `Dereference`, which is a `Type`.  We could change that to return only `Named`, or convert to some `DereferenceToUnderlyingNamed` or something, but that all feels beyond the scope of this PR.

mlevesquedion (approved):
LGTM

PR #165, opened 2020-11-02
Add tests for select statements

mlevesquedion:
This PR adds tests covering select statements.

Some cases are currently failing, I am investigating why that is the case. 

Some cases involve interprocedural interactions and are not expected to pass.

- [x] Tests pass
- [] (N/A) Running against a large codebase such as [Kubernetes](https://github.com/kubernetes/kubernetes) does not error out.
- [] (N/A) Appropriate changes to README are included in PR

### DIFF
--- a/internal/pkg/levee/testdata/src/example.com/tests/select/tests.go
+++ b/internal/pkg/levee/testdata/src/example.com/tests/select/tests.go
@@ -1,0 +1,148 @@
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package sel
+
+import (
+	"example.com/core"
+)
+
+func TestSelectSourceSingleCase(sources <-chan core.Source) {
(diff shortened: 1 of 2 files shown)


internal/pkg/levee/testdata/src/example.com/tests/select/tests.go
```diff
@@ -0,0 +1,125 @@
...
+		core.Sink(s) // want "a source has reached a sink"
+	default:
+	}
+}
+
+func TestTaintedAndSinkedInDifferentBranches(objects chan interface{}) {
+	select {
+	case objects <- core.Source{}:
+	case s := <-objects:
+		// TODO: we do not expect a report here, because objects is only tainted if the
```
PurelyApplied, line 64: Were we trying to keep the `TODO want` as a searchable pattern for suppressed tests?  I don't have any hard opinions either way, since IntelliJ has a `TODO` pane.

I'm of mixed opinion as to whether or not I'm okay with this emitting a Diagnostic in any case.  Partially in the "prefer false-positive over false-negative," but I don't think I can formally justify it in this case.  I'm going to mull over it for a while...

PurelyApplied (approved):

internal/pkg/levee/testdata/src/example.com/tests/select/tests.go
mlevesquedion: > Were we trying to keep ...
Absolutely, my bad.

>  I'm of mixed opinion ...
Actually, now that I think of it, if we're putting a source in the channel in one branch, then that is an indication that this channel may contain sources in general, so we may want to consider the channel to be tainted. However, this is a bit of a slippery slope. Currently we have similar tests involving `if` statements, where a value is tainted in one branch and sinked in the other. In these tests, we do not expect a report to be produced. We may also want to differentiate between locally created values and values received through parameters (this distinction may be moot if we implement interprocedural analysis).

PR #166, opened 2020-11-02
Do not traverse to lookup index

mlevesquedion:
This PR fixes an issue where in some cases we aren't traversing to values that we should be traversing to.

Specifically, consider this case:
```go
func Test() {
	xs := []interface{}{nil}
	xs[0] = core.Source{}
	core.Sink(xs) // want "a source has reached a sink"
}
```
This test case is actually failing right now because of the following check in `reachableFromSource`:
```go
if sInstr.Block() == target.Block() && sIndex > targetIndex {
        return false
} 
```
In summary: if the target instruction occurs before the Source-producing instruction, then the target instruction is not reachable from the source (so don't traverse to it).

This test case was added to prevent an unexpected report in this case:
```go
func TestDoesNotReachSinkAfterSourceThroughValueCreatedBeforeSource() {
	// Taint should not propagate to this value.
	k := newKey()

	_ = map[string]core.Source{}[k.name]

	core.Sink(k.Name())
}
```

It seems that the "fix" that I did in `reachableFromSource` was a little excessive. To handle this case properly, it suffices to avoid traversing to the `Index` in a `Lookup`. Indeed, looking up a value in a tainted map does not taint the key (e.g.).

I'm not sure what other test cases might be good to have. There are probably other cases in which we want to avoid incorrect propagation, but due to the risk of false negatives I think it is best to remove `reachableFromSource` altogether for now. (Removing it fails no tests).

- [x] Tests pass

### DIFF
--- a/internal/pkg/source/source.go
+++ b/internal/pkg/source/source.go
@@ -223,8 +223,9 @@ func (s *Source) visitOperands(from ssa.Node, operands []*ssa.Value, maxInstrRea
 			}
 		}
 
-		// don't traverse to the key in a lookup
-		// if a map is tainted, looking up a value in the map doesn't taint the key
+		// Don't traverse to the key in a lookup.
+		// For example, if a map is tainted, looking up a value in the map
+		// doesn't taint the key, so we shouldn't traverse to the key.
 		if look, ok := from.(*ssa.Lookup); ok && *o == look.Index {
 			continue
 		}
--- a/internal/pkg/source/source.go
+++ b/internal/pkg/source/source.go
@@ -97,9 +97,6 @@ func (s *Source) dfs(n ssa.Node, maxInstrReached map[*ssa.BasicBlock]int, lastBl
 	}
 
 	if instr, ok := n.(ssa.Instruction); ok {
-		if !s.reachableFromSource(instr) {
-			return
-		}
 		// If the referrer is in a different block from the one we last visited,
 		// and it can't be reached from the block we are visiting, then stop visiting.
 		if ib := instr.Block(); lastBlockVisited != nil &&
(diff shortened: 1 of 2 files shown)


internal/pkg/source/source.go
```diff
@@ -222,15 +192,15 @@ func (s *Source) canReach(start *ssa.BasicBlock, dest *ssa.BasicBlock) bool {
 	return false
 }
 
-func (s *Source) visitOperands(n ssa.Node, operands []*ssa.Value, maxInstrReached map[*ssa.BasicBlock]int, lastBlockVisited *ssa.BasicBlock) {
+func (s *Source) visitOperands(from ssa.Node, operands []*ssa.Value, maxInstrReached map[*ssa.BasicBlock]int, lastBlockVisited *ssa.BasicBlock) {
```
mlevesquedion, line 195: I am renaming this because it was being shadowed inside the loop. I did not notice the shadowing and wasted time debugging.

PurelyApplied (approved):
Statement: We need to formalize the `switch` for all `Instruction`s to propagate taint between values definitively.
Cardinality: `n = n+1`

mlevesquedion:
On the one hand, I agree that handling special cases in this way is ugly and makes the code harder to reason about. I'm not sure that listing all possible instructions is the solution, however.

The current approach is basically "traverse by default". In general, it doesn't matter what the instruction is: we will traverse to it, unless there is a specific restriction. If we listed every instruction, then I'm worried we would end up with a lot of duplicate or near-duplicate code. Most branches would be an explicit version of "traverse to everything". What's more, *we would still need the special cases*. So I think if we were to explicitly handle each instruction, we would be in a worse situation.

Somewhat related: we should certainly aim to have a test suite that covers every kind of instruction, in a variety of situations.

PR #168, opened 2020-11-02
Add staticcheck to CI

mlevesquedion:
I have found dead code with `staticcheck` twice (#158, #167), so I think it is a good idea to include it in the CI.

In addition to dead code, [staticcheck](https://staticcheck.io/) performs a wide range of code quality analyses. It isn't reporting anything right now, but over time I suspect it will help us keep the code base a bit cleaner. 

- [x] Tests pass


PurelyApplied (approved):
I don't love the install-and-check pattern, since it makes our tests order-dependent.  I don't think it's worth teasing out in Travis, but might be a reason to migrate to Workflows in the future.

mlevesquedion:
Agreed!

PR #169, opened 2020-11-03
Replace JSON configs with YAML

mlevesquedion:
Now that YAML is the recommended way to configure the analyzer, in this PR I propose removing (almost) all traces of JSON from our codebase.

I think having every configuration file be YAML has two main benefits:
1. Consistency
2. YAML is easier to read and edit (this may be personal preference talking)

The only remaining traces of JSON are in `config/regexp/*.go`. This code uses `json` Unmarshaling, which is compatible with the `yaml` package. I think it's fine to leave it like that.

- [x] Tests pass


configuration/README.md
```diff
@@ -73,7 +73,7 @@ As just two examples, this may be used to avoid analyzing test code, or to suppr
 
 The following configuration could be used to identify possible instances of credential logging in Kubernetes.
 
-[example-config.json](example-config.json)
+[example-config.yaml](example-config.yaml)
```
mlevesquedion, line 76: The actual renaming was done in #163. This is just modifying the pointer to point to the new name.

configuration/example.sh
```diff
@@ -5,7 +5,7 @@ GB=$(go env GOBIN) GB=${GB:-$(go env GOPATH)/bin}
 # go vet dislikes relative pathing.
 SCRIPT_DIR=$( cd "$(dirname "$0")" >/dev/null 2>&1 ; pwd -P )
 # analysis arguments
-CONFIG_FILE=${SCRIPT_DIR}/example-config.json
+CONFIG_FILE=${SCRIPT_DIR}/example-config.yaml
```
mlevesquedion, line 8: The actual renaming was done in #163. This is just modifying the pointer to point to the new name.

.travis.yml
```diff
@@ -28,12 +28,11 @@ script:
   )
 - >
   echo "Checking for license headers..." && (
-    FILES=$(grep -rL 'https://www.apache.org/licenses/LICENSE-2.0' cmd internal pkg | grep -vE ".(json|ssa)")
+    FILES=$(grep -rL 'https://www.apache.org/licenses/LICENSE-2.0' cmd internal pkg | grep -vE ".(yaml|ssa)")
```
PurelyApplied, line 31: Presumably just swept up in a find-and-replace, but the advantage of comments in YAML is that we can and should include the license header.

PurelyApplied (changes requested):
Looks good to me.  Thank you and your apparently enormous well of patience for wading through these.

Minor issue in the Travis, but LGTM overall.

mlevesquedion:
I must confess that I am actually very impatient. Mostly I just ran this command:

```bash
for f in $(fd json); do
    vim $f
done
```

And I had https://www.json2yaml.com/ open on the side.

.travis.yml
mlevesquedion: SG. Done.

PurelyApplied (approved):

PR #170, opened 2020-11-05
Remove tests that should have been deleted

mlevesquedion:
- [x] Tests pass


PurelyApplied (approved):
I may have thought, skimming the previous PR, that the test files would be inferred the way they are in `levee`.  We should perhaps become consistent in whether we infer test files or not.

mlevesquedion:
Good point. I just realized that `analysistest` understands `...`, so I'll see if we could just use that everywhere.

PR #171, opened 2020-11-05
Use fieldtags analyzer in sourceinfer analyzer

mlevesquedion:
This PR introduces the fieldtags analyzer in the sourceinfer analyzer. This allows sources identified by field tags to be used to infer further sources derived from them.

This PR depends on #173.

- [x] Tests pass


internal/pkg/sourceinfer/testdata/src/example.com/tests/core/field.go
```diff
@@ -47,6 +47,10 @@ type (
 	SourceHolderHolder struct { // want SourceHolderHolder:"inferred source"
 		sh SourceHolder
 	}
+
+	SourceByFieldTag struct { // want SourceByFieldTag:"inferred source"
+		secret string `levee:"source"`
+	}
 )
```
PurelyApplied, line 54: I'll preface this with saying we may need to formalize more precise language in general.

I would call `SourceByFieldTag` a primary source type, not an inferred one, since it contains a source as a field which was identified directly by configuration.  This doesn't like an inference to me.

I _would_ like to see a test case inferred from a field tag, i.e. `InferredFromTag struct { SourceByFieldTag }`.

internal/pkg/sourceinfer/analyzer.go
```diff
@@ -224,9 +231,13 @@ func inferSources(pass *analysis.Pass, conf *config.Config, objGraph objectGraph
 		if seen[o] {
 			continue
 		}
-		if !(isSourceType(conf, o.Type()) || pass.ImportObjectFact(o, &inferredSourceFact{})) {
+
+		if v, ok := o.(*types.Var); !(ok && taggedFields.IsSource(v)) &&
+			!isSourceType(conf, o.Type()) &&
+			!pass.ImportObjectFact(o, &inferredSourceFact{}) {
```
PurelyApplied, line 237: This conditional has gotten pretty heavy.  Consider using De Morgan's to collapse the negations, i.e.

```golang
		if v, ok := o.(*types.Var); !(
			ok && taggedFields.IsSource(v) ||
				isSourceType(conf, o.Type()) ||
				pass.ImportObjectFact(o, &inferredSourceFact{})) {
			continue
		}
```

mlevesquedion: Interestingly, when I wrote the conditional it started out in the form that you're suggesting, and then I used De Morgan's to turn it into what it is now. I think it's more readable as it is, but it's not very readable in either case.

I think this PR #173 will help alleviate this.

internal/pkg/sourceinfer/testdata/src/example.com/tests/core/field.go
mlevesquedion: Interesting. To me, when a struct holds a tagged field, the tagged field is the source, and we go on to determine that the struct is a source by virtue of holding the tagged field (some kind of inference, then).

I think your point of view makes sense though. To me it seems like the fieldtags analyzer should identify structs that hold tagged fields as sources, then. I have implemented this in #173.

mlevesquedion:
I have opened a PR at #173 that makes changes to the `fieldtags` analyzer that are required for proper integration with the `sourceinfer` analyzer. The current PR expands upon #173.

PurelyApplied:
Converted to draft until #173 merges.

internal/pkg/sourceinfer/testdata/src/example.com/tests/core/field.go
PurelyApplied: I think we should formalize some language here.  I'm going to use the following terms.  We can talk about explicitly codifying this in the future, but here are my definitions for this moment.

* A _source_ contains data to track.  This aligns well with `*ssa.Type` - it's data in a register somewhere.
* A _source container_ is a `struct` which could contain a source field.  A source container can be sanitized by, e.g., zeroing out the source field(s).
* A _source type_ is either a source container or explicitly named type intended to hold source, e.g. `type Token string`.
* A _primary_ source type is a source type identified directly by user configuration.
* An _inferred_ source type is a source type identified by this analyzer, e.g. `type Foo struct { Token }`.

-----

Using this nomenclature:

* `secret` is a source
* `SourceByFieldTag` a primary source type.
* `InferredFromTag` is an inferred source type.

-----

As mentioned in #173, I think the `fieldtags` analyzer is good where it is.  It's identifying source tags, and that's a good scope for it.  As mentioned in #160, I think this `sourceinfer` analyzer will come to replace `sourcetypes`.  In the near future, I would say that _this_ analyzer should be identifying primary sources.  It already does this for sources types identified by name.  It still needs to do this for source types identified by field tag.  That can be fixed almost trivially.

-----

It would be nice if this distinction were explicit in the diagnostics / facts we are checking.  My original comment 

> I would call `SourceByFieldTag` a primary source type, not an inferred one, since it contains a source as a field which was identified directly by configuration. This doesn't like an inference to me.

was made in that direction.  I don't think it's wrong or even out of scope for this analyzer to identify `SourceByFieldTag`.  I just didn't know about calling it an `"inferred source"`.

In either case, I think that distinction could be part of future iteration.  I don't think it's _wrong_ per se, just imprecise.

mlevesquedion: Thank you for formalizing those.

I think we should prefer *configured source* instead of *primary source*. I understand what you mean, but I think *configured source*'s definition is obvious by its name alone, whereas *primary source* seems tainted by the idea that it will be used for some kind of inference.

To me the most important distinction is *configured sources* vs *inferred sources*. For *inferred sources*, we may want to be able to explain the inference, e.g. "inferred source by virtue of having a field of type X, which is a source" (which you call a *source container*) or "inferred source by virtue of being defined from a type X, which is a source".

I may have been a bit quick to jump to making changes. When I read this part of your comment:

> I *would* like to see a test case inferred from a field tag, i.e. `InferredFromTag struct { SourceByFieldTag }`.

In the context of your previous statement, it sounded like you were saying "this test case shouldn't be among the `sourceinfer` analyzer's tests".

I think if we consider a type holding tagged fields to be a *configured source* then that logic should live in the `fieldtags` analyzer. If we consider it to be an *inferred source*, then that logic should live here. **To me, it is an *inferred* source.** The user configures field tags, and there is no way to tag a struct, so structs holding tagged fields cannot be directly configured to be sources.

> That can be fixed almost trivially.

Indeed, it was before I started changing everything :stuck_out_tongue: 

> In either case, I think that distinction could be part of future iteration. I don't think it's wrong per se, just imprecise.

SG. In that case, I'll close #173, revert this to an earlier state and let you know when this is ready for a fresh look.

PurelyApplied: > SG. In that case, I'll close #173, revert this to an earlier state and let you know when this is ready for a fresh look.

👍 

> `type InferredFromTag struct { SourceByFieldTag }`.

For clarity, I would indeed like to see that test in `sourceinfer`'s scope.

> To me, it is an inferred source.

Yeah, that's fine.  We haven't formally codified any of this language, so until we do, we can't really way one way or t'other.

> [All the rest]

We can start a thread to formalize and codify this somewhere else.  Maybe in an `Issue`.  Either way, this discussion shouldn't hold up this PR.

mlevesquedion:
I have reverted the changes to `fieldtags`, and I have extracted the bit of code that was making the conditional feel especially heavy. I think the conditional is reasonable now.

This is ready for review.

PurelyApplied (approved):

PR #172, opened 2020-11-05
Use test glob to run all tests

mlevesquedion:
This PR is in response to discussion on #170.

In packages where we just want to run all the tests, we can get away with using a `./...` pattern to catch all the test files.

- [x] Tests pass
- (N/A) [] Running against a large codebase such as [Kubernetes](https://github.com/kubernetes/kubernetes) does not error out.
- (N/A) [] Appropriate changes to README are included in PR

### DIFF
--- a/internal/pkg/levee/levee_test.go
+++ b/internal/pkg/levee/levee_test.go
@@ -16,10 +16,6 @@ package internal
 
 import (
 	"flag"
-	"fmt"
-	"io/ioutil"
-	"path/filepath"
-	"strings"
 	"testing"
 
 	"github.com/google/go-flow-levee/internal/pkg/debug"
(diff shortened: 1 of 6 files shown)


internal/pkg/levee/levee_test.go
```diff
@@ -33,43 +29,8 @@ func TestLevee(t *testing.T) {
...
-		if err := checkForGoFiles(path); err != nil {
-			t.Fatalf("Could not verify presence of Go files in test directory: %v", err)
-		}
-
-		patterns = append(patterns, path)
-	}
-	return
-}
-
-func checkForGoFiles(path string) error {
```
mlevesquedion, line 62: Empty directories are handled transparently by `./...`. `analysistest` won't complain about them.

PurelyApplied (approved):
Well now isn't that nice and clean.  🎉 

I see a few `analysistest.Run` calls that could still be converted.  They could stay how they are as outliers, just wanted to make sure it wasn't something you missed.

The one in `.../source/analyzer_test.go` only has a single target, but is still shares the basic pattern to be replaced with `./...`.  The others seems a bit precarious and could stay how they are without offending my sensibilities too much.

mlevesquedion:
:tada: 

Yeah, there's a few outliers in there that I think can remain with explicit patterns for now.

PR #173, opened 2020-11-06
Make fieldtags analyzer identify sources

mlevesquedion:
This PR is in preparation for inclusion in the `sourceinfer` analyzer: #171. It makes the `fieldtags` analyzer identify as sources types that hold at least one tagged field. It also accumulates the results of the analyzer over an import chain. This is needed so that e.g. if a tagged struct `Foo` is declared in package `foo` and imported in package `bar`, then in a type definition such as `type Bar foo.Foo` we can infer that `Bar` is a source because `foo.Foo` is a source.

- [x] Tests pass


internal/pkg/sourcetype/testdata/src/sourcetype/test.go
```diff
@@ -27,7 +27,7 @@ type SourceContainingTaggedField struct { // want SourceContainingTaggedField:"s
 
 type AliasStruct = Source // want AliasStruct:"source type"
 
-type NamedType Source // want NamedType:"source type"
+type NamedType Source // TODO want NamedType:"source type"
```
mlevesquedion, line 30: This is handled by the `sourceinfer` analyzer in #171.

PurelyApplied:
> This PR is in preparation for inclusion in the `sourceinfer` analyzer: #171. It makes the `fieldtags` analyzer identify as sources types that hold at least one tagged field.

This feels well beyond the scope of a field tags analyzer.  `fieldtags` is currently doing a good job of highlighting the `*types.Var` of fields that are tagged with one of the source-identifying tags.  I don't think it needs to do more than that.

> This is needed so that e.g. if a tagged struct `Foo` is declared in package `foo` and imported in package `bar`, then in a type definition such as `type Bar foo.Foo` we can infer that `Bar` is a source because `foo.Foo` is a source.

This sounds like something that falls in scope for the `sourceinfer` analyzer.  

I checked out #171, reverted all the changes in `fieldtags`, and corrected the call in `sourceinfer`.  This passes all tests present in #171 (after adding `// want Tagged:"inferred source"` to the `source.go`).

Am I missing something here?

mlevesquedion:
In [your comment](https://github.com/google/go-flow-levee/pull/171#discussion_r518269641) on #171, it sounded like you were saying that a struct with tagged fields should be identified as a source already, i.e. that it shouldn't be inferred to be a source.

Now you're saying that you want `Tagged` to be *inferred* as a source. But this exact test was already in #171:
```go
type SourceByFieldTag struct { // want SourceByFieldTag:"inferred source"
		secret string `levee:"source"`
}
```

Now I am also wondering what I might be missing :sweat_smile:.

PurelyApplied:
I think the real takeaway from that comment is that we need more precise language.

I hadn't seen your response in #171, continuing this thread there.

mlevesquedion:
Closing in response to [discussion](https://github.com/google/go-flow-levee/pull/171#discussion_r518997066) on #171.

PR #174, opened 2020-11-06
Do not infer sources from functions

mlevesquedion:
This PR prevents named types from being inferred as sources due to holding functions that have source arguments or return values.

As a happy side effect, the implementation of the fix also greatly simplifies the code.

## Context

This PR removes the `ast` visitors that were previously used to find objects. The issue with the visitors is that they were grabbing all `selectors` and `identifiers`, which included those in function type definitions, e.g. the `Source` in `func() Source`. It's possible to avoid those by adding logic to each visitor to stop traversing when they hit an `ast.FuncType` node, but I wasn't too keen to add some duplicated special case behavior, so I looked for a simpler way.

The key insight I had is that we don't actually need to traverse the `ast`. We can find the information we need by "traversing" the type. We already had some code for doing this, i.e. `findNamedTypes`. That code already avoids functions, because it does not traverse through `type.Signatures`, so types within functions aren't an issue. However, it now needs to traverse through `type.Structs` in order to find struct fields, because those can be tagged. Since fields now needed to be handled in this function, it made sense to extract the objects instead of the types. Indeed, we don't want `string` in our graph, but we do want the `types.Var` that represents a `string "levee:\"source\""` field. I think it also makes sense from the client's point of view: the function that creates the object graph is interested in the objects.

- [x] Tests pass

### DIFF
--- a/internal/pkg/sourceinfer/analyzer.go
+++ b/internal/pkg/sourceinfer/analyzer.go
@@ -115,27 +115,30 @@ func createObjectGraph(pass *analysis.Pass, ins *inspector.Inspector) objectGrap
 			return
 		}
 
-		typeBeingDefined := pass.TypesInfo.ObjectOf(ts.Name)
+		objectBeingDefined := pass.TypesInfo.ObjectOf(ts.Name)
 
-		for n := range findNamedTypes(pass.TypesInfo.TypeOf(ts.Type)) {
-			objGraph[n.Obj()] = append(objGraph[n.Obj()], typeBeingDefined)
+		for o := range findObjects(pass.TypesInfo.TypeOf(ts.Type)) {
+			objGraph[o] = append(objGraph[o], objectBeingDefined)
 		}
 	})
 
 	return objGraph
 }
 
-// findNamedTypes finds named types within a type that can potentially be sources.
-// Among other things, this excludes basic types and functions (Signatures).
-func findNamedTypes(t types.Type) map[*types.Named]bool {
-	namedTypes := map[*types.Named]bool{}
+// Given a type, findObjects finds objects associated with types
+// within that type that could be sources.
+// This includes named types and struct fields.
+// Functions (Signatures) and Interface types are avoided, because
+// those cannot be sources.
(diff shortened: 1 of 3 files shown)


PurelyApplied (commented):
The generic names of `findObjects` and `objects := map[...]...` and the in-function definition of the recursive `find` make it hard to follow along with what is meant to be happening here.  My expectation for a commit that says "do not infer sources from functions" would be that the switch where you are examining types would introduce some `func` logic.  What appears to happen is that we are more deeply traversing `Struct`s that we were previously skipping entirely.

I can observe the expected behavior in the tests and see the news tests fail if the other changes are reverted.  I guess maybe a more descriptive docstring on `findObjects` and on `find` could fix up my misunderstanding.

mlevesquedion:
Looking at the description again, I think I made a poor job of explaining what the changes are and why they are needed.

Starting from the beginning:

The issue with the visitors is that they were grabbing all `selectors` and `identifiers`, which included those in function type definitions, e.g. the `Source` in `func() Source`. It's possible to avoid those by adding logic to each visitor to stop traversing when they hit an `ast.FuncType` node, but I wasn't too keen to add some duplicated special case behavior, so I looked for a simpler way.

The key insight I had is that we don't actually need to traverse the `ast`. We can find the information we need by "traversing" the type. We already had some code for doing this, i.e. `findNamedTypes`. That code already avoids functions, because it does not traverse through `type.Signatures`, so types within functions aren't an issue. However, it now needs to traverse through `type.Structs` in order to find struct fields, because those can be tagged. Since fields now needed to be handled in this function, it made sense to extract the objects instead of the types. Indeed, we don't want `string` in our graph, but we do want the `types.Var` that represents a `string "levee:\"source\""` field. I think it also makes sense from the client's point of view: the function that creates the object graph is interested in the objects.

I have improved the comments and change the name `find` to `traverse`, which actually indicates what it is doing. PTAAL.

(I've also updated the PR description.)

PurelyApplied (approved):
The improved comments help, thanks.

PR #176, opened 2020-11-09
Add range tests

mlevesquedion:
This PR adds tests covering range operations over common collections.

Ranging over a slice uses an index. The index is compared with the length of the slice to determine when to stop. (Ranging over an array is pretty much identical, except that the length is known in advance).

Ranging over a map uses an iterator created with the `Range` instruction. The next element is obtained with the `Next` instruction.

Ranging over a channel uses an implicit `x, ok := <-chan` operation to determine when to stop.

- [x] Tests pass


PurelyApplied (approved):
Potentially out of scope, but should add a range over a tainted `[]interface{}`?  I.e.,

```golang
for _, s := range []interface{core.Source{Data: "secret} } {
  core.Sink(s)
}
```

mlevesquedion:
I think that's a reasonable case to add here. Added.

PurelyApplied (approved):

PR #177, opened 2020-11-09
Refactor: move result decls closer to usage

mlevesquedion:
A small refactoring for clarity. 

Moving declarations for analyzer results closer to where they are used to make it clearer which result is used in which step of the algorithm.

- [x] Tests pass

### DIFF
--- a/internal/pkg/sourceinfer/analyzer.go
+++ b/internal/pkg/sourceinfer/analyzer.go
@@ -87,10 +87,9 @@ func run(pass *analysis.Pass) (interface{}, error) {
 	}
 
 	ins := pass.ResultOf[inspect.Analyzer].(*inspector.Inspector)
-	ft := pass.ResultOf[fieldtags.Analyzer].(fieldtags.ResultType)
-
 	objectGraph := createObjectGraph(pass, ins)
 
+	ft := pass.ResultOf[fieldtags.Analyzer].(fieldtags.ResultType)
 	inferredSources := inferSources(pass, conf, ft, objectGraph)
 
 	return inferredSources, nil


PurelyApplied (commented):
I don't know that this reads better than the current.  "Start by loading previous analyzer results" seems like a good initial phase.  I guess you could drill this down into a conversation about the boons and banes of lazy object initialization, if you wanted.  You could probably bike-shed this in any direction you like, really.

* Why do we assign these to `ins` and `ft` rather than inline them directly into the single function calls that consume them?
* For that matter, why do we assign `objectGraph` rather than in-lining everything into the `return` statement?
* Why declare them at all, rather than pull them out of `pass` at lower levels?

If you feel really strongly about this, sure, we can move something down two lines.  It just feels superfluous in a 12 line function.

mlevesquedion:
I agree that it's not a big readability gain, but I do think it is a gain. My main intent was to make it clear which step is using which dependency. It's not exactly unclear right now either, since each one is used in only one function call and the function is quite short, as you mentioned.

Since overall (taking both of our views into account) it's not clear whether this is actually helpful, I'll close the PR.

For the record, some thoughts:

> Why do we assign these to ins and ft rather than inline them directly into the single function calls that consume them?

I think the expressions that produce `ins` and `ft` are complex enough that inlining them into the call would hurt readability.

> For that matter, why do we assign objectGraph rather than in-lining everything into the return statement?

It sounds like you're playing devil's advocate here, but I think that would be very hard to read.

> Why declare them at all, rather than pull them out of pass at lower levels?

I think this is an especially good question. My take:
* `pass` is a big heap of data and functionality.
* For me, analyzer results are specific pieces of data (even though, actually, they are typically functions or bundles of data and functions), so it makes sense to have them as a separate input to a particular step.
* It is convenient to just pass the whole `pass`, for easy access to all kinds of utilities.

This is just my way of seeing things, I'm sure there are other ways that are equally valid. I think the cleanest thing to do would be to define an interface that has just the `pass` methods that a particular step needs, but that has other problems.

Issue #130, opened 2020-09-24
Use propagation logic from the main analyzer in the fieldpropagator analyzer


mlevesquedion:
Currently, the analysis performed by the `fieldpropagator` analyzer to determine if a given method on a `Source` type is a `Field propagator` is very basic, and it only works for simple getters. As just one example, the following method is not identified as a `Field propagator`:

```go
func (s Source) DerivedFromData() string { // want DerivedFromData:"field propagator identified"
	derived := fmt.Sprintf("%s", s.data)
	return derived
}
```

The main analyzer would identify `derived` as being tainted by `s.data`, but the `fieldpropagator` analyzer does not. Currently, we do not propagate taint to the return value of a method on a `Source` type unless the method is a `Field propagator`. This means that we would not find a potential leak if it involves a `Field propagator` like the one above.

I think both analyzers should use the same propagation logic.

(This will also apply to interprocedural analysis when we get there.)

PR #178, opened 2020-11-10
Explicitly handle each node type

mlevesquedion:
This PR implements explicit handling of each node type. The idea is to make taint propagation a deliberate choice, instead of the current "just traverse to everything, except if X, Y or Z". Indeed, in most cases, propagating everywhere doesn't make much sense.

I think the resulting code is easier to understand. The epic `switch` looks intimidating, but for every node type, where to traverse is specified explicitly, in a single place. Also, `visitReferrers` actually means "visit the referrers" now (same for `visitOperands`).

Performing this work has made me discover some issues/potentially missing test cases:
- [x] `BinOp` needs tests, e.g. to make sure we aren't propagating to the other (non-tainted) operand : #179 
- [x] `Phi` needs tests, e.g. to make sure we aren't propagating to the other (non-tainted) operand : #180 
- [x] `MapUpdate` needs tests to make sure we are propagating only to the `Map`, not to the `Key` or the `Val`. #181
- [x] Tests for Index/IndexAddr : #187

The above issues have all been demonstrated via test cases, which are fixed by this PR.

For cases where the correct propagation is ambiguous, I have opened an issue here: #188. These will need further investigation. In the meantime, this PR maintains the current behavior of "traverse everywhere" for these cases.

- [x] Tests pass

### DIFF
--- a/internal/pkg/source/source.go
+++ b/internal/pkg/source/source.go
@@ -170,9 +170,6 @@ func (s *Source) visit(n ssa.Node, maxInstrReached map[*ssa.BasicBlock]int, last
 		s.visitReferrers(n, maxInstrReached, lastBlockVisited)
 		s.visitOperands(n, maxInstrReached, lastBlockVisited)
 
-	case *ssa.Lookup:
-		s.visitReferrers(n, maxInstrReached, lastBlockVisited)
-
 	case *ssa.FieldAddr:
 		deref := utils.Dereference(t.X.Type())
 		typPath, typName := utils.DecomposeType(deref)
@@ -183,40 +180,29 @@ func (s *Source) visit(n ssa.Node, maxInstrReached map[*ssa.BasicBlock]int, last
 		s.visitReferrers(n, maxInstrReached, lastBlockVisited)
 		s.visitOperands(n, maxInstrReached, lastBlockVisited)
 
-	case *ssa.Field:
-		// TODO: this should be the same as the FieldAddr branch
-		// don't forget to add the missing tests
-		s.visitReferrers(n, maxInstrReached, lastBlockVisited)
-		s.visitOperands(n, maxInstrReached, lastBlockVisited)
-
-	// TODO: add missing tests for each of these if applicable
-	// BinOp: tainted string concatenated with non-tainted string
-	// Phi: tainted in one possibility, sinked in both
-
-	// These nodes' operands should not be visited, because they can only receive
-	// taint from their operands, not propagate taint to them.
-	case *ssa.BinOp, *ssa.ChangeInterface, *ssa.ChangeType, *ssa.Convert, *ssa.Extract, *ssa.MakeChan, *ssa.MakeMap, *ssa.MakeSlice, *ssa.Phi, *ssa.Range, *ssa.Slice, *ssa.UnOp:
-		s.visitReferrers(n, maxInstrReached, lastBlockVisited)
-
-	// These nodes don't have operands; they are Values, not Instructions.
-	case *ssa.Const, *ssa.Global, *ssa.Parameter:
-		s.visitReferrers(n, maxInstrReached, lastBlockVisited)
-
-	// These nodes don't have referrers; they are Instructions, not Values.
-	case *ssa.Go, *ssa.Store:
-		s.visitOperands(n, maxInstrReached, lastBlockVisited)
-
 	// Only the Map itself can be tainted by an Update.
 	// The Key can't be tainted.
 	// The Value can propagate taint to the Map, but not receive it.
 	case *ssa.MapUpdate:
-		// TODO: add missing test (check that the key is not tainted, e.g.)
 		s.dfs(t.Map.(ssa.Node), maxInstrReached, lastBlockVisited, false)
 
 	// The only Operand that can be tainted by a Send is the Chan.
 	// The Value can propagate taint to the Chan, but not receive it.
 	case *ssa.Send:
 		s.dfs(t.Chan.(ssa.Node), maxInstrReached, lastBlockVisited, false)
+
+	// These nodes' operands should not be visited, because they can only receive
+	// taint from their operands, not propagate taint to them.
+	case *ssa.BinOp, *ssa.ChangeInterface, *ssa.ChangeType, *ssa.Convert, *ssa.Extract, *ssa.Field, *ssa.MakeChan, *ssa.MakeMap, *ssa.MakeSlice, *ssa.Phi, *ssa.Range, *ssa.Slice, *ssa.UnOp:
+		s.visitReferrers(n, maxInstrReached, lastBlockVisited)
+
+	// These nodes don't have operands; they are Values, not Instructions.
+	case *ssa.Const, *ssa.Global, *ssa.Lookup, *ssa.Parameter:
+		s.visitReferrers(n, maxInstrReached, lastBlockVisited)
+
+	// These nodes don't have referrers; they are Instructions, not Values.
+	case *ssa.Go, *ssa.Store:
+		s.visitOperands(n, maxInstrReached, lastBlockVisited)
 
 	// These nodes are both Instructions and Values, and have no special restrictions.
 	case *ssa.Index, *ssa.IndexAddr, *ssa.MakeInterface, *ssa.Select, *ssa.TypeAssert:
--- a/internal/pkg/source/source.go
+++ b/internal/pkg/source/source.go
@@ -91,34 +91,41 @@ func (s *Source) dfs(n ssa.Node, maxInstrReached map[*ssa.BasicBlock]int, lastBl
 		return
 	}
 
-	mirCopy := map[*ssa.BasicBlock]int{}
-	for m, i := range maxInstrReached {
-		mirCopy[m] = i
-	}
-
 	if instr, ok := n.(ssa.Instruction); ok {
+		instrIndex, ok := indexInBlock(instr)
+		if !ok {
+			return
+		}
+
 		// If the referrer is in a different block from the one we last visited,
 		// and it can't be reached from the block we are visiting, then stop visiting.
 		if lastBlockVisited != nil && instr.Block() != lastBlockVisited && !s.canReach(lastBlockVisited, instr.Block()) {
 			return
 		}
 
-		instIndex, ok := indexInBlock(instr)
-		if !ok {
-			return
-		}
-
-		// If this call's index is lower than the highest in its block,
+		// If this call's index is lower than the highest seen so far in its block,
 		// then this call is "in the past". If this call is a referrer,
 		// then we would be propagating taint backwards in time, so stop traversing.
 		// (If the call is an operand, then it is being used as a value, so it does
 		// not matter when the call occurred.)
-		if _, ok := instr.(*ssa.Call); ok && instIndex < maxInstrReached[instr.Block()] && isReferrer {
+		if _, ok := instr.(*ssa.Call); ok && instrIndex < maxInstrReached[instr.Block()] && isReferrer {
+			return
+		}
+	}
+
+	mirCopy := map[*ssa.BasicBlock]int{}
+	for m, i := range maxInstrReached {
+		mirCopy[m] = i
+	}
+
+	if instr, ok := n.(ssa.Instruction); ok {
+		instrIndex, ok := indexInBlock(instr)
+		if !ok {
 			return
 		}
 
-		if mirCopy[instr.Block()] < instIndex {
-			mirCopy[instr.Block()] = instIndex
+		if mirCopy[instr.Block()] < instrIndex {
+			mirCopy[instr.Block()] = instrIndex
 		}
 
 		lastBlockVisited = instr.Block()
--- a/internal/pkg/source/source.go
+++ b/internal/pkg/source/source.go
@@ -193,7 +193,7 @@ func (s *Source) visit(n ssa.Node, maxInstrReached map[*ssa.BasicBlock]int, last
 
 	// These nodes' operands should not be visited, because they can only receive
 	// taint from their operands, not propagate taint to them.
-	case *ssa.BinOp, *ssa.ChangeInterface, *ssa.ChangeType, *ssa.Convert, *ssa.Extract, *ssa.Field, *ssa.MakeChan, *ssa.MakeMap, *ssa.MakeSlice, *ssa.Phi, *ssa.Range, *ssa.Slice, *ssa.UnOp:
+	case *ssa.BinOp, *ssa.ChangeInterface, *ssa.ChangeType, *ssa.Convert, *ssa.Extract, *ssa.MakeChan, *ssa.MakeMap, *ssa.MakeSlice, *ssa.Phi, *ssa.Range, *ssa.Slice, *ssa.UnOp:
 		s.visitReferrers(n, maxInstrReached, lastBlockVisited)
 
 	// These nodes don't have operands; they are Values, not Instructions.
@@ -213,12 +213,9 @@ func (s *Source) visit(n ssa.Node, maxInstrReached map[*ssa.BasicBlock]int, last
 		s.dfs(t.X.(ssa.Node), maxInstrReached, lastBlockVisited, false)
 
 	// These nodes are both Instructions and Values, and have no special restrictions.
-	case *ssa.MakeInterface, *ssa.Select, *ssa.TypeAssert:
+	case *ssa.Field, *ssa.FreeVar, *ssa.MakeInterface, *ssa.Select, *ssa.TypeAssert:
 		s.visitReferrers(n, maxInstrReached, lastBlockVisited)
 		s.visitOperands(n, maxInstrReached, lastBlockVisited)
-
-	// FreeVars are handled by sourcesFromClosure.
-	case *ssa.FreeVar:
 
 	// These nodes cannot propagate taint.
 	case *ssa.Builtin, *ssa.DebugRef, *ssa.Defer, *ssa.Function, *ssa.If, *ssa.Jump, *ssa.MakeClosure, *ssa.Next, *ssa.Panic, *ssa.Return, *ssa.RunDefers:
--- a/internal/pkg/source/source.go
+++ b/internal/pkg/source/source.go
@@ -183,11 +183,13 @@ func (s *Source) visit(n ssa.Node, maxInstrReached map[*ssa.BasicBlock]int, last
 	// Only the Map itself can be tainted by an Update.
 	// The Key can't be tainted.
 	// The Value can propagate taint to the Map, but not receive it.
+	// MapUpdate has no referrers, it is only an Instruction, not a Value.
 	case *ssa.MapUpdate:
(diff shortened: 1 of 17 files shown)


PurelyApplied:
First off, 🎉.  Thank you for taking the time to look at each of the `Node` types.

> out of 42 node types

Just as a point of order, I only see 41 implemented and in the docs - 41 Nodes (30 Values, 35 Instructions, many both).  But that's still a heap load.

> Performing this work has made me discover some issues/potentially missing test cases:

I suspected this would happen, which is why I was pushing on it.  So double-thanks.

-----

High level comments:

Overall, I like it.  Good comments on the various switch case groups.  For consistency, I'd probably add a one-liner even to the more obvious ones, like `Call` or `FieldAddr`.  I might want to take a closer look into some of the instructions being skipped, like `Builtin`, `FreeVar` and maybe `Return`, but that might be lower-level than you're looking for right now.

I think this is pushing against our long-standing issue of `source` analyzer doing the actual propagation logic.  This seems pretty well modular, so it probably is orthogonal to that effort.

It might help to conceptualize the directional flow of taint better if we had a layer of separation between `Value` and `Instruction` in our propagation.  We are (and have long been) traversing from `Node` to `Node`, but conceptually, taint progresses from `Value` to `Value` along edges defined by `Instruction`s.  It might read better if we had start with a `Value`, got it's `referrers()`, and determined what neighboring `Value`s taint might flow through using a switch like this.  But that also feels like a refactor that is orthogonal to this work and could come in a future PR, if at all.

> `MapUpdate` needs tests to make sure we aren't propagating to the key, and that the key isn't tainting the map.

For the expression `m[k] = v`, I would expect:

* If `k` is tainted, `m` becomes tainted.  (`v` remains untainted.)
* If `v` is tainted, `m` becomes tainted.  (`k` remains untainted.)

We're probably on the same page, but your comment seems focused on the second case.  Or is it not a `MapUpdate` instruction if `k` is not already in `m`?

----

👍 overall.  Glad to see this improvement.

mlevesquedion:
Thanks for pushing on this, and thanks for your thoughts!

> ... 41 nodes ...

Oops. I was looking at the entire list in `ssa/doc.go`, which includes some non-nodes (actually it looks like I miscounted those as well). There are indeed 41, according to `vim-go`'s `:GoImplements`.

> look into some of the instructions being skipped, like Builtin, FreeVar and maybe Return, but that might be lower-level than you're looking for right now.

A closer look at some/all of the instructions would be very welcome. This stuff can be tricky, so I think some duplication of effort could be good.

For the specific ones that you are mentioning here, here are some of my thoughts:
* `Builtin` is just a `Value`, conceptually it is a reference to one of the `Builtin` functions, so it can't be tainted. If there is a `Call` to a `Builtin`, we will propagate the taint via the `Call`. (I am fairly confident of this, but additional tests would not be unwelcome.)
* `FreeVar`s are handled by `sourcesFromClosure`.
* `Return` can't have `Referrers`, and it is not possible for a value being returned to taint another value that is also being returned via the `Return` instruction.

(Maybe I should add these extra thoughts to the switches? Maybe each instruction should get its own `case` + comment? That might get unwieldy, though.)

> It might help to conceptualize...

I think that would be difficult to do. Sometimes we need to propagate in ways that don't really make sense from a code execution point of view, so I'm not sure we would be able to realize this idea of traversing from `Value` to `Value` via `Instruction`s.

> For the expression m[k] = v, I would expect: ...

Agreed, those are my same thoughts. We should never propagate to anything but the `Map`.

PurelyApplied:
> Builtin is just a Value, conceptually it is a reference to one of the Builtin functions, so it can't be tainted. If there is a Call to a Builtin, we will propagate the taint via the Call. (I am fairly confident of this, but additional tests would not be unwelcome.)

Ahh, yes.  I thought it was one that was both a `Value` and an `Instruction`, and as an `Instruction`, I would expect, say, `copy` to propagate taint.  My mistake.

> Return can't have Referrers, and it is not possible for a value being returned to taint another value via the Return instruction.

Similarly, I thought `Return` had a `Value`, but it looks like it's just the "We're done now" `Instruction`.  The underlying `Return.Value` should already be tainted if the return value should be tainted.

> FreeVars are handled by sourcesFromClosure.

For both `FreeVars` and `Return.Value`s, it was more a comment about completion than consumption.  If we're explicitly relying here on how things are handled elsewhere regarding `FreeVars`, we might want to at least add that comment in this `switch`.  But yeah, not necessary in the current implementation.

mlevesquedion:
> Ahh, yes. I thought it was one that was both a Value and an Instruction, and as an Instruction, I would expect, say, copy to propagate taint. My mistake.

That would have made sense, that was also my expectation before looking at the docs.

> If we're explicitly relying here on how things are handled elsewhere regarding FreeVars, we might want to at least add that comment in this switch.

Done.

PurelyApplied:
I'm not sure how much more work here needs to be done before this graduates from a `Draft` to an "official" PR.

I don't feel like this PR needs to be help up by decisions about how to handle various instructions.  Just getting this switch in place would allow easier iteration for correcting, e.g., how we handle `BinOp` or `Range` instructions.

It's up to you, but I'd be happy with an "in-place, no observable behavior change" refactor PR that allows easy iteration on specific instructions in future PRs.

mlevesquedion:
Yep, I agree! I had the same sentiment and I was about to write a comment expressing it.

I plan to review the switch to make sure that cases where things are a bit ambiguous don't lose any propagation for now, and create an issue for each of those to investigate them further.

I'll round this out and send this for review as soon as I'm ready.

internal/pkg/source/source.go
```diff
@@ -91,88 +91,156 @@ func (s *Source) dfs(n ssa.Node, maxInstrReached map[*ssa.BasicBlock]int, lastBl
...
+			return
+		}
+	}
+
 	mirCopy := map[*ssa.BasicBlock]int{}
 	for m, i := range maxInstrReached {
 		mirCopy[m] = i
 	}
 
 	if instr, ok := n.(ssa.Instruction); ok {
```
mlevesquedion, line 121: I'm doing a second `if` that has the same head as the previous one because I wanted to detangle two things:
1. should we keep traversing?
2. update the max instruction in block, if applicable
These should really be replaced by calls to dedicated functions. I'm happy to do it in this PR, but I could do it in a future PR.

mlevesquedion:
This PR is ready for review. Please read the updated description for a current summary of what this PR represents.

internal/pkg/source/source.go
mlevesquedion: I went ahead and extracted the first if (as well as other related instructions) into its own function. I think this reads much better.

internal/pkg/source/source.go
```diff
@@ -75,104 +75,185 @@ func New(in ssa.Node, config classifier) *Source {
...
 // dfs performs Depth-First-Search on the def-use graph of the input Source.
 // While traversing the graph we also look for potential sanitizers of this Source.
 // If the Source passes through a sanitizer, dfs does not continue through that Node.
-func (s *Source) dfs(n ssa.Node, maxInstrReached map[*ssa.BasicBlock]int, lastBlockVisited *ssa.BasicBlock) {
-	if s.marked[n] {
-		return
-	}
-	// booleans can't meaningfully be tainted
-	if isBoolean(n) {
+func (s *Source) dfs(n ssa.Node, maxInstrReached map[*ssa.BasicBlock]int, lastBlockVisited *ssa.BasicBlock, isReferrer bool) {
```
mlevesquedion, line 85: This function needed to be modified because the check for a call that is a referrer used to be done in `visitReferrers`. Since `Call` is the only special case, it feels cleaner to do it here.

internal/pkg/source/source.go
```diff
@@ -75,104 +75,185 @@ func New(in ssa.Node, config classifier) *Source {
...
+	}
 
-	for _, r := range referrers {
-		switch v := r.(type) {
-		case *ssa.Call:
-			if callee := v.Call.StaticCallee(); callee != nil && s.config.IsSanitizer(utils.DecomposeFunction(callee)) {
-				s.sanitizers = append(s.sanitizers, &sanitizer.Sanitizer{Call: v})
-			}
+	// booleans can't meaningfully be tainted
+	if isBoolean(n) {
```
PurelyApplied, line 119: `string` notwithstanding, do _any_ of the basic types meaningfully carry taint?  Certainly `string` as a convertable type to `[]byte` or `[]rune` could carry taint, but does an individual `byte` or `rune` meaningfully carry taint?

If we're excluding basic types, we should consider excluding all numeric and pointer types.

Conversely, if memory serves, this came from the special handling surrounding `val, ok := SomeExtract`.  If that's now / going to be being properly handled in the big instruction switch, then we shouldn't need to handle that here.

internal/pkg/source/source.go
```diff
@@ -75,104 +75,185 @@ func New(in ssa.Node, config classifier) *Source {
...
+	// These nodes don't have operands; they are Values, not Instructions.
+	case *ssa.Const, *ssa.FreeVar, *ssa.Global, *ssa.Lookup, *ssa.Parameter:
+		s.visitReferrers(n, maxInstrReached, lastBlockVisited)
+
+	// These nodes don't have referrers; they are Instructions, not Values.
+	case *ssa.Go, *ssa.Store:
+		s.visitOperands(n, maxInstrReached, lastBlockVisited)
+
+	// These nodes are both Instructions and Values, and currently have no special restrictions.
+	case *ssa.Field, *ssa.MakeInterface, *ssa.Select, *ssa.TypeAssert:
```
PurelyApplied, line 229: Should `*ssa.Field` have the same special handling that `ssa.FieldAddr` has?

PurelyApplied, line 229: Related to an above comment, but I might expect the special handling of not propagating through the `CommaOK` value of a type assert to be handled here - at least if we weren't otherwise claiming that a boolean can't be tainted.

PurelyApplied (commented):
I'm happy with how this is shaking out.  🎉 

-----

Out of scope for this PR, but it is an overhaul...

`source.RefersTo` is unused and should be pruned.

internal/pkg/source/source.go
```diff
@@ -75,104 +75,185 @@ func New(in ssa.Node, config classifier) *Source {
...
+	// The actual integer Index should not be visited.
+	// Everything but the actual integer Index should be visited.
+	case *ssa.IndexAddr:
+		s.visitReferrers(n, maxInstrReached, lastBlockVisited)
+		s.dfs(t.X.(ssa.Node), maxInstrReached, lastBlockVisited, false)
+
+	// Only the Map itself can be tainted by an Update.
+	// The Key can't be tainted.
+	// The Value can propagate taint to the Map, but not receive it.
+	// MapUpdate has no referrers, it is only an Instruction, not a Value.
```
PurelyApplied, line 205: Certainly out of scope for this PR, possibly out of scope for this analyzer, but do we need to consider where keys are tainted?  E.g.

```golang
func TestFoo() {
	s := core.Source{Data: "password1234"}
	m := map[core.Source]string{s: s.Data}
	for src, str := range m {
		core.Sink(src) // want "a source has reached a sink"
		core.Sink(str) // want "a source has reached a sink"
	}
}
```

internal/pkg/source/source.go
```diff
@@ -75,104 +75,185 @@ func New(in ssa.Node, config classifier) *Source {
...
+		lastBlockVisited = instr.Block()
+	}
 
-	s.visitOperands(n, n.Operands(nil), mirCopy, lastBlockVisited)
+	s.visit(n, mirCopy, lastBlockVisited)
 }
 
-func (s *Source) visitReferrers(n ssa.Node, maxInstrReached map[*ssa.BasicBlock]int, lastBlockVisited *ssa.BasicBlock) {
-	referrers := s.referrersToVisit(n, maxInstrReached)
+func (s *Source) shouldNotVisit(n ssa.Node, maxInstrReached map[*ssa.BasicBlock]int, lastBlockVisited *ssa.BasicBlock, isReferrer bool) bool {
```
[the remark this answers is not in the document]
mlevesquedion: In general I would agree, but in this case the function is only used in the negative case, so if it were named `shouldVisit` we would need to write `!shouldVisit` in the calling code, which is just as readable as `shouldNotVisit`, but feels a bit backwards to me: we're asking if we *should* visit, but we're actually interested in knowing if we *shouldn't* visit, so we immediately flip the result of the call.

internal/pkg/source/source.go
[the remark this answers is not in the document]
mlevesquedion: This is a bit pathological, but if each `byte`/`rune` from a tainted `[]byte` or `[]rune` were to be sinked individually, avoiding `bytes` and `runes` would miss that. (e.g. something like `for _, r := range "secret" { fmt.Printf(r) }`; again, this is kind of pathological). A similar thing could occur with some numeric types, e.g. in theory a cryptographic key could be sinked byte-by-byte.

> we should consider excluding ... pointer types.

I think that could cause us to miss a lot of things. We don't know what a sink is going to do with its argument, e.g. dereference it.

> Conversely, if memory serves, ...

You are correct, but we also wanted to avoid things like `fmt.Println(isSource(Source{}))`, so I think avoiding all booleans is justified.

internal/pkg/source/source.go
```diff
@@ -75,104 +75,185 @@ func New(in ssa.Node, config classifier) *Source {
...
+		}
 
-			if !s.config.IsSourceField(typPath, typName, fieldName) {
-				continue
-			}
+		// This is to avoid attaching calls where the source is the receiver, ex:
+		// core.Sinkf("Source id: %v", wrapper.Source.GetID())
+		if recv := t.Call.Signature().Recv(); recv != nil && s.config.IsSourceType(utils.DecomposeType(utils.Dereference(recv.Type()))) {
+			return
+		}
```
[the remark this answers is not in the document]
mlevesquedion: Good question. This is actually one of the more significant architectural flaws in the codebase right now, IMO.

The reason is that `GetSecret` will be identified as a `fieldpropagator`, which will be found in `levee` when checking for `Calls`, which will start a new traversal with the result of `GetSecret` as a source. `fieldpropagators` should be used in the core traversal logic, not anywhere else. I had a PR for that a long time ago, #77, but it got lost in the weeds. I think I ended up closing it because I wanted the `fieldpropagators` to be consumed by `sourcetype` or something of the sort.

Currently the propagation logic in `fieldpropagator` is not nearly as developed as that in `source`, and in fact I believe that logic should probably be shared: #130.

internal/pkg/source/source.go
mlevesquedion: Hmm. With this case in mind, I think in general we may want to consider *anything* coming out of a tainted map to be tainted. The current tests don't do that, e.g.

```go
func TestRangeOverMapWithSourceAsKey() {
	m := map[core.Source]string{core.Source{Data: "password1234"}: "don't sink me"}
	for src, str := range m {
		core.Sink(src) // want "a source has reached a sink"
		core.Sink(str)
	}
}
```

In the above cases, the fix would be to propagate through the `Next` instruction. In the current PR, we do not propagate through those instructions at all; on master, we do. I think for now it would be best to add that case you mentioned and to keep propagating through `Next` instructions.

(For `MapUpdate` though I'm confident we don't want to traverse to anything other than the `Map`).

internal/pkg/source/source.go
mlevesquedion: I think almost certainly, yes. See #188.

mlevesquedion: We could do that by checking if the `TypeAssert` is `CommaOk`, and if so, avoiding the 1st `Referrer`. I think it's simpler to rely on the code for avoiding booleans.

Looking into this, it looks like the `CommaOk` case for type asserts was not covered, so I added a line to an existing test:
```
func TestSourcePointerAssertedFromParameterEfaceCommaOk(e interface{}) {
	s, ok := e.(*core.Source)
	core.Sink(ok) <-- this is the line I added, no report should be produced here
}
```

PurelyApplied: Sounds good to me.

internal/pkg/source/source.go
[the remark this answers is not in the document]
PurelyApplied: Gotcha.  And yeah, you know opinion on the pipeline of analyzers here.  If these were unified in `levee` instead of propagation happening all in `source`... but that's a conversation for a future PR.

internal/pkg/source/source.go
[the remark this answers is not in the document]
PurelyApplied: For me, it's about reading the function while you're in it.  While I'm looking at this function, I don't necessarily know all of its use-cases.  I also don't know that consumption won't change to invert this into a double negative.  At least, that's where I'm coming from, philosophically.

Then again, Naming is Hard:tm: and it's a bikeshed.  It's fine as it is.

internal/pkg/source/source.go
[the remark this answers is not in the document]
PurelyApplied: > I think that could cause us to miss a lot of things. We don't know what a sink is going to do with a pointer argument, e.g. dereference it.

I was imprecise.  I meant within the context of basic types - specifically, the [`types.BasicKind`](https://godoc.org/go/types#BasicKind)s, which fall into:
* Invalid
* Strings
* Numeric types, including boolean
* _Unsafe_ pointers
* Untyped types
* The `Byte` and `Rune` alias for `Uint8` and `Int32` respectively.

So when I said "pointers," I should have said "unsafe pointers."  I feel like `unsafe`, like `reflect`, is out of scope for contextual analysis for us.

So of the above, I would consider taint as able to propagate to strings (typed and untyped), and optionally runes (typed and untypted) and bytes.

You could make a similar pathological argument that someone could coerce a `string -> []rune -> []int32 -> []int64` and back, but that feels like the sort of thing that is hard to avoid criticism in code review.  \shrug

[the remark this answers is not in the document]
PurelyApplied: Also, this is another thing that falls into "could easily be its own PR."  I was really just stirring conversation while it was front of mind, but we should shift this conversation to in Issue to not block this PR.

internal/pkg/source/source.go
PurelyApplied: (I'm going to use sources in keys and strings in values in this comment.  The same logic applies in reverse.)

I think if we propagate through `Next`, we might end up tainting keys categorically where we shouldn't.  It could be worth a shot and seeing how it plays out though.

But the more I think about it, the more I think this might be out of scope.  It feels like we're needing to decide categorically whether or not both keys and values are tainted if a map contains a source type value.  Otherwise, we would have to make a distinction between the map "safely" holding its values versus having been tainted by receiving a tainted value as a key.

PurelyApplied (approved):
At this point, I think all of my concerns are small and modular enough to be their own future PRs.  Thanks again for looking into this.

internal/pkg/source/source.go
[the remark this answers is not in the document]
mlevesquedion: Agreed. Issue here: https://github.com/google/go-flow-levee/issues/191

internal/pkg/source/source.go
mlevesquedion: SG. I've added it to #188.

PR #179, opened 2020-11-10
Add missing BinOp test

mlevesquedion:
Currently, we are propagating to all `Operands` + `Referrers` of `BinOp`s. This means that if one `Operand` of a `BinOp` is tainted, we will propagate to the other `Operand`. Doing so never makes sense. Indeed, according to the `ssa` documentation for `BinOp`, the operation can be one of the following:

```go
	// ADD SUB MUL QUO REM          + - * / %
	// AND OR XOR SHL SHR AND_NOT   & | ^ << >> &^
	// EQL NEQ LSS LEQ GTR GEQ      == != < <= < >=
```

(To my knowledge) All of these operations produce new values, so taint cannot propagate to the other operand if one of the operands is tainted. It should only propagate to the result.

This PR introduces a failing test that demonstrates the erroneous behavior. This behavior could be fixed via a special case in the current code, or via the explicit handling proposed in #178.

## Additional comments

For this test to work, `prefix` has to be a parameter. Otherwise, `ssa` inlines the string everywhere, which prevents any propagation from happening.

- [x] Tests pass

### DIFF
--- a/internal/pkg/levee/testdata/src/example.com/tests/binop/tests.go
+++ b/internal/pkg/levee/testdata/src/example.com/tests/binop/tests.go
@@ -20,7 +20,7 @@ import (
 	"example.com/core"
 )
 
-func TestConcatenatingTaintedAndNonTainteStrings(prefix string) {
+func TestConcatenatingTaintedAndNonTaintedStrings(prefix string) {
 	s := core.Source{Data: "password1234"}
 	message := fmt.Sprintf("source: %v", s)
 	fullMessage := prefix + message
--- a/internal/pkg/levee/testdata/src/example.com/tests/binop/tests.go
+++ b/internal/pkg/levee/testdata/src/example.com/tests/binop/tests.go
@@ -1,0 +1,31 @@
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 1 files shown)


internal/pkg/levee/testdata/src/example.com/tests/binop/tests.go
```diff
@@ -0,0 +1,31 @@
...
+
+	"example.com/core"
+)
+
+func TestConcatenatingTaintedAndNonTainteStrings(prefix string) {
+	s := core.Source{Data: "password1234"}
+	message := fmt.Sprintf("source: %v", s)
+	fullMessage := prefix + message
+	// TODO: no report should be produced for "prefix"
+	core.Sink(prefix)      // want "a source has reached a sink"
```
vinayakankugoyal, line 28: I thought we were adding the TODO here?

vinayakankugoyal (approved):

internal/pkg/levee/testdata/src/example.com/tests/binop/tests.go
mlevesquedion: Good callout. For `want` tests that is true, I agree that `TODO want` should be preferred, and indeed that is the way all such `TODO`'s are written (to my knowledge). This case however is what I would call a `do not want` test, and for those I usually put the `TODO` on the line above (see e.g. `testdata/src/example.com/tests/sanitization/tests.go`).  I could put it in the same comment as the `want`, but I'm worried it would be less readable that way.

WDYT?

mlevesquedion:
I failed to mention it in the description originally, but for this test to work, `prefix` has to be a parameter. Otherwise, `ssa` inlines it everywhere, which prevents any propagation from happening.

I've updated the description to mention this.

internal/pkg/levee/testdata/src/example.com/tests/binop/tests.go
vinayakankugoyal: That seems reasonable to me. Thanks!

vinayakankugoyal (approved):

PurelyApplied (approved):

PR #180, opened 2020-11-10
Add missing Phi test

mlevesquedion:
This PR is very similar to #179, but it adds a test for a different kind of `Instruction`.

Currently, we are propagating to all `Operands` + `Referrers` of `Phi` nodes. This means that if one Operand of a Phi is tainted, we will propagate to the other Operand(s). Doing so never makes sense. Indeed, a `Phi` node is a decision point where (during execution) the value that should be used is determined by the path that was taken. If one of the possible values is tainted, we cannot conclude that any of the other values is tainted.

This PR introduces a failing test that demonstrates the erroneous behavior. This behavior could be fixed via a special case in the current code, or via the explicit handling proposed in #178.

## Why is the test so convoluted?

Some traversal behaviors, not directly related to `Phi` nodes, are already affording us some amount of protection against incorrect traversals :
* We don't traverse through `Alloc`s, except if they are allocating arrays. This is why the untainted operand is a pointer (pointers don't have Allocs). Since it is a pointer, making it a parameter means it won't be replaced by an explicit `nil` (which would get inlined everywhere, preventing any propagation from occurring).
* We don't traverse to instructions that are in a block that isn't reachable from the current instruction's block. This is why all the instructions are in a `for` loop, it makes every block reachable from every other block.

- [x] Tests pass

### DIFF
--- a/internal/pkg/levee/testdata/src/example.com/tests/phi/tests.go
+++ b/internal/pkg/levee/testdata/src/example.com/tests/phi/tests.go
@@ -32,3 +32,13 @@ func TestPhiNodeDoesntPropagateTaintToOperands(i *core.Innocuous) {
 		core.Sink(ss) // want "a source has reached a sink"
 	}
 }
+
+func TestPhiNodeIsTaintedByTaintedOperand(s core.Source, i *core.Innocuous) {
+	var ss, ii interface{} = s, i
+	if true {
+		ii = ss
+	}
+	core.Sink(ss) // want "a source has reached a sink"
+	core.Sink(ii) // want "a source has reached a sink"
+	core.Sink(i)
+}
--- a/internal/pkg/levee/testdata/src/example.com/tests/phi/tests.go
+++ b/internal/pkg/levee/testdata/src/example.com/tests/phi/tests.go
@@ -18,15 +18,17 @@ import (
 	"example.com/core"
 )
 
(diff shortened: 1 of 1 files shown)


internal/pkg/levee/testdata/src/example.com/tests/phi/tests.go
```diff
@@ -0,0 +1,32 @@
...
+func TestPhiNodeWithTaintedAndNonTaintedOperands(i *core.Innocuous) {
+	for {
+		s := core.Source{}
+		var x, y interface{} = s, i
+		if true {
+			x = y
+		}
+		// TODO: no report should be produced for i
+		core.Sink(i) // want "a source has reached a sink"
+		core.Sink(x) // want "a source has reached a sink"
```
PurelyApplied, line 30: I'd throw in the `core.Sink(y)` that we don't want report on here, and then reduplicate the test for `y=x` phi where we *would* want `core.Sink(y)` to emit.

PurelyApplied (approved):
Looks good to me.  Could flesh out the `phi` tests a little bit.

internal/pkg/levee/testdata/src/example.com/tests/phi/tests.go
mlevesquedion: > I'd throw in the `core.Sink(y)` ...

Good idea. Done.

> ... reduplicate the test for y=x phi where we would want core.Sink(y) to emit ...

Also good idea. Also done.

PR #182, opened 2020-11-11
Allow configuration via regexp or string literals.

PurelyApplied:
* Expand testing of matchers, now that the interface is reduced to basic types.
* Update docs.
* Unspecified elements continue to match vacuously.

- [x] Tests pass

### DIFF
--- a/configuration/README.md
+++ b/configuration/README.md
@@ -18,7 +18,7 @@ You may specify these with a combination of string literals or regexp.
 Use `Package`, `Type`, and `Field` to specify by string literal.
 Use `PackageRE`, `TypeRE`, and `FieldRE` to specify by regexp.
 If neither a literal nor a regexp is provided, a wildcard matcher is assumed.
-Providing both a literal and a regexp matcher is considered misconfiguration and will error.
+Providing both a literal and a regexp matcher is considered a misconfiguration and will error.
 
 ```yaml
 Sources:
@@ -46,7 +46,7 @@ As with source configuration, these may be specified by either a provided string
 Use `Package`, `Receiver`, and `Method` to specify by string literal.
 Use `PackageRE`, `ReceiverRE`, and `MethodRE` to specify by regexp.
 If neither a literal nor a regexp is provided, a wildcard matcher is assumed.
-Providing both a literal and a regexp matcher is considered misconfiguration and will error.
+Providing both a literal and a regexp matcher is considered a misconfiguration and will error.
 
 ```yaml
 Sinks:
--- a/configuration/README.md
+++ b/configuration/README.md
@@ -17,22 +17,14 @@ Sources are identified by package, type, and field names.
 You may specify these with a combination of string literals or regexp.
 Use `Package`, `Type`, and `Field` to specify by string literal.
 Use `PackageRE`, `TypeRE`, and `FieldRE` to specify by regexp.
-You may use a combination as needed.
+If neither a literal nor a regexp is provided, a wildcard matcher is assumed.
+Providing both a literal and a regexp matcher is considered misconfiguration and will error.
 
 ```yaml
 Sources:
-# Specify using string literals
-- Package: "literal/package/path"
-  Type: "typeName"
-  Field: "fieldName"
-# Specify using regexp
-- PackageRE: "<package path regexp>"
-  TypeRE: "<type name regexp>"
-  FieldRE: "<field name regexp>"
-# Specify using string literals and regexp
-- Package: "literal/package/path"
-  TypeRE: ".*"
-  Field: "fieldName"
+- Package: "full/package/path"
+  # Neither Type nor TypeRE are specified - match all type names
+  FieldRE: "Token|Password|Secret" 
 ```
 
 Sources may also be identified via field tags:
@@ -49,37 +41,27 @@ FieldTags:
   Val: source
 ```
 
-Sinks and sanitizers are identified by package, method, and (optional) receiver name.
-Again, these may be specified by either a provided string literal or regexp.
+Sinks and sanitizers are identified by package, method, and (if applicable) receiver name.
+As with source configuration, these may be specified by either a provided string literal or regexp.
+Use `Package`, `Receiver`, and `Method` to specify by string literal.
+Use `PackageRE`, `ReceiverRE`, and `MethodRE` to specify by regexp.
+If neither a literal nor a regexp is provided, a wildcard matcher is assumed.
+Providing both a literal and a regexp matcher is considered misconfiguration and will error.
 
 ```yaml
 Sinks:
-- Package:  "literal/package/path" 
-  ReceiverRE: <type name regexp>
-  MethodRE: <method name regexp>
+- PackageRE:  ".*/sinks(/v1)?"  # Regexp match a collection of packages 
+  # Neither Receiver nor ReceiverRE is provided - match any (or no) receiver
+  Method: "Sink"  # Match only functions named exactly "Sink"
 Sanitizers:
-- PackageRE: <package path regexp>
-  ReceiverRE: <type name regexp>
+- # Neither Package nor PackageRE is provided - match any package
+  ReceiverRE: "Safe
   Method: "mySanitizer"
 ```
 
+To explicitly match an empty string, such as top-level functions without a receiver, explicitly configure an empty string matcher, e.g., `Receiver: ""`.
+
 Taint propagation is performed automatically and does not need to be explicitly configured.
-
-### Important Note
-
-Configuration of the above matchers does not require listing of all arguments.
-E.g., the following is a valid source configuration:
-```yaml
-Sources:
-- Package: "literal/package/path"
-  Type: "MySourceType"
-```
-
-Neither `Field` nor `FieldRE` have been provided.
-In this case, we match all fields of `MySourceType`, assuming a wildcard matcher for `FieldRE`.
-Similar behavior exists in for all attributes, e.g. providing neither `Type` nor `TypeRE` will match all type names.
-
-To explicitly match an empty string, such as top-level functions without a receiver, explicitly define the attribute by the corresponding string literal, `Receiver: ""`, or an anchored regexp, `ReceiverRE: "^$"`.
 
 ### Restricting analysis scope
 
@@ -92,7 +74,7 @@ Exclude:
 ```
 
 The above will match the function beginning with "my" in the `myproject/mypackage` package.
-Since no receiver matcher was provided, it will match any method beginning with "my" bound to any receiver.
+Since no receiver matcher was provided, it will match any method beginning with "my" bound to any (or no) receiver.
 
 As just two examples, this may be used to avoid analyzing test code, or to suppress "false positive" reports.
 
--- a/configuration/README.md
+++ b/configuration/README.md
@@ -13,14 +13,26 @@ For design details concerning value and instruction classification, see [/design
 
 Configuration is provided to `go-flow-levee` via YAML.
 
-Objects of interest are identified primarily via regexp. An empty regexp will match any string.
+Sources are identified by package, type, and field names.
+You may specify these with a combination of string literals or regexp.
+Use `Package`, `Type`, and `Field` to specify by string literal.
+Use `PackageRE`, `TypeRE`, and `FieldRE` to specify by regexp.
+You may use a combination as needed.
 
-Sources are identified via regexp according to package, type, and field names.
 ```yaml
 Sources:
+# Specify using string literals
+- Package: "literal/package/path"
+  Type: "typeName"
+  Field: "fieldName"
+# Specify using regexp
 - PackageRE: "<package path regexp>"
   TypeRE: "<type name regexp>"
   FieldRE: "<field name regexp>"
+# Specify using string literals and regexp
+- Package: "literal/package/path"
+  TypeRE: ".*"
+  Field: "fieldName"
 ```
 
 Sources may also be identified via field tags:
@@ -37,35 +49,50 @@ FieldTags:
   Val: source
 ```
 
-Sinks and sanitizers are identified via regexp according to package, method, and (optional) receiver name.
+Sinks and sanitizers are identified by package, method, and (optional) receiver name.
+Again, these may be specified by either a provided string literal or regexp.
 
 ```yaml
 Sinks:
-- PackageRE: <package path regexp>
+- Package:  "literal/package/path" 
   ReceiverRE: <type name regexp>
   MethodRE: <method name regexp>
 Sanitizers:
 - PackageRE: <package path regexp>
   ReceiverRE: <type name regexp>
-  MethodRE: <method name regexp>
+  Method: "mySanitizer"
 ```
 
 Taint propagation is performed automatically and does not need to be explicitly configured.
 
-For matchers that accept a `ReceiverRE` regexp matcher, an unspecified string will match any (or no) receiver.
-To match only methods without any receiver (i.e., a top-level function), use the matcher `^$` to match an empty-string receiver name.
+### Important Note
+
+Configuration of the above matchers does not require listing of all arguments.
+E.g., the following is a valid source configuration:
+```yaml
+Sources:
+- Package: "literal/package/path"
+  Type: "MySourceType"
+```
+
+Neither `Field` nor `FieldRE` have been provided.
+In this case, we match all fields of `MySourceType`, assuming a wildcard matcher for `FieldRE`.
+Similar behavior exists in for all attributes, e.g. providing neither `Type` nor `TypeRE` will match all type names.
+
+To explicitly match an empty string, such as top-level functions without a receiver, explicitly define the attribute by the corresponding string literal, `Receiver: ""`, or an anchored regexp, `ReceiverRE: "^$"`.
 
 ### Restricting analysis scope
 
-Functions can be explicitly excluded from analysis using regexps,
+Functions can be explicitly excluded from analysis using string literals or regexps,
 constructed similarly to those used to identify sanitizers and sinks:
 ```yaml
 Exclude:
-- PathRE: "^myproject/mypackage$"
-  MethodRE: "^myfunction$"
+- Package: "myproject/mypackage"
+  MethodRE: "^my.*"
 ```
 
-The above will match the function `myfunction` from the `myproject/mypackage` package. It will also match a method named `myfunction` in the same package.
+The above will match the function beginning with "my" in the `myproject/mypackage` package.
+Since no receiver matcher was provided, it will match any method beginning with "my" bound to any receiver.
 
 As just two examples, this may be used to avoid analyzing test code, or to suppress "false positive" reports.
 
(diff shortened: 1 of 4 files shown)


internal/pkg/config/config.go
```diff
@@ -120,31 +121,95 @@ func (c Config) IsSourceField(path, typeName, fieldName string) bool {
...
-	return fm.PackageRE.MatchString(path) && fm.ReceiverRE.MatchString(receiver) && fm.MethodRE.MatchString(name)
+	return matchEither(fm.Package, fm.PackageRE, path) &&
+		matchEither(fm.Receiver, fm.ReceiverRE, receiver) &&
+		matchEither(fm.Method, fm.MethodRE, name)
+}
+
+// TODO This is a terrible name.  matchAnyOrNil is not better.
+// Matches match against a string literal or regexp.
+// Returns vacuous true when both matchers are nil.
+func matchEither(literal *string, r *regexp.Regexp, match string) bool {
```
PurelyApplied, line 211: I'm bad at naming, and then once I give a bad name, it's hard for me to shake it out of my head.

The `nil` handling is to be consistent with existing behavior.  If we want to say a user must specify at least one of the literal or the regexp for a given matcher attribute, we can, but that had wide-reaching impact on our own test configs.  That can be part of a future PR if you think that requirement would be good.

internal/pkg/config/matcher_test.go
```diff
@@ -0,0 +1,244 @@
...
+		shouldMatch       bool
+	}{
+		{
+			desc: "Garbage in garbage out",
+			yaml: `
+Blahblah: foo
+PackageRE: bar`,
+			// TODO Discuss loss of strict loading
+			shouldErrorOnLoad: false, // TODO true
+		},
```
PurelyApplied, line 23: I have discovered an interesting side-effect of using the K8s `yaml` package.

To massage some other pain points, that `yaml` package converts the YAML is it unmarshaling into JSON, and uses `json.Unmarshal` to unpack it.  However, the `json` package doesn't have a "strict" mode.  So when providing our own `UnmarshalJSON` methods, there is no strict mode inherited from our original `yaml.Unmarshal` call.

I can make the implementation of `UnmarshalJSON` inherently strict, but (I think) it requires unpacking into `json.RawMessage` and examining keys explicitly.  I'm not familiar with that at present.  In either event, I think it is a complex enough of an issue that it could be its own PR, if a temporary regression is acceptable.

configuration/README.md
```diff
@@ -13,14 +13,26 @@ For design details concerning value and instruction classification, see [/design
 
 Configuration is provided to `go-flow-levee` via YAML.
 
-Objects of interest are identified primarily via regexp. An empty regexp will match any string.
+Sources are identified by package, type, and field names.
+You may specify these with a combination of string literals or regexp.
+Use `Package`, `Type`, and `Field` to specify by string literal.
+Use `PackageRE`, `TypeRE`, and `FieldRE` to specify by regexp.
+You may use a combination as needed.
```
mlevesquedion, line 20: This seems redundant with the sentence 3 lines above.

configuration/README.md
```diff
@@ -13,14 +13,26 @@ For design details concerning value and instruction classification, see [/design
...
+  Field: "fieldName"
+# Specify using regexp
 - PackageRE: "<package path regexp>"
   TypeRE: "<type name regexp>"
   FieldRE: "<field name regexp>"
+# Specify using string literals and regexp
+- Package: "literal/package/path"
+  TypeRE: ".*"
+  Field: "fieldName"
 ```
```
mlevesquedion, line 36: This feels rather repetitive to me, I'm worried that the signal may be lost in the noise. Maybe giving just 1 concrete example and explaining it would be more helpful? E.g. you could have the type be a regex, the field be a string, and not have a package (or some other mix). Then you could have a few lines explaining what each of those mean.

configuration/README.md
```diff
@@ -37,35 +49,50 @@ FieldTags:
   Val: source
 ```
 
-Sinks and sanitizers are identified via regexp according to package, method, and (optional) receiver name.
+Sinks and sanitizers are identified by package, method, and (optional) receiver name.
+Again, these may be specified by either a provided string literal or regexp.
```
mlevesquedion, line 53: For some readers it may not be clear what `Again` is referring to. I think it would be helpful to be more precise, e.g. `Like for Sources, ` or some other formulation referring to the previous section explicitly.

configuration/README.md
```diff
@@ -37,35 +49,50 @@ FieldTags:
...
-- PackageRE: <package path regexp>
+- Package:  "literal/package/path" 
   ReceiverRE: <type name regexp>
   MethodRE: <method name regexp>
 Sanitizers:
 - PackageRE: <package path regexp>
   ReceiverRE: <type name regexp>
-  MethodRE: <method name regexp>
+  Method: "mySanitizer"
 ```
```
mlevesquedion, line 64: Same suggestion as above, I think just 1 example mixing string literals and regexp, along with an explanation of what the example means, would be more helpful.

configuration/README.md
```diff
@@ -13,14 +13,26 @@ For design details concerning value and instruction classification, see [/design
 
 Configuration is provided to `go-flow-levee` via YAML.
 
-Objects of interest are identified primarily via regexp. An empty regexp will match any string.
+Sources are identified by package, type, and field names.
+You may specify these with a combination of string literals or regexp.
```
mlevesquedion, line 17: What if both "versions" of a matcher are given, e.g. a `Source` has both a `Package` and a `PackageRE`?

(I saw your code that produces an error if this occurs, but I think it could be good to mention it here as well.)

((Currently I think users may not even see the error string if they are running with `go vet`. We should consider printing the error when loading the config in `fieldpropagators` and `source`, or finding a way to have `go vet` itself output the error. This is out of scope for the current PR.))

configuration/README.md
```diff
@@ -37,35 +49,50 @@ FieldTags:
...
+  Method: "mySanitizer"
 ```
 
 Taint propagation is performed automatically and does not need to be explicitly configured.
 
-For matchers that accept a `ReceiverRE` regexp matcher, an unspecified string will match any (or no) receiver.
-To match only methods without any receiver (i.e., a top-level function), use the matcher `^$` to match an empty-string receiver name.
+### Important Note
+
+Configuration of the above matchers does not require listing of all arguments.
```
mlevesquedion, line 70: I think it would be clearer to put the emphasis on the behavior: if a key is not specified, all values for that key will be matched. (As a consequence of that, you may not need to list some keys).

I think in the context of YAML, "key" or "key-value pair" may be clearer than "argument". WDYT? :thinking:

configuration/README.md
```diff
@@ -37,35 +49,50 @@ FieldTags:
...
+E.g., the following is a valid source configuration:
+```yaml
+Sources:
+- Package: "literal/package/path"
```
mlevesquedion, line 74: Nit: I think "full/package/path" may be clearer.

configuration/README.md
```diff
@@ -37,35 +49,50 @@ FieldTags:
...
+  Type: "MySourceType"
+```
+
+Neither `Field` nor `FieldRE` have been provided.
+In this case, we match all fields of `MySourceType`, assuming a wildcard matcher for `FieldRE`.
```
mlevesquedion, line 79: Nit: I think the part after the comma is unnecessary and potentially confusing. It might help to expand a little, e.g. : `If you want to be more explicit, you may use an empty regex such as FieldRE: "". An empty regex will match anything.`

configuration/README.md
```diff
@@ -37,35 +49,50 @@ FieldTags:
...
 ```yaml
 Exclude:
-- PathRE: "^myproject/mypackage$"
-  MethodRE: "^myfunction$"
+- Package: "myproject/mypackage"
+  MethodRE: "^my.*"
 ```
 
-The above will match the function `myfunction` from the `myproject/mypackage` package. It will also match a method named `myfunction` in the same package.
+The above will match the function beginning with "my" in the `myproject/mypackage` package.
```
mlevesquedion, line 94: I think `the function` should be `functions`, since this is a regexp matcher.

configuration/README.md
```diff
@@ -37,35 +49,50 @@ FieldTags:
...
+Similar behavior exists in for all attributes, e.g. providing neither `Type` nor `TypeRE` will match all type names.
+
+To explicitly match an empty string, such as top-level functions without a receiver, explicitly define the attribute by the corresponding string literal, `Receiver: ""`, or an anchored regexp, `ReceiverRE: "^$"`.
 
 ### Restricting analysis scope
 
-Functions can be explicitly excluded from analysis using regexps,
+Functions can be explicitly excluded from analysis using string literals or regexps,
 constructed similarly to those used to identify sanitizers and sinks:
 ```yaml
```
mlevesquedion, line 88: I think this example + the associated explanations are great. I think the examples for Sources and Sinks/Sanitizers should be like this.

internal/pkg/config/config.go
mlevesquedion: Naming is hard.

Why not implement the matching for strings in terms of the matching for regexps? If you anchor the string on both ends you get a string matcher.

internal/pkg/config/config.go
```diff
@@ -120,31 +121,95 @@ func (c Config) IsSourceField(path, typeName, fieldName string) bool {
...
+		return fmt.Errorf("expected only one of Package, PackageRE to be configured")
+	}
+	if raw.Type != nil && raw.TypeRE != nil {
+		return fmt.Errorf("expected only one of Type, TypeRE to be configured")
+	}
+	if raw.Field != nil && raw.FieldRE != nil {
+		return fmt.Errorf("expected only one of Field, FieldRE to be configured")
+	}
+
+	// Copy all fields from raw
```
mlevesquedion, line 156: Why not unmarshal to `s` directly?

mlevesquedion (commented):
I think this is going in the right direction. I haven't had time to look at the tests yet, will do so shortly.

internal/pkg/config/matcher_test.go
mlevesquedion: Thanks for the detailed investigation.

Is `UnmarshalJSON` the only way to do custom unmarshalling? Could you instead add `yaml:...` field tags to the matcher structs and perform the validation after the yaml has been loaded?

I think this wouldn't be too bad of a regression, as long as it is recovered in a timely manner. (Maybe consider creating an issue for visibility.)

internal/pkg/config/matcher_test.go
```diff
@@ -0,0 +1,244 @@
+package config
+
+import (
+	"testing"
+
+	"sigs.k8s.io/yaml"
+)
+
+func TestFuncMatcher(t *testing.T) {
```
mlevesquedion, line 9: The first 4 cases seem to be testing the loading, while the other cases are testing the matching. I think those should be separate tests.

internal/pkg/config/config.go
```diff
@@ -120,31 +121,95 @@ func (c Config) IsSourceField(path, typeName, fieldName string) bool {
...
+	return matchEither(fm.Package, fm.PackageRE, path) &&
+		matchEither(fm.Receiver, fm.ReceiverRE, receiver) &&
+		matchEither(fm.Method, fm.MethodRE, name)
+}
+
+// TODO This is a terrible name.  matchAnyOrNil is not better.
+// Matches match against a string literal or regexp.
+// Returns vacuous true when both matchers are nil.
+func matchEither(literal *string, r *regexp.Regexp, match string) bool {
+	return literal == nil && r == nil || literal != nil && *literal == match || r != nil && r.MatchString(match)
```
mlevesquedion, line 212: This feels a bit heavy. A switch may be easier to read. A switch would also emphasize that these cases are disjoint and that the cases are not ordered (because `||` short-circuits, writing the conditional in this way suggests that there is an order).

internal/pkg/config/matcher_test.go
```diff
@@ -0,0 +1,244 @@
+package config
```
mlevesquedion, line 1: I think you've got good test cases in general, but the tests are also fairly high level, and as such I'm not confident that every case is handled. It also feels like there's a good amount of overlap between cases: all of these tests end up exercising `matchEither` indirectly, and that's where the tricky logic is. I think it would be good to have tests for `matchEither` specifically.

internal/pkg/config/config.go
```diff
@@ -121,21 +121,53 @@ func (c Config) IsSourceField(path, typeName, fieldName string) bool {
...
+
+type literalMatcher string
+
+func (lm literalMatcher) MatchString(s string) bool {
+	return string(lm) == s
+}
+
+type vacuousMatcher struct{}
+
+func (vacuousMatcher) MatchString(s string) bool {
```
mlevesquedion, line 136: Nit: consider removing the parameter name since it is not used.

internal/pkg/config/matcher_test.go
PurelyApplied: > Is UnmarshalJSON the only way to do custom unmarshalling? 

Kind of.  It is the ultimate unmarshaler for the `sigs.k8s.io/yaml` package.

> Could you instead add yaml:... field tags to the matcher structs and perform the validation after the yaml has been loaded?

Since the `sigs.k8s.io/yaml` package ultimately delegates to `json`, then the information would need to live in (non-existent) `json` tags.  Alternatively, we could switch back to the `gopkg.in/yaml` package.

The `gopkg.in` version has its pros and cons.  I went with the `sigs` version because it prevented us needing to tag every field with the corresponding value for config.  The `gopkg.in` version is very brittle to changes in capitalization, as we have in our `PackageRE`.  It expects keys to be strictly-lowercase, where the `sigs` version ends up with more case-agnostic unmarshaling.  But the `gopkg.in` version relies on a `UnmarshalYAML` interface, which provides a little bit more ease of use and would integrate more naturally with strict loading.

Migrading to the `gopkg.in` version would be another path forward.  But that is still a significant enough effort that would warrant its own PR.

internal/pkg/config/config.go
PurelyApplied: From an side conversation, removed this after moving to a polymorphic underlying implementation.

internal/pkg/config/matcher_test.go
mlevesquedion: SG. I don't have a strong preference. I think it's fine to do this in a second PR.

internal/pkg/config/regexp/regexp.go
```diff
@@ -28,7 +28,7 @@ type Regexp struct {
 
 // MatchString delegates matching to the regex package.
 func (mr *Regexp) MatchString(s string) bool {
-	return mr.r == nil || mr.r.MatchString(s)
+	return mr == nil || mr.r == nil || mr.r.MatchString(s)
```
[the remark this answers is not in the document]
mlevesquedion: Is this just for safety? I don't think we need this.

mlevesquedion (commented):
I think this new polymorphic matching looks very good! Much easier to understand. Testing each type of matching should also be easy now.

internal/pkg/config/config.go
mlevesquedion: (Actually switches are also ordered and also short-circuit. This comment is outdated but I thought I should still correct that.)

internal/pkg/config/regexp/regexp.go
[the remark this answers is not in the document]
PurelyApplied: It was a consequence of changing fields from `regexp.Regexp` to `*regexp.Regexp`.  Because nil values can still have interface data, it is possible to call `(*Regexp)(nil).MatchString`, which blew the stack here.

Might not be necessary now that the matchers are polymorphic... will investigate.

internal/pkg/config/matcher_test.go
PurelyApplied: Low level unit tests added for the various matcher types.

internal/pkg/config/matcher_test.go
PurelyApplied: Split out and low-level tests added.

internal/pkg/config/regexp/regexp.go
```diff
@@ -26,6 +26,10 @@ type Regexp struct {
 	r *regexp.Regexp
 }
 
+func New(s string) *Regexp {
+	return &Regexp{r: regexp.MustCompile(s)}
```
mlevesquedion, line 30: Out of scope, but we should consider returning an error if the regexp fails to compile and exposing that to the user in a controlled way, rather than letting it panic.

mlevesquedion: Because as it is, I think this is likely to produce an error that will be difficult for a user to debug. It would be good to be able to say: "hey, while I was trying to unmarshal this `PackageRE` field in your yaml on line X, I ran into this wonky regex that can't be compiled. please fix" (this may be tricky to do, but I think we should move in that general direction).

PurelyApplied: Actually, that's a good point and easy enough to add to this PR.  Done.

internal/pkg/config/matcher_test.go
```diff
@@ -0,0 +1,369 @@
...
+	testCases := []struct {
+		desc, yaml        string
+		shouldErrorOnLoad bool
+	}{
+		{
+			desc: "Unmarshaling is strict",
+			yaml: `
+Blahblah: foo
+PackageRE: bar`,
+			shouldErrorOnLoad: false, // TODO true
```
mlevesquedion, line 34: You are likely already aware of this, but I want to bring attention to the fact that if it weren't for this case, `shouldErrorOnLoad` would be `true` in all cases. Maybe this test should be named differently to reflect that it is testing unmarshaling failure cases (and the boolean removed).

internal/pkg/config/matcher_test.go
```diff
@@ -0,0 +1,369 @@
...
+				t.Errorf("MatchType(%q, %q) = %v; got %v", tc.path, tc.typ, tc.shouldMatchType, !tc.shouldMatchType)
+			}
+			if tc.shouldMatchField != sm.MatchField(tc.path, tc.typ, tc.fieldName) {
+				t.Errorf("MatchField(%q, %q, %q) = %v; got %v", tc.path, tc.typ, tc.fieldName, tc.shouldMatchType, !tc.shouldMatchType)
+			}
+		})
+	}
+}
+
+func TestMatcherTypes(t *testing.T) {
```
mlevesquedion, line 299: Nit: I think each matcher type should have its own test, but I don't think it's worth changing.

internal/pkg/config/matcher_test.go
```diff
@@ -0,0 +1,369 @@
...
+		},
+		{
+			desc:        "literal matcher foo != bar",
+			matcher:     literalMatcher("foo"),
+			s:           "bar",
+			shouldMatch: false,
+		},
+		{
+			desc:        "regexp matcher /foo/ matches foo",
+			matcher:     func() stringMatcher { r, _ := regexp.New("foo"); return r }(),
```
mlevesquedion, line 326: Nice

mlevesquedion (approved):
LGTM

internal/pkg/config/matcher_test.go
PurelyApplied: I'll rename / refactor in the PR that resolves this case.  Otherwise, the test case would need to be commented out or dropped to avoid breaking Travis.

So as much as I hate TODOs, I think it might be the least of the evils for our tracking of suppressed tests.

PR #183, opened 2020-11-11
Add nil-check for safety on interface methods.

PurelyApplied:
`master` is currently dereferencing a `nil` when running against Kubernetes.  This was introduced by #164, either by awkwardly merging with other commits or, more likely, I failed to re-verify a Kubernetes run after responding to requested changes.

- [x] Tests pass

### DIFF
--- a/internal/pkg/fieldpropagator/analyzer.go
+++ b/internal/pkg/fieldpropagator/analyzer.go
@@ -74,8 +74,9 @@ func run(pass *analysis.Pass) (interface{}, error) {
 		}
 		methods := ssaProg.MethodSets.MethodSet(ssaType.Type())
 		for i := 0; i < methods.Len(); i++ {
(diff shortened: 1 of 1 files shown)


mlevesquedion (approved):

mlevesquedion:
It doesn't look to me like this was introduced by #164? I looked at the file history and it looks like there was no nil check from the start, and your changes in #164 were unrelated to this.

In any case, good catch. I wasn't expecting that this could be nil when I wrote it, but now it seems obvious. It's even mentioned in the docs for `MethodValue`:

```go
// ... It returns nil if sel denotes an
// abstract (interface) method.
```

PurelyApplied:
My gut was that 164 caused the immediately prior check to no longer trigger the `continue`, but \shrug.

mlevesquedion:
Hmm. Maybe. I thought those were equivalent? :thinking:

PR #185, opened 2020-11-12
Only traverse to reference args

mlevesquedion:
This PR modifies the propagation for `Call`s so that only reference arguments (as determined by `pointer.CanPoint`) are tainted. It makes no sense to taint non-reference arguments, because the called function will receive a copy of them, so the original values in the caller can't be tainted.

New tests are added to verify this behavior.

- [x] Tests pass


internal/pkg/source/source.go
```diff
@@ -226,7 +229,7 @@ func (s *Source) visit(n ssa.Node, maxInstrReached map[*ssa.BasicBlock]int, last
 		s.visitOperands(n, maxInstrReached, lastBlockVisited)
 
 	// These nodes are both Instructions and Values, and currently have no special restrictions.
-	case *ssa.Field, *ssa.MakeInterface, *ssa.Select, *ssa.TypeAssert:
+	case *ssa.Field, *ssa.MakeInterface, *ssa.Select, *ssa.Slice, *ssa.TypeAssert, *ssa.UnOp:
```
mlevesquedion, line 232: These are both needed for the tests in this PR.

internal/pkg/levee/testdata/src/example.com/tests/collections/maps.go
```diff
@@ -53,7 +53,9 @@ func TestMapRemainsTaintedWhenSourceIsDeleted(s core.Source) {
 	core.Sink(m) // want "a source has reached a sink"
 }
 
-func TestDeletingFromTaintedMapDoesNotTaintTheKey(key string, sources map[string]core.Source) {
+func TestDeletingFromTaintedMapDoesNotTaintKey(key *string, sources map[*string]core.Source) {
```
mlevesquedion, line 56: The modifications in this PR caused this test to pass spuriously. The key now needs to be some kind of pointer in order to demonstrate the incorrect behavior.

internal/pkg/source/source.go
```diff
@@ -462,6 +465,52 @@ func isSourceType(c classifier, t types.Type) bool {
...
+
+	case *types.Basic:
+		return tt.Kind() == types.UnsafePointer
+
+	// These types cannot point.
+	// Signatures represent functions, which for our purpose aren't pointers.
+	// Tuples only occur within functions.
+	case *types.Signature, *types.Tuple:
+		return false
+
```
PurelyApplied, line 486: This feels out of scope for both this function and this PR.  `Signature`s are pointer-like, and I would expect a method named `canPoint` to reflect that.  Whether or not they can be meaningfully tainted is another question, and one that feels like it should be owned by `Source.visit`.

internal/pkg/levee/testdata/src/example.com/tests/colocation/tests.go
```diff
@@ -16,22 +16,138 @@ package colocation
...
+
+func TestTaintedThroughEface(s core.Source, str string, i core.Innocuous) {
+	colocateEface(s, str, i)
+	// Ideally, we wouldn't want reports for either of those values, because they
+	// are not pointers.
+	// However, because they are passed to the call via an interface type,
+	// we have no easy way to know that these values can't actually be tainted.
+	core.Sink(str) // want "a source has reached a sink"
+	core.Sink(i)   // TODO want "a source has reached a sink"
+}
```
PurelyApplied, line 143: Delegating to `pointer.CanPoint` and reverting your changes in handling `UnOp` and `Slice` yields this test and the below as passing.

internal/pkg/source/source.go
PurelyApplied: This might be a good indication that the PR has gotten a bit bigger in scope than intended.

PurelyApplied (changes requested):
This PR feels like it's doing a bit too much.

* Like `reflect`, I don't think we have or should make any promise about tracking taint across use of `unsafe`.  At best, that feels out of scope for this PR.
* For what is pointer-like, I think `pointer` should be authoritative.  Our use case includes also non-pointer wrappers of pointer-like types, so we have two extra cases to check.  That delineation could be more clear.

internal/pkg/source/source.go
PurelyApplied: Poking at the test cases more, I see now that this is necessary for propagating in `taint(&myString, someSource)`.

internal/pkg/source/source.go
```diff
@@ -462,6 +465,52 @@ func isSourceType(c classifier, t types.Type) bool {
 	}
 }
 
+// A type "can point" if it is itself a pointer or pointer-like type, or it contains
+// a pointer or pointer-like type.
+func canPoint(t types.Type) bool {
```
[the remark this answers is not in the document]
mlevesquedion: I think this is a good point. Done. In addition, I changed the function name to `canBeTaintedByCall`, because that's really what the client code is interested in.

internal/pkg/source/source.go
mlevesquedion: That does sound like a better separation of work. I'll address `Signatures` and `Tuples` in a future PR, then.

internal/pkg/levee/testdata/src/example.com/tests/colocation/tests.go
mlevesquedion: Hmm. That's not what I'm seeing. If I revert the change to `UnOp`, the `TestArrOfPointersIsTainted` test case fails. If I revert the change to `Slice`, `TestTaintedEface` and the first want in `TestTaintedThroughEface` both fail. Neither change seems to have any impact on the `TestPointerTaintedThroughEface` case, which is below `TestTaintedThroughEface`.

mlevesquedion:
Thank you for your review!

> Like reflect, I don't think we have or should make any promise about tracking taint across use of unsafe.

For `reflect.Value`, we get that behavior via `pointer.CanPoint`. I think tracking taint through `reflect.Value` is legitimate, since those can be freely converted to and from `interface{}` values. Whether we want to model reflection is a different question and I think we agree that the answer is currently "no". My reasoning was similar in the case of `unsafe.Pointer`, I thought since those can be converted to and from pointers, then they should be considered pointers as well, but I think the reality is a bit more subtle, so I agree that this doesn't need to be in this PR.

internal/pkg/source/source.go
mlevesquedion: Hmm. That's not what I'm seeing. Should we consolidate this discussion with the discussion in this thread: https://github.com/google/go-flow-levee/pull/185/files#r525629809

internal/pkg/levee/testdata/src/example.com/tests/colocation/tests.go
```diff
@@ -16,22 +16,118 @@ package colocation
...
+	colocateReflectValue(s, r)
+	core.Sink(r) // TODO want "a source has reached a sink"
 }
 
-func TestTaintIsPropagatedToColocatedPointerArgumentsThroughEface(s core.Source, i core.Innocuous, ip *core.Innocuous) {
-	taintColocatedEface(s, &i, ip)
-	core.Sink(s)  // want "a source has reached a sink"
-	core.Sink(i)  // TODO want "a source has reached a sink"
-	core.Sink(ip) // want "a source has reached a sink"
+func colocateSingleEface(s core.Source, e interface{})             {}
```
PurelyApplied, line 112: Does `Eface` have a conventional?  Searching around for it, I find it used as the `runtime` unexported type to hold interfaces, which fits here.  I just wanted to be on the same page, and might consider using `Interface` to avoid any confusion at the low cost of four characters.

internal/pkg/source/source.go
```diff
@@ -466,6 +470,30 @@ func isSourceType(c classifier, t types.Type) bool {
...
+
+	switch tt := t.(type) {
+	case *types.Array:
+		return canBeTaintedByCall(tt.Elem())
+
+	case *types.Struct:
+		for i := 0; i < tt.NumFields(); i++ {
+			if canBeTaintedByCall(tt.Field(i).Type()) {
+				return true
+			}
```
PurelyApplied, line 489: My first impression was that this could potentially be infinitely recursive.

Mulling it over, you aren't actually able to have a non-reference recursive struct, because you'd get a build failure.  You'd only be able to recurse if a field of `type Foo struct { ...` were, e.g., `*Foo` or `[]Foo`, but not `Foo` or `[2]Foo`.

All that to say, it might be worth a comment here to explain why we aren't actually worried about the recursive step here.

internal/pkg/levee/testdata/src/example.com/tests/colocation/tests.go
PurelyApplied: Yeah, I think two-weeks-ago-Me got into a weird partial state.

PurelyApplied (approved):
Looks good!  Only two minor question/comment items.

PR #189, opened 2020-11-13
Visit every node when debugging

mlevesquedion:
This PR fixes an issue where some debug graphs were being generated with edges missing. In some cases, almost no edges were being generated, so the resulting DOT graph was actually just a collection of isolated instruction nodes.

The crux of the issue was that the algorithm that traversed the SSA graph made no attempt to visit every node. It only started a traversal with the first instruction of the first block, which only works if the graph is connected. In our tests, most of the time, the graph is connected, so I only ran into this issue a handful of times despite using the graphs every time I have to debug something. When the issue was brought to my attention by @vinayakankugoyal, I decided it was time to do something about it.

To fix the issue, I completely removed the `debug/graph` package (which wasn't needed) and rewrote the rendering code to visit each instruction and add its referrers/operands.

I've added some tests with golden files, like we are currently using for the SSA rendering tests.

## Future improvements

A few ideas for future improvements:
* Add block numbers in the DOT output
* Also render closures

I think both of those would be nice to have, but not pressing at the moment. @vinayakankugoyal I think the second one could be useful to the work you're currently doing, feel free to tackle it.

- [x] Tests pass


internal/pkg/debug/render/dot.go
```diff
@@ -46,53 +45,59 @@ func (r *renderer) Render() string {
 }
 
 func (r *renderer) init() {
-	_, _ = r.WriteString("digraph {\n")
```
mlevesquedion, line 49: I have removed this here and elsewhere. It cannot actually fail, since we are writing to a `strings.Builder`, so explicitly ignoring the `int, error` return values is just noise that makes the code harder to read.

internal/pkg/debug/render/dot.go
```diff
@@ -18,23 +18,22 @@ import (
...
 
 // DOT produces DOT source code representing the SSA graph for a function.
 func DOT(f *ssa.Function) string {
-	return renderDOT(graph.New(f))
+	return renderDOT(f)
 }
 
-func renderDOT(g *graph.FuncGraph) string {
-	return (&renderer{strings.Builder{}, g}).Render()
+func renderDOT(f *ssa.Function) string {
```
benhxy, line 30: Maybe can merge func DOT and func renderDOT?

internal/pkg/debug/render/render_test.go
```diff
@@ -30,24 +30,33 @@ import (
...
 
+func TestDOT(t *testing.T) {
+	testGolden(t, DOT, "DOT", ".dot")
+}
+
 func TestSSA(t *testing.T) {
+	testGolden(t, SSA, "SSA", ".ssa")
+}
+
+func testGolden(t *testing.T, fn func(f *ssa.Function) string, fnName, ext string) {
```
benhxy, line 41: What is "golden". Maybe add some comment?

mlevesquedion: A "golden" file is a file that contains the expected output for a test. I've seen/used that term many times, so I assumed just "golden" would be clear enough. After doing some digging though, it seems that this term is actually [not that common](https://softwareengineering.stackexchange.com/questions/358786/what-is-golden-files).

Renamed to `testGoldenFiles` and added a comment.

internal/pkg/debug/render/dot.go
```diff
@@ -18,23 +18,22 @@ import (
 	"fmt"
 	"strings"
 
-	"github.com/google/go-flow-levee/internal/pkg/debug/graph"
 	"github.com/google/go-flow-levee/internal/pkg/debug/node"
 	"golang.org/x/tools/go/ssa"
 )
 
 // DOT produces DOT source code representing the SSA graph for a function.
 func DOT(f *ssa.Function) string {
```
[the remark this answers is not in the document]
mlevesquedion: Ah, good call. There's no reason for these to be separate functions.

benhxy (approved):
LGTM. Please squash the commits before merging.

PR #193, opened 2020-11-17
Add checkconfig command, backed by funccalls and funcdefs (new) analyzers

mlevesquedion:
This PR introduces a `checkconfig` command that can be used to check what the given configuration matches. This command relies on a new `funcdefs` analyzer, as well as the `funccalls` analyzer introduced in #196.

Presently, the command only covers Sinks and Sanitizers. In a future PR, the `checkconfig` analyzer will consume the `sourceinfer` analyzer, so that it can also report what source types it is finding, as well as why they were determined to be sources, e.g. configured or inferred. At that point, I will remove the `sourcetype` command and analyzer.

## Caveat

Currently, this command fails to find the definitions for `klog` sinks when run from anywhere but the `kubernetes/vendor/k8s.io/klog/v2` directory. I believe this is not an issue with the command itself, however, rather a side-effect of how package listing works:

```shell
kubernetes $ go list ./... | rg klog | wc -l
0 
kubernetes/vendor $ go list ./... | rg klog | wc -l
0 
kubernetes/vendor/k8s.io $ go list ./... | rg klog | wc -l
0 
kubernetes/vendor/k8s.io/klog $ go list ./...
go: warning: "./..." matched no packages
kubernetes/vendor/k8s.io/klog/v2 $ go list ./...
k8s.io/klog/v2
```

- [x] Tests pass


internal/pkg/funccalls/analyzer_test.go
```diff
@@ -1,4 +1,4 @@
-// Copyright 2020 Google LLC
+// Copyright 2019 Google LLC
```
vinayakankugoyal, line 1: Revert this line?

mlevesquedion: Hmm. GitHub messed up, this has nothing to do with my changes.

PR #195, opened 2020-11-19
Add tests for field propagation

mlevesquedion:
This PR description is partly outdated, since this PR no longer includes any code to make the tests pass.

This PR adds (failing) tests for a variety of cases involving fields.

---

This PR adds a test + code to make it pass.

On master, the following case is failing:
```go
func TestTaintField(s core.Source, i *core.Innocuous) {
	i.Data = s.Data
	core.Sink(i)      // want "a source has reached a sink"
	core.Sink(i.Data) // want "a source has reached a sink"
}
```

This is because of this check in `source.go` :
```go
		if !s.config.IsSourceField(typPath, typName, fieldName) {
			return
		}
```

The check is incorrect from a propagation perspective. Indeed, we do want to propagate taint to `i.Data` in the example above, but that won't happen because `Innocuous` isn't a source, so none of its fields are source fields.

Writing the check in the following way solves that issue:
```go
		if s.config.IsSourceType(typPath, typName) && !s.config.IsSourceField(typPath, typName, fieldName) {
			return
		}
```

However, this leaves the issue of traversing through a non-Source field on a Source type, when the non-Source field is being tainted:
```go
func TestTaintNonSourceFieldOnSourceType(s core.Source, i *core.Innocuous) {
	s.ID, _ = strconv.Atoi(s.Data)
	core.Sink(s.ID) // TODO want "a source has reached a sink"
}
```

(Note that this case is also failing on master.)

This will likely be trickier to handle, and under the current approach, I don't see a clean way to do it. The crux of the issue is clearly visible in this graph:

image

The two references to `s.ID` are mapped to distinct instructions in the SSA. The references are tied through an Alloc, but we don't traverse through that Alloc. I note that this would not be an issue with an interpretation-based approach.

Another problem (unrelated to fields) occurs if we tweak the original test case slightly : 
```go
func TestTaintField(s core.Source, i core.Innocuous) {
	i.Data = s.Data
	core.Sink(i)      // TODO want "a source has reached a sink"
	core.Sink(i.Data) // TODO want "a source has reached a sink"
}
```
(The `i` parameter is no longer a pointer.)

When `i` is not a pointer, it gets an Alloc in the function's body, which will prevent the traversal from reaching the sinks. This would also not be an issue with an interpretation-based approach.

- [x] Tests pass


PurelyApplied (commented):
First off, sorry for letting this PR languish for three weeks.

tl;dr: I think that the second test you introduced is within the scope of this PR.  It should be passing if we are addressing root causes.

-----

I think the root issue here is that a tainted wrapper is tainting all of its fields.  Our goal should be:

* Tainting a field should taint the containing struct
* A tainted struct should not taint every field

It's that second point that we should remedy, I think.  That should address the new test you introduced above, i.e.

```golang
func TestTaintNonSourceFieldOnSourceType(s core.Source, i *core.Innocuous) {
	s.ID, _ = strconv.Atoi(s.Data)
	core.Sink(s.ID) // TODO want "a source has reached a sink"
}
```

-----

Even the above statement isn't as straight-forward as I was initially thinking.  For instance, we want to taint a struct when we taint a field directly, such as the tests provided.  But we want to taint everything in a struct if that struct is passed by reference, i.e.

```golang
func Foo(s Source) {
  ruin(&s) // without interprocedural assessment, ruin can do anything, so it _would_ taint every field of s
  sink(s.ID)  // want "..." potentially ruined by pass-by-ref above
}
```

The same would potentially be true for any reference-like field, regardless of whether it's passed by reference or not to a tainting method.

-----

A lot of that feels out of scope for our analysis, or at least would be fine as a Known Issue in the short-term future.  But this is the proverbially sticky wicket.

-----

This block seems to be doing some extra lifting here where we "propagate" a source field from a source type to identify the (conceptually) "real" source.  Out of scope for this PR, but that doesn't feel like propagation logic.  That feels like source-identification logic.

-----

All this to say... I'm not a big fan of this two-line fix.  It feels like it's kicking the can down the road and not fixing the underlying concern.  But the underlying concern might be a bigger monster than I initially thought.  So I'm going to mull this over and maybe come back to it in a day or two.

mlevesquedion:
Thank you for the exposition. I agree with most of your assessment. I think for now the best thing to do with this PR would be to: 1. remove the changes to `source.go`, 2. expand the tests, 3. create a GitHub issue to explain the problem and capture the failing cases.

> Tainting a field should taint the containing struct

I don't think we should consider that structs can be "tainted". I think we should model structs as bags of fields, where each field can be tainted independently. When a struct reaches a sink, we look at the taint status of its fields to see if at least one (possibly nested) field is tainted. (In a similar vein, I don't think we should consider that structs can be sources. A field in a struct can be a source. The struct is merely holding the field.)

I'm not sure we can do this with the traversal alone. At the very least, we'll need some kind of mapping from `struct-value` to `bag-of-fields`. That would be one more thing to lug around. 

Calling a function with a reference to a struct is an interesting case. I agree that without IPA the best we can do is to taint all the fields.

PurelyApplied (approved):
Discussed off-site, but to keep notes here:

> I don't think we should consider that structs can be "tainted". I think we should model structs as bags of fields, [...]

I don't disagree in sentiment, but I do in phrasing.  My stance is that we might want to revise what it means for a struct to be tainted.  I would avoid saying that a struct can't be tainted, because that would feel like it means we shouldn't report when a tainted struct reaches a sink.  It feels to me more that we need to not propagate that taint to every field.

> At the very least, we'll need some kind of mapping [...]

I'm not sure we will.  Taint should still correspond to a given `ssa.Value`, regardless as to what type that value holds.

-----

In any case, the addition of new tests is good to document this discussion and goal.  I'll try to look at selective propagation, but it might need some larger refactoring in propagation to distinguish directionality that we previously haven't considered.  With the holidays nearly upon us, I doubt I'll get to that before the new year.

mlevesquedion:
Agree on "structs can be tainted". I think it's valid to say that e.g. a struct is tainted if at least one of its fields is tainted.

> I'm not sure we will. Taint should still correspond to a given ssa.Value, regardless as to what type that value holds.

Right, if the field is referenced somewhere in the function there will be a value for it. However, I still think we may need something in addition to that, because e.g. two references to the same field can be different values, but if one is tainted, so is the other. All that being said, this is an implementation concern at this point.

PR #196, opened 2020-11-19
Introduce funccalls analyzer

mlevesquedion:
This PR introduces an analyzer that finds calls to sinks and sanitizers.

The purpose of this analyzer is twofold:
1. Allow users to verify that their config is matching what they expect (integration into an actual command in #193)
2. Facilitate identification of calls to sinks and sanitizers by downstream analyzers such as levee and source (integration in #197)

Redundant tests in the `config` package are removed.

- [x] Tests pass

### DIFF
--- a/internal/pkg/funccalls/analyzer.go
+++ b/internal/pkg/funccalls/analyzer.go
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// Package funcs contains an analyzer that performs identification of
+// Package funccalls contains an analyzer that performs identification of
 // sink and sanitizer function definitions and calls. It also identifies
 // excluded functions.
 package funccalls
--- a/internal/pkg/funccalls/analyzer.go
+++ b/internal/pkg/funccalls/analyzer.go
@@ -12,49 +12,62 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package config
+// Package funcs contains an analyzer that performs identification of
+// sink and sanitizer function definitions and calls. It also identifies
+// excluded functions.
+package funccalls
 
 import (
-	"path/filepath"
-	"testing"
-
+	"github.com/google/go-flow-levee/internal/pkg/config"
 	"github.com/google/go-flow-levee/internal/pkg/utils"
 	"golang.org/x/tools/go/analysis"
-	"golang.org/x/tools/go/analysis/analysistest"
 	"golang.org/x/tools/go/analysis/passes/buildssa"
+	"golang.org/x/tools/go/ssa"
 )
 
-var testAnalyzer = &analysis.Analyzer{
-	Name:     "config",
+var Analyzer = &analysis.Analyzer{
+	Name:     "funccalls",
 	Run:      runTest,
-	Doc:      "test harness for the logic related to config",
+	Flags:    config.FlagSet,
+	Doc:      `The funccalls analyzer finds calls to sink and sanitizer functions.`,
 	Requires: []*analysis.Analyzer{buildssa.Analyzer},
 }
 
 func runTest(pass *analysis.Pass) (interface{}, error) {
 	in := pass.ResultOf[buildssa.Analyzer].(*buildssa.SSA)
 
-	conf, err := ReadConfig()
+	conf, err := config.ReadConfig()
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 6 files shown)


internal/pkg/funccalls/analyzer.go
```diff
@@ -0,0 +1,73 @@
...
+	conf, err := config.ReadConfig()
+	if err != nil {
+		return nil, err
+	}
+
+	for _, f := range in.SrcFuncs {
+		if conf.IsExcluded(utils.DecomposeFunction(f)) {
+			continue
+		}
+		for _, b := range f.Blocks {
```
PurelyApplied, line 48: With the parent comment, I would've expected a quick inspection of `f` here rather than examination of `f.Blocks` for invocations.

internal/pkg/funccalls/analyzer.go
```diff
@@ -0,0 +1,73 @@
...
+				callee := c.Call.StaticCallee()
+				if callee == nil {
+					continue
+				}
+				switch {
+				case conf.IsSink(utils.DecomposeFunction(callee)):
+					reportCall(pass, c, callee, "sink")
+				case conf.IsSanitizer(utils.DecomposeFunction(callee)):
+					reportCall(pass, c, callee, "sanitizer")
+				}
```
PurelyApplied, line 63: Just spitballing here:

Would it be better conceptualization to separate sink and sanitizer identification?  I.e.

```golang
func funcAnalyzer(desc string, matchers []config.FuncMatcher) *analysis.Analyzer {
  // ...
}

var (
  sinkAnalyzer = funcAnalyzer("sink", c.Sinks)
  sanitizerAnalyzer = funcAnalyzer("sanitizer", c.Sanitizers)
)
```

Probably too much gilding for one rose, but it would simplify the analyzer to "just identify this one category."  \shrug

PurelyApplied (commented):
High level thoughts:

If part of the the declared purpose is 
> Allow users to verify that their config is matching what they expect

I would expect reporting to occur at function declaration, not invocation.  I would expect the latter to be multiple orders of magnitude more frequent, and a user is not going to want to wade through all that.  In particular, if you use `SinkA` much more frequently than `SinkB`, I wouldn't want to have to wade through pages of one kind of call to verify that my other is actually matching.

mlevesquedion:
Reporting declarations is implemented in #193.

My motivation for adding reporting for calls as well as declarations is that you may not have access to the declaration, or it may be inaccessible for other reasons (e.g. in #193 I discovered that the way we currently list packages won't find `klog` sinks on `kubernetes/kubernetes`).

If a user is only interested in `SinkA`, they could either write a quick throwaway config that contains only `SinkA`, or pipe the output of the tool into `grep SinkA` or use some other filtering method.

internal/pkg/funccalls/analyzer.go
mlevesquedion: Oops. The doc comment is outdated. Fixed.

internal/pkg/funccalls/analyzer.go
mlevesquedion: I think in terms of how this is used it's better to have a single analyzer.

In #193 the intent is to report all configured calls, which includes both sinks and sanitizers.

In #197 `levee` needs to know which calls are sinks and `source` needs to know which calls are sinks and which are sanitizers.

It seems most clients need to know about both kinds of calls.

Another point is that these two analyzers would have nearly identical code, even down to their tests, indeed they would have nearly the same behavior.

vinayakankugoyal:
A more general question about changes like these (where we add a new analyzer): Is there a design discussion/design doc for changes like these or is the discussion solely done in the PR, at which point the code has already been written?

mlevesquedion:
Good question. There isn't a formal process. For more complex features, design discussions are typically done offline before starting the work. For most changes, I don't think an in-depth design is worth the effort, I think it is more efficient to write the code and make adjustments on the PR as necessary.

Adding a new analyzer is not a special case in my opinion.

PurelyApplied:
> A more general question about changes like these (where we add a new analyzer): Is there a design discussion/design doc for changes like these or is the discussion solely done in the PR, at which point the code has already been written?

There's a [design doc](https://github.com/google/go-flow-levee/tree/master/design) here that sketched out a general analysis-flow, but it's rather stale at this point.

Issue #188, opened 2020-11-12
Determine correct propagation behavior for ambiguous cases


mlevesquedion:
PR #178 added explicit handling for each node type.

For most node types, it is obvious what the correct behavior is, e.g. when doing a `MapUpdate`, we shouldn't be traversing to the `Key` or the `Value`, only the `Map`.

For other node types, the correct behavior is not obvious.

Here are the cases that need further investigation, and likely, more tests: 
- [x] `Field`: addressed by #201.
- [x] `Store`: addressed by #199.
- [x] `Select` and `SelectCase`: addressed by #235.
- [x] `MakeInterface`: see #209.
- [x] `TypeAssert`: see #210.
- [x] `Lookup` and `Next`.
- [x] `Slice`: see #236.
- [x] `UnOp`: see below.

mlevesquedion:
I have investigate `ssa.UnOp` instructions and I believe the best we can do for now is to keep visiting the `Operands`. Under the current approach I don't see a straightforward way to handle the following cases without visiting the `Operands`. I think this would be easy to handle with an interpretation-based approach, however.

First, consider this case:
```go
func TestArrOfPointersIsTainted(s core.Source, arr [1]*string) {
	colocateArrOfPointers(s, arr)
	core.Sink(arr) // want "a source has reached a sink"
}
```

The relevant portion of the SSA graph is:

![image](https://user-images.githubusercontent.com/28677454/102559726-8f698600-409e-11eb-942e-f5766e88fdf0.png)

We need to traverse to the `Operand` (in this case, `t1`) when coming from the `Call`, because otherwise we can't get to the `sink`.

Also consider this case:
```go
func TestSanitizedSourceDoesNotTriggerFindingWhenTypeAsserted(s *core.Source) {
	sanitized := core.Sanitize(s)[0].(*core.Source)
	core.Sinkf("Sanitized %v", sanitized)
}
```

The relevant portion of the SSA graph is:

![image](https://user-images.githubusercontent.com/28677454/102559814-c049bb00-409e-11eb-8015-ea8a77bdda60.png)

We need to traverse to the `Operand` in order to find the `Sanitizer`. Otherwise, the `Source` produced by the `TypeAssert` is considered unsanitized. This could be seen as a failure of the `isProducedBySanitizer` function. I have opened an issue here: #237.

PR #199, opened 2020-11-19
Only traverse to store addr

mlevesquedion:
Investigating a new test case suggested by @PurelyApplied led me to find a related case where propagating to the `Value` in `Store` causes a spurious report to be produced:

```go

func TestStoringToTaintedAddrDoesNotTaintStoredValue() {
	myChan := make(chan string)
	s := core.Source{Data: "foo"}
	recv := <-myChan
	s.Data = recv
	core.Sink(recv)
	core.Sink(myChan)
}
```

On master, `core.Sink(recv)` in the above example yields a spurious report. (`core.Sink(myChan)` is there for additional certainty.)

Not propagating to the `Value` seems like the right thing to do, since it makes no sense. However, it does mean that we no longer handle this case:

```go
func TestPanicTypeAssertSource(i interface{}) {
	s := i.(core.Source)
	_ = s
	// The dominating type assertion would panic if i were not a source type.
	core.Sink(i) // TODO want "a source has reached a sink"
}
```

I think this is an acceptable regression, since this test was passing because we were doing a weird propagation, not because we were properly handling the semantics of the code.

Related issue: #188.

- [x] Tests pass

### DIFF
--- a/internal/pkg/levee/testdata/src/example.com/tests/store/tests.go
+++ b/internal/pkg/levee/testdata/src/example.com/tests/store/tests.go
@@ -1,0 +1,28 @@
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package store
+
+import (
+	"example.com/core"
+)
+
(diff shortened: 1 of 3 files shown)


PurelyApplied (approved):

PurelyApplied:
> I think this is an acceptable regression, since this test was passing because we were doing a weird propagation, not because we were properly handling the semantics of the code.

I agree.  It does point out that we might want to distinguish why we are suppressing tests within the `// TODO` comments, though.  This test is suppressed because carrying information forward from a type assertion is currently out of scope.  The `// TODO want` is more of a "we might want to consider this case" than a "this is a regression."

tl;dr: we might want to adopt a habit of `// TODO(#<issue number>)`.

aaaaaand now I've taken too long to type this that you've already merged.  :)

mlevesquedion:
Sorry about that.

I like the `TODO(#issue number)` idea a lot. I think we should have some kind of visibility into those as well, e.g. to make sure that:
1. We know about all the TODOs lurking in the code
2. Every TODO is associated with an issue
I'm thinking maybe something like a badge that shows the number of TODOs (we would want to drive this number down).

PR #201, opened 2020-11-23
Unify traversal through Field and FieldAddr instructions

mlevesquedion:
This PR is part of #188.

This PR uniformizes the way propagation works for `Field` and `FieldAddr` instructions. While the dynamic semantics of these two instructions are a little different, for the purposes of our traversal they should be the handled the same way.

I know we've recently been discussing (#195, #228) how we might need to rethink the way we handle fields, and I think unifying the way these instructions are handled is a necessary step towards that. 


internal/pkg/levee/testdata/src/example.com/tests/fields/tests.go
```diff
@@ -41,7 +46,7 @@ func TestProtoStyleFieldAccessorPIISecondLevel(wrapper struct{ *core.Source }) {
 	core.Sinkf("Source id: %v", wrapper.Source.GetID())
 }
 
-func tesDirectFieldAccessorPIISecondLevel(wrapper struct{ *core.Source }) {
+func TestDirectFieldAccessorPIISecondLevel(wrapper struct{ *core.Source }) {
```
mlevesquedion, line 49: Sorry for the drive-by, unrelated fix, but this bugs me every time I see it.

internal/pkg/levee/testdata/src/example.com/tests/fields/tests.go
```diff
@@ -34,6 +34,14 @@ func TestDirectFieldAccess(c *core.Source) {
 	core.Sinkf("ID: %v", c.ID)
 }
 
+func TestInlinedDirectFieldAccess() {
+	// This pattern is unlikely to occur in real code.
+	// The intent is to get Field instructions in the SSA
+	// so that we can validate that those are handled correctly.
+	core.Sinkf("Data: %v", core.Source{}.Data) // want "a source has reached a sink"
```
PurelyApplied, line 41: Consider including something other than a const-empty string as `Data`.  Even though I know it's not something we handle right now, part of me thinks this shouldn't produce a `Diagnostic`.

PurelyApplied (approved):

PR #204, opened 2020-12-01
Add configurable error message.

PurelyApplied:
The generic message "a source has reached a sink" does not inform a user without knowledge of taint propagation what needs to be done to correct the problem.  In a usecase such as Kubernetes's adoption of analysis as presubmit, it would be useful to link documentation when a diagnostic is found.

Given that an analysis driver is going to aggregate results rather than us directly, it seems to me that we should include the configured error message in each package's pass.  We could try to add some syncing, but I don't know that the reduction would be greatly worth the effort.

Testing is a concern here, as another rough edge of the `analysistest` package.  I'd take your input on what you think would be the best approach.

* Option A: We could report as a package fact.  This would let us include a `// want` at the package level.
* Option B: We could just use a `fmt.Print*`.  But to test that the message is included would mean that we would need to juggle output streams ourselves.

I'm personally inclined towards A, which I've implemented here (but not yet updated the now-failing tests -- I'll probably split it out to a separate configuration and a smaller test set).


internal/pkg/levee/levee.go
```diff
@@ -64,7 +70,7 @@ func run(pass *analysis.Pass) (interface{}, error) {
 				case callee != nil && conf.IsSink(utils.DecomposeFunction(v.Call.StaticCallee())):
 					for _, s := range sources {
 						if s.HasPathTo(instr.(ssa.Node)) && !s.IsSanitizedAt(v) {
-							report(pass, s, v)
+							items = append(items, reportItems{s, v})
```
vinayakankugoyal, line 73: what does reportItems{s, v} do? I have never seen that syntax before.

mlevesquedion: It's an unkeyed struct literal.

vinayakankugoyal: Oops, somehow I got confused that reportItems is a slice. 🤦

mlevesquedion: That's happened to me too.

To be fair, I think the literal should be keyed, it would make this snippet more readable.

internal/pkg/config/config.go
```diff
@@ -256,11 +257,13 @@ func ReadConfig() (*Config, error) {
 		bytes, err := ioutil.ReadFile(configFile)
 		if err != nil {
 			readConfigCachedErr = fmt.Errorf("error reading analysis config: %v", err)
+			fmt.Println(readConfigCachedErr)
```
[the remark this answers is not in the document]
mlevesquedion: Consider moving this `Println` call and the next to a common `if readConfigCachedErr != nil` block.

mlevesquedion (commented):
I think a package fact could be a good way to do this. What are your plans for reporting it though? A fact does not lead to a report, we will still need to produce a report *somewhere*, e.g. we could use NoPos if we don't want a huge path message prefixing the thing, or we could use the package identifier, but that could land in any file, or that might not even have a Pos, I don't know :thinking: 

Have you considered including the custom message on every report? I'm wondering what that would look like. There's a risk that the signal might get lost in the noise, but I think if we did something like this it would probably be fine:
```
.../file.go:12:34: a source has reached a sink,
source: ...
sink: ...
<custom message here>
```

internal/pkg/levee/levee.go
```diff
@@ -35,6 +35,11 @@ var Analyzer = &analysis.Analyzer{
 	Requires: []*analysis.Analyzer{source.Analyzer, fieldpropagator.Analyzer},
 }
 
+type reportItems struct {
+	s *source.Source
+	v *ssa.Call
```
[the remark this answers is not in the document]
PurelyApplied: Actually, within the context of "something we need to report," we should probably call `v -> c -> call -> sink`, since that is what we're reporting.  `s -> source` is a good \*ahem\* call, too.  Updated both.

internal/pkg/config/config.go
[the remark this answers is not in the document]
PurelyApplied: It was, but I didn't mean for those to slip into this PR.  I have a few other thoughts about how config is read / stored / reported, and hopefully I'll get that PR written when I'm back from the holidays.

internal/pkg/config/config.go
```diff
@@ -39,11 +39,12 @@ func init() {
 
 // config contains matchers and analysis scope information
 type Config struct {
-	Sources    []sourceMatcher
-	Sinks      []funcMatcher
-	Sanitizers []funcMatcher
-	FieldTags  []fieldTagMatcher
-	Exclude    []funcMatcher
+	MessageOnError string
```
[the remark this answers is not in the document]
PurelyApplied: Done.

PurelyApplied (commented):
> Have you considered including the custom message on every report? I'm wondering what that would look like. There's a risk that the signal might get lost in the noise, but I think if we did something like this it would probably be fine:
> 
> ```
> .../file.go:12:34: a source has reached a sink,
> source: ...
> sink: ...
> <custom message here>
> ```

Actually, I really like that.  I've also rather wished our reporting of sources and sinks weren't on the same line, because IntelliJ will only make the first position clickable when looking at a "stacktrace."  I'll rework reporting to that, as long as it doesn't play merry hobs with our existing `// want "now a multiline string"`.  But I think that's done via regexp, so introducing `\n`s in the diagnostic string should be okay.  Here's hoping...

internal/pkg/levee/levee.go
```diff
@@ -44,6 +49,7 @@ func run(pass *analysis.Pass) (interface{}, error) {
 	funcSources := pass.ResultOf[source.Analyzer].(source.ResultType)
 	taggedFields := pass.ResultOf[fieldtags.Analyzer].(fieldtags.ResultType)
 
+	var items []reportItems
```
vinayakankugoyal, line 52: I am not seeing how this in getting updated? Did some merge remove some code?

PurelyApplied: Oh, I meant to convert back to Draft.  It's not anymore.  I was dodging doing too much when resolving merge conflicts.

I also meant to put `TODO` in with the `// build items` comments.  But it's going to require a change to the `reportSource*` methods, which I didn't want to do in one push.

Converting back to Draft

PurelyApplied:
Returning to this now that the holidays are behinds us...
```
.../file.go:12:34: a source has reached a sink,
source: ...
sink: ...
<custom message here>
```

Hmmm... Diagnostic needs to be attached to a position.  We've been using the position of the sink.  So if we want to report `source: ... sink:`, it would end up being something like

```
.../file.go:12:34: a source has reached a sink,
source: ...
sink: .../file.go:12:34
<custom message here>
```

I was also wondering if the `<custom message>` should just replace `a source as reached a sink` entirely (and use `"a source has reached a sink"` or similar as a default value if the message is not specified).  That would collapse it down to

```
.../file.go:12:34: <custom message|default: the following source has reached this sink>
	source: ...
```
Thoughts, @mlevesquedion?

mlevesquedion:
Right, and for testing purposes, it's practical to report at the sink, too.

I don't have a hard preference. I think the most important thing is to have the report be on multiple lines, for readability.

I like "the following source has reached this sink". I'm not sure how a user could meaningfully customize that, though. To me it would make more sense to have the custom message be at the end:

```go
.../file.go:12:34: the following source has reached this sink
source: ...
<custom message>
```

(I think it would also be useful to include more info about the source, e.g. its type. That should be in a future PR.)

PurelyApplied:
Tests currently failing require #255.

PurelyApplied:
This PR has evolved a bit at this point, so to recap so you don't have to dig all 11 commits...

* This PR adds `ReportMessage` to configuration, which is included with each diagnostic reported.
* `Config` is now plumbed to reporting methods.
* Mild formatting changes - `,` -> `\n`.
* Add e2e tests for verify formatting.

--- 

Out of scope: We might want to establish a convention for grouping configs with the testdata that is the intended consumption.  If `TestFoo` is only using `test-foo.yaml` and running analysis on `testdata/src/foo`, I think it would make more sense to keep the configuration YAML in `testdata/src/foo` rather than the `testdata/` root.  I've added mine there for consistency for now.

internal/pkg/levee/levee.go
```diff
@@ -70,18 +69,21 @@ func run(pass *analysis.Pass) (interface{}, error) {
...
+			report(conf, pass, src, sink.(ssa.Node))
 			break
 		}
 	}
 }
 
-func report(pass *analysis.Pass, source *source.Source, sink ssa.Node) {
+func report(conf *config.Config, pass *analysis.Pass, source *source.Source, sink ssa.Node) {
 	var b strings.Builder
 	b.WriteString("a source has reached a sink")
```
vinayakankugoyal, line 83: why cant this be part of the format string below?

vinayakankugoyal: This following is pretty readable to me:-
fmt.Fprintf(&b, "a source has reached a sink \n source: %v", pass.Fset.Position(source.Pos()))
but I don't feel too strongly about this, so its ok to keep it like this.

mlevesquedion (approved):
LGTM

vinayakankugoyal (approved):

mlevesquedion:
> We might want to establish a convention for grouping configs with the testdata that is the intended consumption. If TestFoo is only using test-foo.yaml and running analysis on testdata/src/foo, I think it would make more sense to keep the configuration YAML in testdata/src/foo rather than the testdata/ root. I've added mine there for consistency for now.

I agree, we should move all the configs to their relevant directory.

internal/pkg/levee/levee.go
PurelyApplied: \shrug.

I tend to go towards using additional lines when we have `\n` in the format, just to visualize the `\n`.

I suppose if we wanted to go in the entire other direction, we could collapse the entire method to

```golang
msg :=  strings.Trim(fmt.Sprintf("a source has reached a sink\n source: %v\n %v", pass.Fset.Position(source.Pos()), conf.ReportMessage), "\n "))
pass.Reportf(sink.Pos(), msg)
```

I was really just doing a minimal diff and keeping the style that was already there with the string buffer.  I also have no strong feelings about this if you wanted to make a PR to clean it up though.

PR #213, opened 2020-12-08
Add validation to config fields.

### LINKED ISSUE #206: Enforce strict config unmarshalling
In order to minimize the risk of configuration errors, we would like the unmarshalling of the configuration to be strict. Specifically, it should disallow unknown fields in matchers. This avoids errors where a typo in a key name leads to an incorrect configuration. We don't want typos to lead to silent failures, e.g. a matcher is broken and the user thinks their code is safe when it is not.

There are already two test cases covering this behavior to some extent, in `config/matcher_test.go`:

```go
		{
			desc: "Unmarshaling is strict",
			yaml: `
Blahblah: foo
PackageRE: bar`,
			shouldErrorOnLoad: false, // TODO true
		},
```

(It is the same case, in the tests for two different matchers.)

benhxy:
Fixes #206 

Add validation for config yaml fields.
Notice that we have both upper case (e.g. "Key") and lower case (e.g. "key") in our unit tests and sample files. The validation will allow both.

- [x] Tests pass

### DIFF
--- a/internal/pkg/config/config.go
+++ b/internal/pkg/config/config.go
@@ -31,24 +31,18 @@ import (
 
 var (
 	// FlagSet should be used by analyzers to reuse -config flag.
-	FlagSet                      flag.FlagSet
-	configFile                   string
-	validFuncMatcherFields       []string
-	validSourceMatcherFields     []string
-	validFieldTagMatcherFields   []string
-	validFuncMatcherFieldSet     map[string]bool
-	validSourceMatcherFieldSet   map[string]bool
-	validFieldTagMatcherFieldSet map[string]bool
+	FlagSet                    flag.FlagSet
+	configFile                 string
+	validFuncMatcherFields     []string
+	validSourceMatcherFields   []string
+	validFieldTagMatcherFields []string
 )
 
 func init() {
 	FlagSet.StringVar(&configFile, "config", "config.yaml", "path to analysis configuration file")
 	validFuncMatcherFields = []string{"package", "packagere", "receiver", "receiverre", "method", "methodre"}
-	validFuncMatcherFieldSet = toStringSet(validFuncMatcherFields)
 	validSourceMatcherFields = []string{"package", "packagere", "type", "typere", "field", "fieldre"}
-	validSourceMatcherFieldSet = toStringSet(validSourceMatcherFields)
 	validFieldTagMatcherFields = []string{"key", "val", "value"}
-	validFieldTagMatcherFieldSet = toStringSet(validFieldTagMatcherFields)
 }
 
 // Config contains matchers and analysis scope information.
@@ -161,7 +155,7 @@ type rawFieldTagMatcher struct {
 }
 
 func (ft *fieldTagMatcher) UnmarshalJSON(bytes []byte) error {
-	if err := validateFieldNames(&bytes, "fieldTagMatcher", validFieldTagMatcherFields, validFieldTagMatcherFieldSet); err != nil {
+	if err := validateFieldNames(&bytes, "fieldTagMatcher", validFieldTagMatcherFields); err != nil {
 		return err
 	}
 
@@ -207,7 +201,7 @@ type rawSourceMatcher struct {
 }
 
 func (s *sourceMatcher) UnmarshalJSON(bytes []byte) error {
-	if err := validateFieldNames(&bytes, "sourceMatcher", validSourceMatcherFields, validSourceMatcherFieldSet); err != nil {
+	if err := validateFieldNames(&bytes, "sourceMatcher", validSourceMatcherFields); err != nil {
 		return err
 	}
 
@@ -260,7 +254,7 @@ type rawFuncMatcher struct {
 }
 
 func (fm *funcMatcher) UnmarshalJSON(bytes []byte) error {
-	if err := validateFieldNames(&bytes, "funcMatcher", validFuncMatcherFields, validFuncMatcherFieldSet); err != nil {
+	if err := validateFieldNames(&bytes, "funcMatcher", validFuncMatcherFields); err != nil {
 		return err
 	}
 
@@ -315,23 +309,21 @@ func ReadConfig() (*Config, error) {
 	return readConfigCached, readConfigCachedErr
 }
 
-func validateFieldNames(bytes *[]byte, matcherType string, validFields []string, validFieldSet map[string]bool) error {
+func validateFieldNames(bytes *[]byte, matcherType string, validFields []string) error {
 	rawMap := make(map[string]interface{})
 	if err := json.Unmarshal(*bytes, &rawMap); err != nil {
 		return err
 	}
 	for label := range rawMap {
-		if !validFieldSet[strings.ToLower(label)] {
+		matched := false
+		for _, valid := range validFields {
+			if strings.EqualFold(label, valid) {
+				matched = true
+			}
+		}
+		if !matched {
 			return fmt.Errorf("%v is not a valid config field for %s, expect one of: %v", label, matcherType, validFields)
 		}
 	}
 	return nil
-}
-
-func toStringSet(ss []string) map[string]bool {
-	ret := make(map[string]bool)
-	for _, s := range ss {
-		ret[s] = true
-	}
-	return ret
 }
--- a/internal/pkg/config/config.go
+++ b/internal/pkg/config/config.go
@@ -33,16 +33,13 @@ var (
 	// FlagSet should be used by analyzers to reuse -config flag.
 	FlagSet                    flag.FlagSet
 	configFile                 string
-	validFuncMatcherFields     []string
-	validSourceMatcherFields   []string
-	validFieldTagMatcherFields []string
+	validFuncMatcherFields     []string = []string{"package", "packagere", "receiver", "receiverre", "method", "methodre"}
+	validSourceMatcherFields   []string = []string{"package", "packagere", "type", "typere", "field", "fieldre"}
+	validFieldTagMatcherFields []string = []string{"key", "val", "value"}
 )
 
 func init() {
 	FlagSet.StringVar(&configFile, "config", "config.yaml", "path to analysis configuration file")
-	validFuncMatcherFields = []string{"package", "packagere", "receiver", "receiverre", "method", "methodre"}
-	validSourceMatcherFields = []string{"package", "packagere", "type", "typere", "field", "fieldre"}
-	validFieldTagMatcherFields = []string{"key", "val", "value"}
 }
 
 // Config contains matchers and analysis scope information.
--- a/internal/pkg/config/config.go
+++ b/internal/pkg/config/config.go
@@ -31,36 +31,42 @@ import (
 
 var (
 	// FlagSet should be used by analyzers to reuse -config flag.
-	FlagSet                    flag.FlagSet
-	configFile                 string
-	validFuncMatcherFields     map[string]bool
-	validSourceMatcherFields   map[string]bool
-	validFieldTagMatcherFields map[string]bool
+	FlagSet                      flag.FlagSet
+	configFile                   string
+	validFuncMatcherFields       []string
+	validSourceMatcherFields     []string
+	validFieldTagMatcherFields   []string
+	validFuncMatcherFieldSet     map[string]bool
+	validSourceMatcherFieldSet   map[string]bool
+	validFieldTagMatcherFieldSet map[string]bool
 )
 
 func init() {
 	FlagSet.StringVar(&configFile, "config", "config.yaml", "path to analysis configuration file")
-	validFuncMatcherFields = map[string]bool{
-		"package": true, "Package": true,
-		"packageRE": true, "PackageRE": true,
-		"receiver": true, "Receiver": true,
-		"receiverRE": true, "ReceiverRE": true,
-		"method": true, "Method": true,
-		"methodRE": true, "MethodRE": true,
+	validFuncMatcherFields = []string{
+		"package", "Package",
+		"packageRE", "PackageRE",
+		"receiver", "Receiver",
+		"receiverRE", "ReceiverRE",
+		"method", "Method",
+		"methodRE", "MethodRE",
 	}
-	validSourceMatcherFields = map[string]bool{
-		"package": true, "Package": true,
-		"packageRE": true, "PackageRE": true,
-		"type": true, "Type": true,
-		"typeRE": true, "TypeRE": true,
-		"field": true, "Field": true,
-		"fieldRE": true, "FieldRE": true,
+	validFuncMatcherFieldSet = toStringSet(validFuncMatcherFields)
+	validSourceMatcherFields = []string{
+		"package", "Package",
+		"packageRE", "PackageRE",
+		"type", "Type",
+		"typeRE", "TypeRE",
+		"field", "Field",
+		"fieldRE", "FieldRE",
 	}
-	validFieldTagMatcherFields = map[string]bool{
-		"key": true, "Key": true,
-		"val": true, "Val": true,
-		"value": true, "Value": true,
+	validSourceMatcherFieldSet = toStringSet(validSourceMatcherFields)
+	validFieldTagMatcherFields = []string{
+		"key", "Key",
+		"val", "Val",
+		"value", "Value",
 	}
+	validFieldTagMatcherFieldSet = toStringSet(validFieldTagMatcherFields)
 }
 
 // Config contains matchers and analysis scope information.
@@ -173,15 +179,8 @@ type rawFieldTagMatcher struct {
 }
 
 func (ft *fieldTagMatcher) UnmarshalJSON(bytes []byte) error {
-	// Unknown config fields are not allowed.
-	rawMap := make(map[string]interface{})
-	if err := json.Unmarshal(bytes, &rawMap); err != nil {
+	if err := validateFieldNames(&bytes, "fieldTagMatcher"); err != nil {
 		return err
-	}
-	for label := range rawMap {
-		if !validFieldTagMatcherFields[label] {
-			return fmt.Errorf("%v is not a valid config field, expect one of: %v", label, validFieldTagMatcherFields)
-		}
 	}
 
 	raw := rawFieldTagMatcher{}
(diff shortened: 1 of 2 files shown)


internal/pkg/config/matcher_test.go
```diff
@@ -296,6 +296,38 @@ Field: qux`,
...
+			yaml: `
+Key: good
+UnknownField: bad`,
+			shouldErrorOnLoad: true,
+		}, {
+			desc: "valid field tag config",
+			yaml: `
+Key: good
+Value: AlsoGood`,
+			shouldErrorOnLoad: true,
```
mlevesquedion, line 315: This case is supposed to be valid, so there shouldn't be an error, right?

internal/pkg/config/matcher_test.go
```diff
@@ -31,7 +31,7 @@ func TestFuncMatcherUnmarshaling(t *testing.T) {
 			yaml: `
 Blahblah: foo
 PackageRE: bar`,
-			shouldErrorOnLoad: false, // TODO(#206): true
+			shouldErrorOnLoad: true,
```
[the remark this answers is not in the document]
mlevesquedion: Also consider renaming the test to `TestFuncMatcherUnmarshalingErrorCases` or something that indicates the focus on errors.

internal/pkg/config/matcher_test.go
```diff
@@ -296,6 +296,38 @@ Field: qux`,
 	}
 }
 
+func TestFieldTagMatcherUnmarshalling(t *testing.T) {
+	testCases := []struct {
+		desc, yaml        string
+		shouldErrorOnLoad bool
```
mlevesquedion, line 302: Nit: prefer `wantErr`, the test already indicates that we're interested in "loading" (unmarshalling).

internal/pkg/config/matcher_test.go
```diff
@@ -296,6 +296,38 @@ Field: qux`,
...
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.desc, func(t *testing.T) {
+			ftm := fieldTagMatcher{}
+			err := yaml.UnmarshalStrict([]byte(tc.yaml), &ftm)
+
+			if (err != nil) != tc.shouldErrorOnLoad {
+				t.Errorf("error expectation = %v, but got err=%v", tc.shouldErrorOnLoad, err)
```
mlevesquedion, line 325: [Got before want](https://github.com/golang/go/wiki/TestComments#got-before-want)

mlevesquedion: Nit: why does `err=%v` not have whitespace, but `expectation = %v` does?

internal/pkg/config/config.go
```diff
@@ -29,15 +29,40 @@ import (
...
+	FlagSet                    flag.FlagSet
+	configFile                 string
+	validFuncMatcherFields     map[string]bool
+	validSourceMatcherFields   map[string]bool
+	validFieldTagMatcherFields map[string]bool
+)
 
 func init() {
 	FlagSet.StringVar(&configFile, "config", "config.yaml", "path to analysis configuration file")
+	validFuncMatcherFields = map[string]bool{
```
mlevesquedion, line 43: Have you considered writing each field once in title case and using e.g. `strings.Title` when doing the comparisons?

internal/pkg/config/config.go
```diff
@@ -29,15 +29,40 @@ import (
...
 )
 
-// FlagSet should be used by analyzers to reuse -config flag.
-var FlagSet flag.FlagSet
-var configFile string
+var (
+	// FlagSet should be used by analyzers to reuse -config flag.
+	FlagSet                    flag.FlagSet
+	configFile                 string
+	validFuncMatcherFields     map[string]bool
```
mlevesquedion, line 36: I think it would be preferable to move each of these close to the struct that contains the actual field definitions. It would make it easier to verify that the set of strings is correct.

internal/pkg/config/config.go
```diff
@@ -137,7 +161,39 @@ func (vacuousMatcher) MatchString(string) bool {
...
+
+func (ft *fieldTagMatcher) UnmarshalJSON(bytes []byte) error {
+	// Unknown config fields are not allowed.
+	rawMap := make(map[string]interface{})
+	if err := json.Unmarshal(bytes, &rawMap); err != nil {
+		return err
+	}
+	for label := range rawMap {
+		if !validFieldTagMatcherFields[label] {
+			return fmt.Errorf("%v is not a valid config field", label)
```
mlevesquedion, line 182: I think it would be good for the error message to be a bit more verbose, e.g. to include the fact that you were trying to parse field tag configuration when you encountered the invalid field. Also consider listing the valid fields, e.g. `valid fields include: ...`.

internal/pkg/config/config.go
```diff
@@ -137,7 +161,39 @@ func (vacuousMatcher) MatchString(string) bool {
...
+		}
+	}
+
+	raw := rawFieldTagMatcher{}
+	if err := json.Unmarshal(bytes, &raw); err != nil {
```
mlevesquedion, line 187: I think you could unmarshal into `ft` directly.

internal/pkg/config/config.go
```diff
@@ -169,12 +225,23 @@ type rawSourceMatcher struct {
...
+		}
+	}
+
 	raw := rawSourceMatcher{}
 	if err := json.Unmarshal(bytes, &raw); err != nil {
 		return err
 	}
 
-	// validation: do not double-specify any attribute with literal and regexp
+	// Only one of literal and regexp field can be specified.
```
mlevesquedion, line 244: :+1:

internal/pkg/config/config.go
```diff
@@ -169,12 +225,23 @@ type rawSourceMatcher struct {
...
 	if raw.Package != nil && raw.PackageRE != nil {
 		return fmt.Errorf("expected only one of Package, PackageRE to be configured")
```
mlevesquedion, line 246: For a future PR, if you wish: I think this error message should mention the fact that the code was trying to unmarshal a source definition.

mlevesquedion (commented):
Mostly LGTM, just a few small changes/things to discuss. Lots of nice improvements to documentation in here as well. Good stuff!

internal/pkg/config/config.go
```diff
@@ -137,7 +161,39 @@ func (vacuousMatcher) MatchString(string) bool {
 	return true
 }
 
-// Returns the first non-nil matcher.  If all are nil, returns a vacuousMatcher.
+type fieldTagMatcher struct {
+	Key string
+	Val string
+}
+
+type rawFieldTagMatcher struct {
```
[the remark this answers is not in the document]
benhxy: If we use the same struct, the call to UnmarshalJSON will become recursive and cause stackoverflow.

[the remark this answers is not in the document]
mlevesquedion: Ah yes. Then I guess an alternative would be to set the fields from the map. I think what you have here is preferable though.

internal/pkg/config/config.go
```diff
@@ -137,7 +161,39 @@ func (vacuousMatcher) MatchString(string) bool {
...
+	Key string
+	Val string
+}
+
+func (ft *fieldTagMatcher) UnmarshalJSON(bytes []byte) error {
+	// Unknown config fields are not allowed.
```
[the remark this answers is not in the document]
benhxy: Since different matchers use different stringsets for validation, the only common part is "unmarshal to string map". That's not a lot of duplication to remove.

internal/pkg/config/matcher_test.go
benhxy: Well, in the config file you used "val", not "Value" lol
Let me know if you want to support all four: "val", "Val", "value", "Value".

mlevesquedion: Sorry for not being clearer. The test name is "valid field tag config", so I would expect this case to test for success. I think it's good to have a "success" case. Otherwise, this case seems redundant with the previous one.

mlevesquedion: Consider using lowercase keys in the test to validate that these are accepted.

PurelyApplied: I agree that this seems counter-intuitive.  Probably the biggest issue here is that `Value: AlsoGood` really suggests that the value is correct.  If I were to leave it how it is, I'd write it as `Value: ShouldBeVal`.

Either here or in a future PR, I would say that we should accept a `Value` here as well.  At which point, I might update this test case to be something like

```
Key: good
Val: AlsoGood
Value: BadForRedundancy
```

internal/pkg/config/config.go
benhxy: Added where the error is from to error message

internal/pkg/config/config.go
benhxy: Added

internal/pkg/config/matcher_test.go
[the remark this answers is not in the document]
benhxy: Updated the ntest

internal/pkg/config/config.go
```diff
@@ -85,6 +105,7 @@ func (c Config) IsExcluded(path, recv, name string) bool {
 	return false
 }
 
+// IsSink determines whether a function is a secret sink.
```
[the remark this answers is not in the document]
benhxy: Updated the docs

internal/pkg/config/config.go
benhxy: maybe we should just compare case-insensitively?

internal/pkg/config/matcher_test.go
benhxy: Updated

internal/pkg/config/matcher_test.go
benhxy: Haha somebody before me wrote got after want. I'll try to correct all of them.

internal/pkg/config/config.go
benhxy: I prefer to have global variables on top, so we don't miss initiating them...

internal/pkg/config/config.go
mlevesquedion: Right. I don't think we really care about the casing, so maybe casefolding is the way to go.
@PurelyApplied, any thoughts on this?

internal/pkg/config/config.go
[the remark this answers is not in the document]
mlevesquedion: I was thinking of something like:
```go
func validateFields(bytes []byte, validFields map[string]bool, matcherName string) error {
    rawMap := make(map[string]interface{})
	if err := json.Unmarshal(bytes, &rawMap); err != nil {
		return err
	}
	for label := range rawMap {
		if !validFields[label] {
			return fmt.Errorf("%q is not a valid config field for %q, expect one of: %v", label, matcherName, validFieldTagMatcherFields)
   }
   return nil
}
```
If we need to change this logic at some point, or e.g. the error message, we'll be able to change it in only one place, instead of three.

internal/pkg/config/config.go
```diff
@@ -169,20 +226,31 @@ type rawSourceMatcher struct {
...
 	raw := rawSourceMatcher{}
 	if err := json.Unmarshal(bytes, &raw); err != nil {
 		return err
 	}
 
-	// validation: do not double-specify any attribute with literal and regexp
+	// Only one of literal and regexp field can be specified.
 	if raw.Package != nil && raw.PackageRE != nil {
-		return fmt.Errorf("expected only one of Package, PackageRE to be configured")
+		return fmt.Errorf("expected only one of Package, PackageRE to be configured for a source")
```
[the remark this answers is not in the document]
mlevesquedion: Also see `configured for a function` below.

internal/pkg/config/config.go
```diff
@@ -137,7 +162,39 @@ func (vacuousMatcher) MatchString(string) bool {
...
+
+func (ft *fieldTagMatcher) UnmarshalJSON(bytes []byte) error {
+	// Unknown config fields are not allowed.
+	rawMap := make(map[string]interface{})
+	if err := json.Unmarshal(bytes, &rawMap); err != nil {
+		return err
+	}
+	for label := range rawMap {
+		if !validFieldTagMatcherFields[label] {
+			return fmt.Errorf("%v is not a valid config field, expect one of: %v", label, validFieldTagMatcherFields)
```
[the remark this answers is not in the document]
benhxy: Updated

internal/pkg/config/config.go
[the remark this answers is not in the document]
benhxy: Updated

internal/pkg/config/matcher_test.go
```diff
@@ -73,8 +66,8 @@ MethodRE: bar`,
 			fm := funcMatcher{}
 			err := yaml.UnmarshalStrict([]byte(tc.yaml), &fm)
 
-			if (err != nil) != tc.shouldErrorOnLoad {
-				t.Errorf("error expectation = %v, but got err=%v", tc.shouldErrorOnLoad, err)
+			if err == nil {
+				t.Error("want error, but got err = nil")
```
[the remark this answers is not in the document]
benhxy: Updated

internal/pkg/config/config.go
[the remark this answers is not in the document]
benhxy: Updated

internal/pkg/config/config.go
```diff
@@ -267,3 +332,39 @@ func ReadConfig() (*Config, error) {
...
+	case "fieldTagMatcher":
+		validFields = validFieldTagMatcherFields
+		validFieldSet = validFieldTagMatcherFieldSet
+	case "funcMatcher":
+		validFields = validFuncMatcherFields
+		validFieldSet = validFuncMatcherFieldSet
+	case "sourceMatcher":
+		validFields = validSourceMatcherFields
+		validFieldSet = validSourceMatcherFieldSet
+	}
```
mlevesquedion, line 354: What is the benefit of figuring out the fields from the `matcherType` in this function, as opposed to passing in the appropriate fields in each of the callers? A matcher should know what its set of valid fields is. If we add a matcher later, I don't think we should need to change this function.

internal/pkg/config/config.go
```diff
@@ -29,15 +29,47 @@ import (
...
+	validSourceMatcherFields     []string
+	validFieldTagMatcherFields   []string
+	validFuncMatcherFieldSet     map[string]bool
+	validSourceMatcherFieldSet   map[string]bool
+	validFieldTagMatcherFieldSet map[string]bool
+)
 
 func init() {
 	FlagSet.StringVar(&configFile, "config", "config.yaml", "path to analysis configuration file")
+	validFuncMatcherFields = []string{
```
mlevesquedion, line 46: What is the benefit of maintaining both a slice and a map? IMO we could just keep the slice. Also, what is the benefit of duplicating each field, as opposed to calling `strings.ToLower` before comparing, or comparing with `strings.EqualFold`? IMO doing that would be simpler and less error-prone.

internal/pkg/config/config.go
PurelyApplied: I don't really have an opinion on it one way or t'other.  We wound up with case insensitivity as a side-benefit of using the `k8s` YAML package rather than the `go-yaml` package.  But that choice was driven by not wanting to have to tag every single field with

```golang
type Foo struct{
  FieldName string `yaml: "FieldName"`
}
```

The `go-yaml` unmarshaller gets very upset at in-name capitalization that you have to go manually correct.

-----

All this to say, case folding sounds fine to me.

benhxy: Cool. Added casefolding to comparison.

internal/pkg/config/config.go
benhxy: I was thinking of not copying arrays or maps in function calls to improve performance. But I guess that's too much complexity.

internal/pkg/config/config.go
benhxy: The slice is for error output and map for lookup. I didn't want to nest two for-loops when doing checking.

benhxy: Removed map and only use slice to compare.

internal/pkg/config/config.go
```diff
@@ -29,15 +29,23 @@ import (
...
+	FlagSet                    flag.FlagSet
+	configFile                 string
+	validFuncMatcherFields     []string
+	validSourceMatcherFields   []string
+	validFieldTagMatcherFields []string
+)
 
 func init() {
 	FlagSet.StringVar(&configFile, "config", "config.yaml", "path to analysis configuration file")
+	validFuncMatcherFields = []string{"package", "packagere", "receiver", "receiverre", "method", "methodre"}
```
PurelyApplied, line 43: An honest, non-Socratic question: is there an advantage to having the map initialized in `init()` rather than at its `var` declaration?  The `FlagSet` didn't initialize in-line, so it had to live in an `init`, but arrays could be in-lined.  But more on the in the main comment.

PurelyApplied (commented):
Michael covered almost all of my concerns.

Implementation-wise, I'm not a huge fan of the separation of the structs themselves and these field-maps.  Back when _someone_ \<shifty eye darting\> introduced these methods without strict behavior, I had envisioned leveraging the `json.RawMessage` struct.  [The documentation](https://golang.org/pkg/encoding/json/#RawMessage) has an example not entirely dissimilar from our use case.

Something like

```golang


func (fm *funcMatcher) UnmarshalJSON(bytes []byte) error {
	raw := map[string]json.RawMessage{}
	if err := json.Unmarshal(bytes, &raw); err != nil {
		return err
	}

	for key, msg := range raw {
		switch strings.ToLower(key) {
		case "package":
			var l *literalMatcher
			// TODO handle error
			_ = json.Unmarshal(msg, &l)
			fm.Package = l
		case "receiver", "method", "et cetera":
			// similar
		default:
			// error
		}
	}
	// ...
}
```

If our struct fields ever change, everything we need to update will be here in the Unmarshal rather than spread out across field maps.  It's not a big thing, but it feels cleaner to me.

I could really go either way on it though.

internal/pkg/config/config.go
benhxy: Updated

benhxy:
> Michael covered almost all of my concerns.
> 
> Implementation-wise, I'm not a huge fan of the separation of the structs themselves and these field-maps. Back when _someone_ <shifty eye darting> introduced these methods without strict behavior, I had envisioned leveraging the `json.RawMessage` struct. [The documentation](https://golang.org/pkg/encoding/json/#RawMessage) has an example not entirely dissimilar from our use case.
> 
> Something like
> 
> ```go
> func (fm *funcMatcher) UnmarshalJSON(bytes []byte) error {
> 	raw := map[string]json.RawMessage{}
> 	if err := json.Unmarshal(bytes, &raw); err != nil {
> 		return err
> 	}
> 
> 	for key, msg := range raw {
> 		switch strings.ToLower(key) {
> 		case "package":
> 			var l *literalMatcher
> 			// TODO handle error
> 			_ = json.Unmarshal(msg, &l)
> 			fm.Package = l
> 		case "receiver", "method", "et cetera":
> 			// similar
> 		default:
> 			// error
> 		}
> 	}
> 	// ...
> }
> ```
> 
> If our struct fields ever change, everything we need to update will be here in the Unmarshal rather than spread out across field maps. It's not a big thing, but it feels cleaner to me.
> 
> I could really go either way on it though.

Haha I'll leave that change to another PR.

mlevesquedion (approved):
LGTM

PR #214, opened 2020-12-10
Refactor - Extract propagation logic from source into levee

PurelyApplied:
Currently, we perform our taint propagation logic when a new source is identified.  To enable a more modular, testable framework, the `source` analyzer should be responsible for identification of sources, and the `levee` analyzer should be responsible for taint propagation.

This PR extracts propagation logic from `source.go` to be consumed directly by `levee`.  By intent, it has not altered any analysis logic or core structuring beyond what was necessary.  There remain many small refactorings that could still be done, but I have deferred these given that I am already at +/- 400 lines.

As a point of pride, each commit in this PR's chain attempts to be atomic.  This can be helpful in review to avoid many lines of diff that result from an export or other signature changes.  Tests pass at any given commit.\*  Commits will be squashed before merge.

Changes in summary beyond migration of code:
* `source.Source.Node`, `source.Classifier`, `source.IsSourceType` have been exported.
* `Dfs` as an entry-point handles initial traversal-root logic explicitly.  (Previously, handled by `if s.node == n { ...` in `visit()`.)
* `Dfs` is explicitly called on sources in `levee`.
* Current `levee` logic is based on examining the dfs result per source.  This is a strong candidate for future refactor, to remove the need for `DFSRecord`.

----

\*`source_test.go` has been suppressed.  It's not clear to me what is the intended scope of these tests.  If you can make sense of them, I'll migrate / rewrite them for `propagation.go`.  Otherwise, our existing e2e tests are our propagation tests.


internal/pkg/levee/levee.go
```diff
@@ -61,11 +69,15 @@ func run(pass *analysis.Pass) (interface{}, error) {
 				callee := v.Call.StaticCallee()
 				switch {
 				case fieldPropagators.IsFieldPropagator(v):
-					sources = append(sources, source.New(v, conf, taggedFields))
+					newSrc := source.New(v)
+					propagationRecords[v] = propagation.Dfs(v, conf, taggedFields)
+
```
mlevesquedion, line 74: This whitespace feels extraneous.

internal/pkg/levee/propagation/propagation.go
```diff
@@ -0,0 +1,350 @@
...
+func Dfs(n ssa.Node, conf source.Classifier, taggedFields fieldtags.ResultType) DFSRecord {
+	record := DFSRecord{
+		Root:         n,
+		Marked:       make(map[ssa.Node]bool),
+		PreOrder:     nil,
+		Sanitizers:   nil,
+		Config:       conf,
+		TaggedFields: taggedFields,
+	}
+	var lastBlockVisited *ssa.BasicBlock = nil
```
mlevesquedion, line 35: `= nil` is not needed.

mlevesquedion: In fact, consider inlining this variable.

internal/pkg/levee/propagation/propagation.go
```diff
@@ -0,0 +1,350 @@
...
+	record.visitReferrers(n, maxInstrReached, lastBlockVisited)
+	record.dfs(n, maxInstrReached, lastBlockVisited, false)
+
+	return record
+}
+
+// dfs performs Depth-First-Search on the def-use graph of the input Source.
+// While traversing the graph we also look for potential sanitizers of this Source.
+// If the Source passes through a sanitizer, dfs does not continue through that Node.
+func (s *DFSRecord) dfs(n ssa.Node, maxInstrReached map[*ssa.BasicBlock]int, lastBlockVisited *ssa.BasicBlock, isReferrer bool) {
```
mlevesquedion, line 47: Nit: `s` no longer feels like an appropriate receiver name now that the full name no longer starts with an `s`.

internal/pkg/levee/propagation/propagation.go
```diff
@@ -0,0 +1,350 @@
...
+	}
+
+	return b.String()
+}
+
+// compress removes the elements from the graph that are not required by the
+// taint-propagation analysis. Concretely, only propagators, sanitizers and
+// sinks should constitute the output. Since, we already know what the source
+// is, it is also removed.
+func (s *DFSRecord) compress() []ssa.Node {
```
mlevesquedion, line 259: Out of scope: This feels like it should be removed in the near future. I believe this was originally added to support the `String` method above, which itself was used in tests?

mlevesquedion (commented):
Good refactoring! 

I think `source_test.go` can be removed altogether. All of that behavior is covered in much greater depth in the tests for `levee`.

Also: The CI is currently failing because of missing headers in propagation.go, as well as unused objects according to `staticcheck`.

internal/pkg/source/source.go
```diff
@@ -43,295 +39,37 @@ type classifier interface {
 // its referrers.
 // Source.sanitized notes sanitizer calls that sanitize this Source
```
mlevesquedion, line 40: This doc comment is out of date.

internal/pkg/levee/propagation/propagation.go
```diff
@@ -0,0 +1,350 @@
...
+
+	"github.com/google/go-flow-levee/internal/pkg/fieldtags"
+	"github.com/google/go-flow-levee/internal/pkg/sanitizer"
+	"github.com/google/go-flow-levee/internal/pkg/source"
+	"github.com/google/go-flow-levee/internal/pkg/utils"
+	"golang.org/x/tools/go/pointer"
+	"golang.org/x/tools/go/ssa"
+)
+
+type DFSRecord struct {
```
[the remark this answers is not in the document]
PurelyApplied: I went with `Propagation` since it's the name of the package and more implementation-agnostic.

internal/pkg/levee/propagation/propagation.go
```diff
@@ -0,0 +1,350 @@
...
+	Config       source.Classifier
+	TaggedFields fieldtags.ResultType
+}
+
+func Dfs(n ssa.Node, conf source.Classifier, taggedFields fieldtags.ResultType) DFSRecord {
+	record := DFSRecord{
+		Root:         n,
+		Marked:       make(map[ssa.Node]bool),
+		PreOrder:     nil,
+		Sanitizers:   nil,
```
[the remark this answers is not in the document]
PurelyApplied: Hmmm... I tend to specify to demonstrate "this is intentionally nil, not an oversight."  But you're right that it's not the Go style.  Pruned.

internal/pkg/levee/propagation/propagation.go
```diff
@@ -0,0 +1,350 @@
...
+)
+
+type DFSRecord struct {
+	Root         ssa.Node
+	Marked       map[ssa.Node]bool
+	PreOrder     []ssa.Node
+	Sanitizers   []*sanitizer.Sanitizer
+	Config       source.Classifier
+	TaggedFields fieldtags.ResultType
+}
```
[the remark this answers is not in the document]
PurelyApplied: None of them, apparently!  Definite oversight on my part, good catch.

internal/pkg/levee/levee.go
```diff
@@ -61,11 +69,15 @@ func run(pass *analysis.Pass) (interface{}, error) {
 				callee := v.Call.StaticCallee()
 				switch {
 				case fieldPropagators.IsFieldPropagator(v):
-					sources = append(sources, source.New(v, conf, taggedFields))
+					newSrc := source.New(v)
```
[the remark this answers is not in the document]
PurelyApplied: Done.

internal/pkg/levee/levee.go
```diff
@@ -61,11 +69,15 @@ func run(pass *analysis.Pass) (interface{}, error) {
...
+					propagationRecords[v] = propagation.Dfs(v, conf, taggedFields)
+
+					sources = append(sources, newSrc)
 
 				case callee != nil && conf.IsSink(utils.DecomposeFunction(v.Call.StaticCallee())):
```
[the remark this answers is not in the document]
PurelyApplied: I'll add it to the list of out of scope fixes I'd like.

internal/pkg/levee/levee.go
PurelyApplied: Trimmed.

internal/pkg/levee/levee.go
```diff
@@ -61,11 +69,15 @@ func run(pass *analysis.Pass) (interface{}, error) {
...
 					for _, s := range sources {
-						if s.HasPathTo(instr.(ssa.Node)) && !s.IsSanitizedAt(v) {
+						record := propagationRecords[s.Node]
+						if record.HasPathTo(instr.(ssa.Node)) && !record.IsSanitizedAt(v) {
```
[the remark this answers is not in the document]
PurelyApplied: How about `prop`, given `propagation` would collide with the package name?

internal/pkg/levee/propagation/propagation.go
PurelyApplied: Done.

internal/pkg/levee/propagation/propagation.go
PurelyApplied: I think it's in scope to prune this, since it's being moved anyway.

internal/pkg/source/source.go
PurelyApplied: Trimmed.

internal/pkg/levee/levee.go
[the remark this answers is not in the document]
mlevesquedion: SG

internal/pkg/levee/propagation/propagation.go
PurelyApplied: I had been under the impression that the dfs would mutate that value, but that appears to be incorrect.  `nil` passed directly to consumers.

internal/pkg/levee/propagation/propagation.go
```diff
@@ -0,0 +1,350 @@
+package propagation
```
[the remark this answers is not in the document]
mlevesquedion: I think having a separate package is fine. That said, IMO this code should live in the `levee` package (in its own file, `propagation.go`, like it is now). My main argument there is that the `propagation` package is exclusively tested via the `levee` package's tests. This is similar to the current situation where most of the logic in `source` is tested via `levee`.

internal/pkg/levee/levee.go
```diff
@@ -45,6 +46,13 @@ func run(pass *analysis.Pass) (interface{}, error) {
 	fieldPropagators := pass.ResultOf[fieldpropagator.Analyzer].(fieldpropagator.ResultType)
 	taggedFields := pass.ResultOf[fieldtags.Analyzer].(fieldtags.ResultType)
 
+	propagationRecords := map[ssa.Node]propagation.Propagation{}
```
mlevesquedion, line 49: Nit: `Records` feels like it's no longer needed. Just `propagations` maybe?

mlevesquedion (approved):
LGTM.

PR #218, opened 2020-12-14
Freshen source package docs

mlevesquedion:
Much of the documentation in the `source` package is unclear or lacking detail. This PR aims to make the documentation more complete.

I will be giving the `levee` package a similar treatment in a future PR.

### DIFF
--- a/internal/pkg/source/source.go
+++ b/internal/pkg/source/source.go
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// Package source contains the logic related to the concept of the source which may be tainted.
+// Package source can be used to identify SSA values that are Sources.
 package source
 
 import (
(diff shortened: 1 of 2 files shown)


vinayakankugoyal (approved):

vinayakankugoyal:
These comments are invaluable to a newcomer to go-flow-levee. Thanks for adding the package docs.

internal/pkg/source/source.go
```diff
@@ -26,6 +26,8 @@ import (
 	"golang.org/x/tools/go/ssa"
 )
 
+// A Classifier is used to identify Objects of interest,
+// such as Sources, Sanitizers, and Sinks.
```
PurelyApplied, line 30: Out of scope, but this `config` interface really feels like it belongs with `package config`.  Its inclusion here feels vestigial.

PurelyApplied (approved):

internal/pkg/source/source.go
mlevesquedion: Agreed. While we're on this topic, I'm not convinced that `Classifier` is needed right now. So far I don't believe it has shielded us from any changes to `config.Config`, and in fact when we changed `config.Config`'s interface we had to change `Classifier` as well. I think we should consider consuming `config.Config` directly.

PurelyApplied: Also a good point.  I think it may had been introduced to accommodate some tests, but may have only been used by the `source_test`s that were removed in my recent refactor.

PR #220, opened 2020-12-14
Freshen levee package docs

mlevesquedion:
This PR aims to improve the documentation for the `levee` and `propagation` packages. Specifically, it adds a few documentation comments and removes/rectifies a few incorrect pieces of documentation.

### DIFF
--- a/internal/pkg/levee/propagation/propagation.go
+++ b/internal/pkg/levee/propagation/propagation.go
@@ -57,6 +57,16 @@ func Dfs(n ssa.Node, conf source.Classifier, taggedFields fieldtags.ResultType) 
 	return record
 }
 
+// dfs performs a depth-first search of the graph formed by SSA Referrers and
+// Operands relationships. Along the way, visited nodes are marked and stored
+// in a slice which captures the visitation order. Sanitizers are also recorded.
+// maxInstrReached and lastBlockVisited are used to give the traversal some
+// degree of flow sensitivity. Specifically:
+// - maxInstrReached records the highest index of an instruction visited
+//   in each block. This is used to avoid visiting past instructions, e.g.
+//   a call to a sink where the argument was tainted after the call happened.
+// - lastBlockVisited is used to determine whether the next instruction to visit
+//   can be reached from the current instruction.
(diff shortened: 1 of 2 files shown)


internal/pkg/levee/propagation/propagation.go
```diff
@@ -51,9 +57,6 @@ func Dfs(n ssa.Node, conf source.Classifier, taggedFields fieldtags.ResultType)
 	return record
 }
 
-// dfs performs Depth-First-Search on the def-use graph of the input Source.
```
vinayakankugoyal, line 54: Why remove this comment? I think the second part of the comment was very useful for a newb like me. Did the behavior mentioned in the comment change?

vinayakankugoyal (approved):

internal/pkg/levee/propagation/propagation.go
mlevesquedion: The last sentence is actually incorrect. When the traversal encounters a Sanitizer, it merely records that the sanitizer has been reached. Sanitizers are not otherwise treated differently from other calls.

Removing the comment is the wrong thing to do here, though. I have rewritten the comment instead, thank you for your input! WDYT about this new comment?

vinayakankugoyal: LGTM! Thanks for adding great documentation to this method.

vinayakankugoyal (approved):

internal/pkg/levee/propagation/propagation.go
```diff
@@ -36,6 +40,8 @@ type Propagation struct {
 	taggedFields fieldtags.ResultType
 }
 
+// Dfs performs a depth-first search of the graph formed by SSA Referrers and
+// Operands relationships.
```
PurelyApplied, line 44: Consider adding `beginning at the root node n` to the doc comment to distinguish this exported `Dfs` as the entrypoint for a given root vs the recursive step `dfs` below.

PurelyApplied (approved):

internal/pkg/levee/propagation/propagation.go
mlevesquedion: Done.

PR #223, opened 2020-12-14
Identify field-propagators in source analyzer.

PurelyApplied:
Minor additional refactor that was not in scope for #214.  This moves source identification currently in `levee` as the result of simple getters of source fields into the source-identifying analyzer.

### DIFF
--- a/internal/pkg/source/source.go
+++ b/internal/pkg/source/source.go
@@ -119,30 +119,19 @@ func sourcesFromBlocks(fn *ssa.Function, conf *config.Config, taggedFields field
 	var sources []*Source
 	for _, b := range fn.Blocks {
 		for _, instr := range b.Instrs {
-			// This type switch is used to catch instructions that could produce sources.
-			// All instructions that do not match one of the cases will hit the "default"
-			// and they will not be examined any further.
+			var isSource bool
+
+			// Identify those instructions which produce a source.
 			switch v := instr.(type) {
-			// drop anything that doesn't match one of the following cases
-			default:
-				continue
 
-			// Do not mark allocation values returned by sanitizers as new sources
+			// Values produced by sanitizers
 			case *ssa.Alloc:
-				if isProducedBySanitizer(v, conf) {
-					continue
-				}
+				isSource = !isProducedBySanitizer(v, conf) && IsSourceType(conf, taggedFields, v.Type())
 
-			// Do not mark values returned by sanitizers as new sources.
-			// Do mark as new sources values returned by field propagators.
+			// Return values produced by sanitizers cannot be sources,
 			case *ssa.Call:
-				if isProducedBySanitizer(v, conf) {
-					continue
-				}
-
-				if propagators.IsFieldPropagator(v) {
-					sources = append(sources, New(v))
-				}
+				isSource = !isProducedBySanitizer(v, conf) &&
+					(propagators.IsFieldPropagator(v) || IsSourceType(conf, taggedFields, v.Type()))
 
 			// An Extract is used to obtain a value from an instruction that returns multiple values.
 			// If the extracted value is a Pointer to a Source, it won't have an Alloc, so we need to
@@ -150,27 +139,25 @@ func sourcesFromBlocks(fn *ssa.Function, conf *config.Config, taggedFields field
 			case *ssa.Extract:
 				t := v.Tuple.Type().(*types.Tuple).At(v.Index).Type()
 				if _, ok := t.(*types.Pointer); ok && IsSourceType(conf, taggedFields, t) {
-					sources = append(sources, New(v))
+					isSource = true
 				}
-				continue
 
-			// value received from chan
+			// Values received from <-chan operations can be sources based on type
 			case *ssa.UnOp:
-				// not a <-chan operation
-				if v.Op != token.ARROW {
-					continue
+				isSource = v.Op == token.ARROW && IsSourceType(conf, taggedFields, v.Type())
+
+			// Identify sources by type in field access, element access of a collection, or map and chan collection creation.
+			// Array and slice collections are created via Alloc.  TODO: Then why is there a ssa.MakeSlice?
+			case *ssa.Field, *ssa.FieldAddr,
+				*ssa.Index, *ssa.IndexAddr, *ssa.Lookup,
+				*ssa.MakeMap, *ssa.MakeChan:
+				if v := instr.(ssa.Value); IsSourceType(conf, taggedFields, v.Type()) {
+					isSource = true
 				}
-
-			// value obtained through a field or an index operation
-			case *ssa.Field, *ssa.FieldAddr, *ssa.Index, *ssa.IndexAddr, *ssa.Lookup:
-
-			// chan or map value (arrays and slices are created using Allocs)
-			case *ssa.MakeMap, *ssa.MakeChan:
 			}
 
-			// all of the instructions that the switch lets through are values as per ssa/doc.go
-			if v := instr.(ssa.Value); IsSourceType(conf, taggedFields, v.Type()) {
-				sources = append(sources, New(v.(ssa.Node)))
+			if isSource {
+				sources = append(sources, New(instr.(ssa.Node)))
 			}
 		}
 	}
--- a/internal/pkg/source/source.go
+++ b/internal/pkg/source/source.go
@@ -127,14 +127,14 @@ func sourcesFromBlocks(fn *ssa.Function, conf *config.Config, taggedFields field
 			default:
 				continue
 
-			// Do not mark allocation values returned by sanitizers as new sources
+			// Values produced by sanitizers are not sources.
 			case *ssa.Alloc:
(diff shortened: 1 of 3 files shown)


internal/pkg/source/source.go
```diff
@@ -126,12 +127,23 @@ func sourcesFromBlocks(fn *ssa.Function, conf *config.Config, taggedFields field
...
 				}
 
+			// Do not mark values returned by sanitizers as new sources.
+			// Do mark as new sources values returned by field propagators.
+			case *ssa.Call:
+				if isProducedBySanitizer(v, conf) {
+					continue
+				}
+
+				if propagators.IsFieldPropagator(v) {
```
mlevesquedion, line 143: Nit: consider adding a `continue` instruction at the end of this `if`. Indeed, if this call is identified as a field propagator, then it won't be identified as a source below. (For that reason,  the `continue` isn't actually necessary, but I believe it may help with readability.) It may be a good idea to add a comment explaining all of this.)

internal/pkg/source/source.go
```diff
@@ -126,12 +127,23 @@ func sourcesFromBlocks(fn *ssa.Function, conf *config.Config, taggedFields field
 			default:
 				continue
 
-			// local variable or value returned from a call
-			case *ssa.Alloc, *ssa.Call:
-				if isProducedBySanitizer(v.(ssa.Value), conf) {
+			// Do not mark allocation values returned by sanitizers as new sources
```
mlevesquedion, line 130: Nit: I think this comment would be more helpful if it were a bit higher level, e.g. `values produced by sanitizers cannot be sources`.

mlevesquedion (commented):
Thank you for making this change! Just earlier today I was discussing with @vinayakankugoyal 
 how this is one of the needlessly confusing parts of the codebase. No more.

Mostly LGTM, just a few nits. I'm fairly confident that this code is equivalent to what we have on master, I just want to take a closer look tomorrow to make absolutely sure.

internal/pkg/source/source.go
```diff
@@ -114,7 +115,7 @@ func sourcesFromClosures(fn *ssa.Function, conf *config.Config, taggedFields fie
 }
 
 // sourcesFromBlocks finds Source values created by instructions within a function's body.
-func sourcesFromBlocks(fn *ssa.Function, conf *config.Config, taggedFields fieldtags.ResultType) []*Source {
+func sourcesFromBlocks(fn *ssa.Function, conf *config.Config, taggedFields fieldtags.ResultType, propagators fieldpropagator.ResultType) []*Source {
```
mlevesquedion, line 118: Some of these signatures are starting to get heavy. It's also kind of annoying to have to pipe all of these arguments down several calls. I think it would make sense to bundle `conf`, `taggedFields` and `propagators` into some kind of unifying abstraction. Maybe something like... `Classifier`? :grimacing:

I think a good way to do it would be to embed `config.Config`, `fieldtags.ResultType` and `fieldpropagator.ResultType` into a new struct type. There may be name clashes that we would need to resolve, though.

I think this change could be done in a future PR.

WDYT?

PurelyApplied: Yeah... I think "upstream analyzer stuff" could be a good unit of abstraction - config + any previous analyzer `Result`s.  Of course, at that point we could just pass config and `pass` to every function.

On the one hand, it feels like a code-smell reliance on globals.  I suppose globals aren't inherently evil, particularly when they're static, so this might be an unwarranted knee-jerk aversion to it.

But yes, a future refactor PR might be a better place for it.

internal/pkg/source/source.go
PurelyApplied: Hmmmm... I think this loop control has gotten a little out of hand.  The fact that the switch needed a comment to explain its flow (which of course I failed to read) might be evidence enough of that.

What do you think of the this most recent commit?  Switch to flag if the instruction produces a source, and add the node if it does.  Short-circuiting is now handled by boolean evaluation rather than a `continue` jump.

Upon this closer inspection, I also noticed a comment which seems inaccurate...  See the TODO.

This is creeped past the PR intent, so I can revert all this and split it into a different PR if you prefer.

internal/pkg/source/source.go
```diff
@@ -114,51 +115,49 @@ func sourcesFromClosures(fn *ssa.Function, conf *config.Config, taggedFields fie
...
-			// value received from chan
+			// Values received from <-chan operations can be sources based on type
 			case *ssa.UnOp:
-				// not a <-chan operation
-				if v.Op != token.ARROW {
-					continue
+				isSource = v.Op == token.ARROW && IsSourceType(conf, taggedFields, v.Type())
+
+			// Identify sources by type in field access, element access of a collection, or map and chan collection creation.
+			// Array and slice collections are created via Alloc.  TODO: Then why is there a ssa.MakeSlice?
```
PurelyApplied, line 150: The original comment:
```golang
// chan or map value (arrays and slices are created using Allocs)
case *ssa.MakeMap, *ssa.MakeChan:	
```
seems incorrect, or at least incomplete.  There is a `ssa.MakeSlice` instruction.  An array will necessarily have a corresponding alloc for the underlying array, but is that guarantee present for a `MakeSlice`?

_Well_ outside the scope of this PR now, but I figured I'd bring it up.

internal/pkg/source/source.go
```diff
@@ -114,51 +115,49 @@ func sourcesFromClosures(fn *ssa.Function, conf *config.Config, taggedFields fie
...
 // sourcesFromBlocks finds Source values created by instructions within a function's body.
-func sourcesFromBlocks(fn *ssa.Function, conf *config.Config, taggedFields fieldtags.ResultType) []*Source {
+func sourcesFromBlocks(fn *ssa.Function, conf *config.Config, taggedFields fieldtags.ResultType, propagators fieldpropagator.ResultType) []*Source {
 	var sources []*Source
 	for _, b := range fn.Blocks {
 		for _, instr := range b.Instrs {
-			// This type switch is used to catch instructions that could produce sources.
-			// All instructions that do not match one of the cases will hit the "default"
-			// and they will not be examined any further.
+			var isSource bool
```
PurelyApplied, line 122: If you think it reads better, this entire switch could now be extracted to a `func isSourceInstruction(...)`, which would collapse this loop to something very succinct.

internal/pkg/source/source.go
mlevesquedion: Good catch. Judging from the code in `ssa/builder.go`, it looks like `MakeSlice` is produced when the length or capacity in a call to `make` is a variable. This is a blind spot in our testing. I'm not a huge fan of the naked TODO, but if you intend to follow this PR up with a fix right away, I suppose it's fine. If you don't have the cycles, I can make the fix.

internal/pkg/source/source.go
mlevesquedion: I do think that would read better. If you want to extract that function, go ahead.

I think the code was mostly fine, I'm not sure about this `isSource` variable. You're calling `IsSourceType` in many places now, instead of just one earlier. This makes it harder to tell, at a glance, what the exclusions are for a given instruction type, because they all have this call. If you extract a function though you may be able to use early returns to get rid of that `isSource` variable.

internal/pkg/source/source.go
mlevesquedion: I have replied on the other threads.

I think I would prefer the revert + split, thank you for suggesting that. I agree that this PR is doing a bit too much now.

internal/pkg/source/source.go
PurelyApplied: > I'm not a huge fan of the naked TODO,
Me either.  Even annotated ones bother be a bit.  I personally used them as a `DO NOT SUBMIT; TALK ABOUT THIS FIRST` marker.

As for the actual issue, it's another out-of-scope thing.  I'll chase it up with an Issue.

internal/pkg/source/source.go
PurelyApplied: Good call.  This PR has been rolled back to the thing version, and the `continue` has been added per this thread's first comment.  To avoid too many merge conflicts, I'll open (separate) PRs for consolidating arguments and for this loop control after this PR is merged.

internal/pkg/source/source.go
PurelyApplied: For a future PR, but once the body is a separate method, it'll just read 

```golang
for _, b := range fn.Blocks {
    if isSourceInstruction(...) {
      sources = append(sources, instr)
    }
  }
}
```

So the variable name is dropped altogether.

Of course, between now and the actual PR, I'll need to come up with a better name for the method...  it's naming all the way down, huh?

internal/pkg/source/source.go
mlevesquedion: There seems to be an issue with your markdown here, maybe a missing line between the quote and your reply?

> I personally used them as a DO NOT SUBMIT

+1, good to know. I also do that.

mlevesquedion (approved):
LGTM

PR #225, opened 2020-12-15
Refactor propogation package's naming.

vinayakankugoyal:
Dfs is an implementation detail of the propagation package, what it really does it propagate the taint. So I am updating the names of the functions to reflect that. The logic of changing the name is as follow:

1. replace word "visit" with "taint" in function names (here taint is used as a verb: contaminate or pollute (something).)
2. replace word "dfs" with "propagateTaint" in function names (here taint is used as a noun: a trace of a bad or undesirable quality or substance.)


internal/pkg/levee/levee.go
```diff
@@ -50,7 +50,7 @@ func run(pass *analysis.Pass) (interface{}, error) {
 
 	for _, sources := range sourcesMap {
 		for _, s := range sources {
-			propagations[s.Node] = propagation.Dfs(s.Node, conf, taggedFields)
+			propagations[s.Node] = propagation.PropagateTaint(s.Node, conf, taggedFields)
```
mlevesquedion, line 53: I think this name is better than `Dfs`, but it could still be improved. `propagation.PropagateTaint` as a whole is a bit repetitive. From the [Go blog](https://blog.golang.org/package-names):

> Avoid stutter. Since client code uses the package name as a prefix when referring to the package contents, the names for those contents need not repeat the package name. The HTTP server provided by the http package is called Server, not HTTPServer. Client code refers to this type as http.Server, so there is no ambiguity.

internal/pkg/levee/propagation/propagation.go
```diff
@@ -43,22 +43,22 @@ type Propagation struct {
 
 // Dfs performs a depth-first search of the graph formed by SSA Referrers and
 // Operands relationships, beginning at the given root node.
-func Dfs(n ssa.Node, conf *config.Config, taggedFields fieldtags.ResultType) Propagation {
+func PropagateTaint(n ssa.Node, conf *config.Config, taggedFields fieldtags.ResultType) Propagation {
 	record := Propagation{
```
mlevesquedion, line 47: This isn't directly related to your changes, but since you're renaming things, consider renaming this variable to `prop` instead of `record`.

internal/pkg/levee/propagation/propagation.go
```diff
@@ -43,22 +43,22 @@ type Propagation struct {
 
 // Dfs performs a depth-first search of the graph formed by SSA Referrers and
```
mlevesquedion, line 44: The name in the comment no longer matches the function's name.

internal/pkg/levee/propagation/propagation.go
```diff
@@ -130,7 +130,7 @@ func (prop *Propagation) shouldNotVisit(n ssa.Node, maxInstrReached map[*ssa.Bas
 	return false
 }
 
-func (prop *Propagation) visit(n ssa.Node, maxInstrReached map[*ssa.BasicBlock]int, lastBlockVisited *ssa.BasicBlock) {
+func (prop *Propagation) taint(n ssa.Node, maxInstrReached map[*ssa.BasicBlock]int, lastBlockVisited *ssa.BasicBlock) {
```
mlevesquedion, line 133: I think this should be called something like `taintNeighbors`.

internal/pkg/levee/levee.go
mlevesquedion: (See also the internal `propagateTaint` function.)

mlevesquedion (commented):
I think this will help make the code's intent clearer. Good stuff!

I have just a couple of nits, though since this is a pretty significant change, I will likely need to take another look at this.

internal/pkg/levee/levee.go
vinayakankugoyal: Any suggestion for the name?

internal/pkg/levee/propagation/propagation.go
vinayakankugoyal: Thank for pointing this out, Ill fix this once we decide on the names of the functions of this package.

internal/pkg/levee/levee.go
mlevesquedion: Just `Taint` would work IMO. I'd like to hear other reviewers' opinions on this though.

PurelyApplied: I am not to be trusted on such things as naming.

That said, `Taint` feels like what affects the neighbors.  This is called on the source.  To me, that seems like `propagation.From(...)`, if we're going for maximally succinct.

While the exported `Dfs` and the recursive-step `dfs` names didn't seem like a conflict to me, I'm not sure I like an exported `Taint` and an unexported `taint`.  But that can also be solved by Michael's suggested for `taint -> taintNeighbors`.  So `Taint` here would probably be fine.

internal/pkg/levee/propagation/propagation.go
vinayakankugoyal: Done.

internal/pkg/levee/propagation/propagation.go
vinayakankugoyal: Done.

internal/pkg/levee/propagation/propagation.go
vinayakankugoyal: Done.

internal/pkg/levee/levee.go
vinayakankugoyal: Updated to 
Dfs -> Taint
dfs -> taint
visit -> taintNeighbors

PurelyApplied (approved):
The merge conflicts with `master` will never end.

Some of this is going to collide with #250, too.

PR #227, opened 2020-12-15
Debug: Add missing parameter edges

mlevesquedion:
This PR adds missing referrers edges for parameters to the debug output. Missing edges can make it hard to understand exactly how/why a given traversal is working.

---

Before:

image

After:

image

### DIFF
--- a/internal/pkg/debug/render/dot.go
+++ b/internal/pkg/debug/render/dot.go
@@ -56,6 +56,10 @@ func (r *renderer) writeSubgraphs() {
 }
 
 func (r *renderer) writeEdges() {
+	for _, p := range r.f.Params {
+		r.writeReferrers(p)
+	}
+
 	for _, b := range r.f.Blocks {
 		for _, i := range b.Instrs {
 			r.writeReferrers(i.(ssa.Node))
--- a/internal/pkg/debug/render/testdata/TestParams.dot
+++ b/internal/pkg/debug/render/testdata/TestParams.dot
@@ -16,6 +16,9 @@ digraph {
 		"t8 = fmt.Println(t7...)\n(Call)" [shape=rectangle];
 		"return\n(Return)" [shape=diamond];
 	}
+	"a\n(Parameter)" -> "t2 = make interface{} <- int (a)\n(MakeInterface)" [color=red];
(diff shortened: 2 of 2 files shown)


PurelyApplied:
Well out of scope for this PR:

I'm a little concerned about drift between our existing propagation logic now in `levee/propagate` and what gets rendered by `debug`.  We've talked elsewhere about how a struct shouldn't necessarily taint all its fields, and I worry allowing the tools to own their own neighbor logic will put us in a position of looking at a graph that isn't representative of what our actual propagation is doing.  I suppose there is value in seeing the full set of referrers/operands, especially if we could see how that differed from how taint will propagate.

What do you think of a pair of PRs that...

* Refactors propagation to separate `<get neighbors>` from `<visit/taint neighbors>`
* Export the necessary `<get neighbors>` methods from `propagate` for direct use in `debug`.

mlevesquedion:
I understand your concerns. I think it depends on what we expect these debugging tools to provide. In my mind the main intent of the debugging output is to provide easy access to:
1. The SSA code
2. The SSA graph made up of Operands/Referrers relationships

These are agnostic of any analysis implementation. They are merely a way to see the SSA. When I was getting started on the project, being able to see the SSA was tremendously helpful for my understanding, and it is still very helpful for debugging.

With that in mind, there is already a significant drift between the debugging output and what our analyzer is doing. While in very early days it was sufficient to look at the graph to understand the analyzer's behavior (because it was pretty much just a DFS on that graph), nowadays the debugging output is not enough. One also has to be familiar with the implementation.

I believe in order to bridge that gap we would need to add taint information to the graph. Currently, one has to identify source and perform the propagation mentally, but if e.g. we gave special colors to nodes of interest (e.g. sources could be green, sinks could be blue, etc.) to nodes that are reachable from a source, that would make it much easier to understand what the analyzer is doing.

I am a bit wary of tying the debugging code to the propagation code. Certainly, we want our debugging helpers to be maximally useful, but we have to make sure that they are also worth their weight.

All that to say, I am not sure about those 2 hypothetical PR's.

PurelyApplied:
That works for me.  I guess I was really just trying to make sure the scope of the tool was well-defined.  Visualization of the problem space rather than the solution is absolutely valuable.

PurelyApplied (approved):

Issue #231, opened 2020-12-16
[FALSE NEGATIVE]: go panic(source)


vinayakankugoyal:
levee treats panic as a sink but does create a report for `go panic(source)`.

PR #232, opened 2020-12-16
Remove title suggestions in issue templates

mlevesquedion:
Looking at #231, I realized that the title in the issue is redundant with the labels. Besides being redundant, it makes for a noisy title, which unnecessarily stands out from the other issues.


vinayakankugoyal (approved):

PurelyApplied (approved):
Are the keys required?  With `title` and `assignees` being empty, I would think they should just be dropped.

PR #233, opened 2020-12-16
Adding a script to go-flow-levee to check for new findings compared to master version of levee.

vinayakankugoyal:
I did this step several times and so adding a script that I was using for this, to save others some time.

- [X] Running against a large codebase such as [Kubernetes](https://github.com/kubernetes/kubernetes) does not error out. (See [DEVELOPING.md](https://github.com/google/go-flow-levee/blob/master/DEVELOPING.md) for instructions on how to do that.)


new-findings.sh
```diff
@@ -0,0 +1,68 @@
...
+MASTER_FINDINGS_SORTED=$(mktemp)
+BRANCH_FINDINGS_SORTED=$(mktemp)
+
+sort $MASTER_FINDINGS > $MASTER_FINDINGS_SORTED
+rm $MASTER_FINDINGS
+
+sort $BRANCH_FINDINGS > $BRANCH_FINDINGS_SORTED
+rm $BRANCH_FINDINGS
+
+diff $MASTER_FINDINGS_SORTED $BRANCH_FINDINGS_SORTED
```
mlevesquedion, line 68: Prefer: `diff <(sort "$MASTER_FINDINGS") <(sort "$BRANCH_FINDINGS")`

new-findings.sh
```diff
@@ -0,0 +1,68 @@
+#!/bin/bash
+
```
mlevesquedion, line 2: Consider using `set -x` so that each command will be printed as the script runs. (This would avoid the need for most of the `echo`s).

new-findings.sh
```diff
@@ -0,0 +1,68 @@
...
+GO_SRC_PATH="$(go env GOPATH)/src"
+CURRENT_BRANCH=$(git branch --show-current)
```
mlevesquedion, line 4: Nit: I think `FEATURE_BRANCH` would be a more accurate name

new-findings.sh
```diff
@@ -0,0 +1,68 @@
...
+
+cd $GO_SRC_PATH/k8s.io/kubernetes
+
+CFG_PATH="$(realpath ./hack/testdata/levee/levee-config.yaml)"
+
+make clean
+
+BRANCH_FINDINGS=$(mktemp)
+
+make vet WHAT="-vettool=$(which levee) -config=$CFG_PATH" 2> $BRANCH_FINDINGS
```
mlevesquedion, line 55: This code is duplicated from the master run above. Consider extracting a function.

new-findings.sh
```diff
@@ -0,0 +1,68 @@
+#!/bin/bash
```
mlevesquedion, line 1: Could you run `shellcheck` against this script and fix any issues?

new-findings.sh
```diff
@@ -0,0 +1,68 @@
...
+
+go install
+
+echo "Running levee(master) against k8s"
+
+cd $GO_SRC_PATH/k8s.io/kubernetes
+
+CFG_PATH="$(realpath ./hack/testdata/levee/levee-config.yaml)"
+
+make clean
```
PurelyApplied, line 27: Do you have evidence that `make clean` is necessary?  I would think / hope that it isn't, and it will bloat your runtime.  Ditto below.

new-findings.sh
```diff
@@ -0,0 +1,68 @@
...
+
+MASTER_FINDINGS=$(mktemp)
+
+make vet WHAT="-vettool=$(which levee) -config=$CFG_PATH" 2> $MASTER_FINDINGS
```
PurelyApplied, line 31: Absent a `make clean`, consider using `make -C <kubernetes> ...` rather than an explicit `cd`.  Ditto below.

new-findings.sh
PurelyApplied, line 1: I'm not sure this script should live in the root.  I'd prefer introducing a `scripts/` or `hack/` folder personally.

new-findings.sh
```diff
@@ -0,0 +1,68 @@
+#!/bin/bash
+
+GO_SRC_PATH="$(go env GOPATH)/src"
+CURRENT_BRANCH=$(git branch --show-current)
+
+if [[ ${CURRENT_BRANCH} == "master" ]]; then
```
PurelyApplied, line 6: Quote wrap `${CURRENT_BRANCH}`.  Also check for non-empty.

new-findings.sh
```diff
@@ -0,0 +1,68 @@
+#!/bin/bash
+
+GO_SRC_PATH="$(go env GOPATH)/src"
+CURRENT_BRANCH=$(git branch --show-current)
+
```
PurelyApplied, line 5: Execution is brittle to invocation location.  Use `cd $(dirname $0)` to resolve to script dir.

new-findings.sh
PurelyApplied, line 2: Many of these commands could error out and should interrupt the script.  Add `set -e` to halt on first failure.  Wrap `make vet` commands with `set +e ; make ; set -e` to permit diagnostic reporting.

new-findings.sh
```diff
@@ -0,0 +1,68 @@
...
+CFG_PATH="$(realpath ./hack/testdata/levee/levee-config.yaml)"
+
+make clean
+
+MASTER_FINDINGS=$(mktemp)
+
+make vet WHAT="-vettool=$(which levee) -config=$CFG_PATH" 2> $MASTER_FINDINGS
+
+cd $GO_SRC_PATH/go-flow-levee
```
PurelyApplied, line 33: My own checkout is not on the go src path.  As above, capture local directory with `dirname $0` to resolve `levee`'s root directory.

new-findings.sh
```diff
@@ -0,0 +1,68 @@
...
+
+echo "Switching to " $CURRENT_BRANCH
+
+git checkout $CURRENT_BRANCH
+
+echo "Installing levee from" $CURRENT_BRANCH
+
+cd cmd/levee
+
+go install
```
PurelyApplied, line 43: Consider using `go build -o` rather than `go install`.

PurelyApplied (commented):
I am mildly opposed to these sort of scripts in general.  They tend to be very environment dependent.  For instance, I see the following implicit assumptions in this script:

* Local `master` branch mirrors origin's
* Current `levee` branch is clean (else `git checkout` can fail)
* Current `kubernetes` branch is buildable
* Invocation is made at `levee`'s root.
* User prefers `go install` over `go build`
* Local `levee` checkout is at the Go src path.  My own is not.
* Kubernetes checkout is at the Go src path.  While this may be required for `make` to behave properly, it's still a point of potential failure.

That said, I'm not entirely opposed to introducing a useful script.  I'd just like to see defense against some of those assumptions.

new-findings.sh
vinayakankugoyal: Yup, it is if you don't do that before each `make vet ...` there is no output.

mlevesquedion: I'm not convinced that it will *systematically* mess up the following `make vet`.

The impact on runtime should be minimal though.

new-findings.sh
mlevesquedion: I would add that this script should be referenced in the developer docs, and maybe it should replace the existing script in `DEVELOPING.md`

new-findings.sh
vinayakankugoyal: Done.

new-findings.sh
vinayakankugoyal: Done.

new-findings.sh
vinayakankugoyal: Done.

new-findings.sh
vinayakankugoyal: Done.

new-findings.sh
vinayakankugoyal: Done.

new-findings.sh
vinayakankugoyal: Done.

new-findings.sh
vinayakankugoyal: Done.

new-findings.sh
vinayakankugoyal: Removed all explicit cd commands in favor of make -C. Thanks!

new-findings.sh
vinayakankugoyal: Done.

new-findings.sh
vinayakankugoyal: Done.

new-findings.sh
vinayakankugoyal: Done.

PurelyApplied (approved):
This feels a lot better to me.  👍

DEVELOPING.md
```diff
@@ -4,25 +4,14 @@
...
-```bash
-#!/bin/bash
-
-set -eux
-
-# set these to appropriate values for your setup
-# GO_FLOW_LEVEE_PATH=path to the root of the go-flow-levee repository on your machine
-# K8S_PATH="$GOPATH"/src/k8s.io/kubernetes
+Note: The script makes the following assumptions
+1. k8s repo is at $(GOPATH)/src/k8s.io/kuberenetes
```
mlevesquedion, line 10: A few nits:
* kube*rn*etes
* $(GOPATH) ... should probably be in backticks
* I think a full sentence would read better, e.g. "The kubernetes repository on your machine is at ..."

DEVELOPING.md
```diff
@@ -4,25 +4,14 @@
...
+2. You are running the script from withing your go-flow-levee directory and have your feature branch checked out.
```
mlevesquedion, line 11: Nit: within_

DEVELOPING.md
```diff
@@ -4,25 +4,14 @@
...
-cd "$GO_FLOW_LEVEE_PATH"/cmd/levee
-go install
-
-cd "$K8S_PATH"
-git checkout master
-git pull
-CFG_PATH="$(realpath ./hack/testdata/levee/levee-config.yaml)"
-make vet WHAT="-vettool=$(which levee) -config=$CFG_PATH"
+```bash
+./hach/new-findings.sh
```
mlevesquedion, line 14: hack

hack/new-findings.sh
```diff
@@ -0,0 +1,35 @@
+#!/bin/bash
```
mlevesquedion, line 1: Nit: `check_kubernetes.sh` or `verify_kubernetes.sh` would be a more accurate name.

DEVELOPING.md
vinayakankugoyal: Done.

DEVELOPING.md
vinayakankugoyal: Done.

hack/new-findings.sh
vinayakankugoyal: switched to verify-kubernetes.sh

DEVELOPING.md
vinayakankugoyal: Done.

PR #234, opened 2020-12-17
Refactor source/propagations in levee

mlevesquedion:
Small refactoring to improve the readability of this code.

* It is better to create a mapping from sources to propagations, because that is what the client code (lower down in the method) actually needs.
* It is better to create one mapping per function, since each source/propagation appears in only one function


internal/pkg/levee/levee.go
```diff
@@ -41,32 +41,29 @@ func run(pass *analysis.Pass) (interface{}, error) {
 	if err != nil {
 		return nil, err
 	}
-	sourcesMap := pass.ResultOf[source.Analyzer].(source.ResultType)
+	fnSources := pass.ResultOf[source.Analyzer].(source.ResultType)
```
vinayakankugoyal, line 44: Does fn mean function here?

internal/pkg/levee/levee.go
```diff
@@ -75,11 +72,10 @@ func run(pass *analysis.Pass) (interface{}, error) {
 	return nil, nil
 }
 
-func reportSourcesReachingSink(pass *analysis.Pass, sources []*source.Source, instr ssa.Instruction, propagations map[ssa.Node]propagation.Propagation) {
-	for _, s := range sources {
-		prop := propagations[s.Node]
+func reportSourcesReachingSink(pass *analysis.Pass, instr ssa.Instruction, propagations map[*source.Source]propagation.Propagation) {
```
vinayakankugoyal, line 75: should this be reportSourcesReachingASink?

internal/pkg/levee/levee.go
```diff
@@ -75,11 +72,10 @@ func run(pass *analysis.Pass) (interface{}, error) {
...
+	for source, prop := range propagations {
 		if prop.HasPathTo(instr.(ssa.Node)) && !prop.IsSanitizedAt(instr) {
```
vinayakankugoyal, line 77: I personally dont like the fact we need to do 2 function calls here. We should change this to 1 function call like:

prop.HasUnsanitizedPathTo(instr)?

mlevesquedion: We think alike ;)

I actually made this exact comment on one of your recent PR's : https://github.com/google/go-flow-levee/pull/226#discussion_r544438336

I think this change is out of scope for this PR. I was planning to do it in a follow-up.

internal/pkg/levee/levee.go
mlevesquedion: Yes. Renamed to `funcSources` for clarity.

internal/pkg/levee/levee.go
mlevesquedion: I personnally think `ReachingSink` is fine, to me having the `A` makes this name slightly less readable.

mlevesquedion: Looking at the signature more closely prompted me to make the following changes:
* `instr` --> `sink`
* Swapped the last two arguments so that the order of arguments mirrors the order in the name (sources before sink)

vinayakankugoyal (approved):

internal/pkg/levee/levee.go
mlevesquedion: PR here: #242

internal/pkg/levee/levee.go
```diff
@@ -41,32 +41,29 @@ func run(pass *analysis.Pass) (interface{}, error) {
...
 	}
-	sourcesMap := pass.ResultOf[source.Analyzer].(source.ResultType)
+	funcSources := pass.ResultOf[source.Analyzer].(source.ResultType)
 	taggedFields := pass.ResultOf[fieldtags.Analyzer].(fieldtags.ResultType)
 
-	propagations := map[ssa.Node]propagation.Propagation{}
-
-	for _, sources := range sourcesMap {
+	for fn, sources := range funcSources {
+		propagations := map[*source.Source]propagation.Propagation{}
```
PurelyApplied, line 48: Since we know the size of the resulting map, consider using `make` here.  I mean, who introduced that empty map literal initialization anyw-- oh...

PurelyApplied (approved):

Issue #211, opened 2020-12-03
Do not propagate taint through a channel that is tainted in one select branch and read from in another branch


mlevesquedion:
Consider the following test case:

```
func TestTaintedAndSinkedInDifferentBranches(objects chan interface{}) {
	select {
	case objects <- core.Source{}:
	case s := <-objects:
		// TODO(211) want no report here, because objects is only tainted if the
		// other branch is taken, and only one branch can be taken
		core.Sink(s) // want "a source has reached a sink"
	}
}

Given the semantics of `select`, one of two things can happen:
1. We put a source on the `objects` channel
2. We get an `interface{}` value from the `objects` channel and we sink it

As indicated by the `TODO`, it is therefore not possible that the channel be tainted *and* that a tainted value is obtained from it in the same call to this function.

However, this code may still be unsafe, because the fact that one of the cases puts a source on the channel indicates that the channel may contain sources in general. Consider e.g. a case where this function is called concurrently in two different goroutines. Such a case would be unsafe code. Handling this behavior in full generality within the scope of a single function would require us to model every interaction with the channel. Alternatively, this could be handled via interprocedural analysis (#99).

This issue is somewhat related to #161 and #210, in the sense that it is related to a behavior where we may want to leverage

PR #235, opened 2020-12-17
Handle select instructions

mlevesquedion:
This PR is a part of #188 and constitutes a deep investigation into the `ssa.Select` instruction.

Our current approach to `Select` instructions is "visit every referrer and operand". This runs the risk of producing false positives:

```go
func TestRecvFromTaintedAndNonTaintedChans(sources <-chan *core.Source, innocs <-chan *core.Innocuous) {
	select {
	case s := <-sources:
	case i := <-innocs:
		core.Sink(innocs) // a false positive report is produced here
		core.Sink(i) // a false positive report is produced here
	}
	core.Sink(innocs) // a false positive report is produced here
}
```

In this case, both `sources` and `innocs` are `Operands` of the `select` instruction. Since `sources` is tainted, visiting the `Select` instruction leads to `innocs` being tainted, which is incorrect.

`Select` instructions are a bit special, in that each of the cases is wrapped in a `SelectState` struct that captures the direction of the operation (send or recv), the channel involved, and if applicable, the value being sent. This means that, in a sense, there are instructions "hidden" within the `Select` instruction. These "hidden" instructions do not appear among the other instructions in a block, so we need special handling for `Select` instructions.


internal/pkg/levee/propagation/propagation.go
```diff
@@ -52,8 +52,8 @@ func Dfs(n ssa.Node, conf *config.Config, taggedFields fieldtags.ResultType) Pro
 	}
 	maxInstrReached := map[*ssa.BasicBlock]int{}
 
-	record.visitReferrers(n, maxInstrReached, nil)
 	record.dfs(n, maxInstrReached, nil, false)
+	record.visitReferrers(n, maxInstrReached, nil)
```
mlevesquedion, line 56: This was added in #214 to replace a check in `visit`, which was: 
```go
if n == prop.root {
    prop.visitReferrers(n, ...)
}
```
The intent of that check was just to ensure that the referrers of a `Source` node are visited.

A side effect of putting `visitReferrers` first is that the traversal order, as recorded by `preOrder` and `marked`, is no longer correct. Indeed, the root is marked *after* its referrers have been visited.

Having the correct order is necessary for the changes in this PR.

internal/pkg/levee/propagation/propagation.go
```diff
@@ -243,6 +246,45 @@ func (prop *Propagation) visitOperands(n ssa.Node, maxInstrReached map[*ssa.Basi
 	}
 }
 
+func (prop *Propagation) visitSelect(sel *ssa.Select, maxInstrReached map[*ssa.BasicBlock]int, lastBlockVisited *ssa.BasicBlock) {
+	// The Send operations have to be processed before the Recv operations in case a channel
+	// is written to in one branch and read from in a different branch. This matters if the
+	// branches are reachable from each other, e.g. when the select is in a loop.
```
mlevesquedion, line 252: This behavior is covered by existing tests, see `TestTaintedAndSinkedInDifferentBranches(InLoop)`.

internal/pkg/levee/propagation/propagation.go
```diff
@@ -243,6 +246,45 @@ func (prop *Propagation) visitOperands(n ssa.Node, maxInstrReached map[*ssa.Basi
...
+	}
+
+	if sel.Referrers() == nil {
+		return
+	}
+
+	recvStateCount := 0
+	for _, s := range sel.States {
+		switch s.Dir {
+		case types.RecvOnly:
```
PurelyApplied, line 269: This could also use a top-level clarifying comment.  `// Propagate taint when receiving a from a tainted channel."

-----

The separation of \<examine all send\> and \<examine all recv\> makes me uncomfortable.  This feels more like a BFS spread out than two disjoint considerations.

-----

I was toying with daisy-chain `select` blocks to explore that separation of Send and Recv logic.  I note that...

```golang
func TestFoo(srcs chan core.Source, i1, i2, i3, i4 chan interface{}) {
	select {
	case x := <-i3:
		i4 <- x
	case y := <-i2:
		i3 <- y
	case z := <-i1:
		i2 <- z
	case s := <-srcs:
		i1 <- s
	}
	core.Sink(srcs) // want "a source has reached a sink"
	core.Sink(i1)   // want "a source has reached a sink"
	core.Sink(i2)   // TODO want "a source has reached a sink"
	core.Sink(i3)   // TODO want "a source has reached a sink"
	core.Sink(i4)   // want "a source has reached a sink"
}

func TestBar(srcs chan core.Source, i1, i2, i3, i4 chan interface{}) {
	select {
	case s := <-srcs:
		i1 <- s
	case z := <-i1:
		i2 <- z
	case y := <-i2:
		i3 <- y
	case x := <-i3:
		i4 <- x
	}
	core.Sink(srcs) // want "a source has reached a sink"
	core.Sink(i1)   // want "a source has reached a sink"
	core.Sink(i2)   // TODO want "a source has reached a sink"
	core.Sink(i3)   // TODO want "a source has reached a sink"
	core.Sink(i4)   // TODO want "a source has reached a sink"
}
```

The order-dependent tainting of `i4` is concerning.

internal/pkg/levee/propagation/propagation.go
```diff
@@ -243,6 +246,45 @@ func (prop *Propagation) visitOperands(n ssa.Node, maxInstrReached map[*ssa.Basi
...
+		case types.SendOnly:
+			if prop.marked[s.Send.(ssa.Node)] {
+				prop.dfs(s.Chan.(ssa.Node), maxInstrReached, lastBlockVisited, false)
+			}
+		}
+	}
+
+	if sel.Referrers() == nil {
+		return
+	}
```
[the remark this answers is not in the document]
mlevesquedion: Ack. Upon closer inspection, it doesn't look like it's possible to generate a `Select` instruction without `Referrers`, e.g. a minimal `select{}` with no cases still has a referring `Extract`.

I think I'd rather keep the `if` for safety, though. The alternative option would be to write a comment explaining why a `Select` with no `Referrers` is currently impossible. WDYT?

internal/pkg/levee/propagation/propagation.go
```diff
@@ -243,6 +246,45 @@ func (prop *Propagation) visitOperands(n ssa.Node, maxInstrReached map[*ssa.Basi
 	}
 }
 
+func (prop *Propagation) visitSelect(sel *ssa.Select, maxInstrReached map[*ssa.BasicBlock]int, lastBlockVisited *ssa.BasicBlock) {
+	// The Send operations have to be processed before the Recv operations in case a channel
+	// is written to in one branch and read from in a different branch. This matters if the
+	// branches are reachable from each other, e.g. when the select is in a loop.
+	for _, s := range sel.States {
+		switch s.Dir {
+		case types.SendOnly:
```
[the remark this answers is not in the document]
mlevesquedion: Good point. This started out as a `switch`, which I failed to convert to an `if` when it was no longer appropriate.

Your interpretation is correct. Comment added.

internal/pkg/levee/propagation/propagation.go
mlevesquedion: Woops. Good old "increment at the end of loop body isn't reached because of earlier continue so when the channel is actually tainted the wrong extract is visited".

Comments + suggested tests added. I've also added a relevant issue number (#211) to the `TODO`s.

internal/pkg/levee/propagation/propagation.go
```diff
@@ -248,6 +251,46 @@ func (prop *Propagation) visitOperands(n ssa.Node, maxInstrReached map[*ssa.Basi
...
+			recvStateCount++
+			if !prop.marked[s.Chan.(ssa.Node)] {
+				continue
+			}
+			// If the channel is tainted, propagate taint to the appropriate Extract
+			for _, r := range *sel.Referrers() {
+				e, ok := r.(*ssa.Extract)
+				// Select returns a tuple whose first 2 elements are irrelevant for our
+				// analysis. The other elements map 1:1 with each of the recv states.
+				// See the docs for ssa.Extract for more details.
```
PurelyApplied, line 284: Something felt a little off-kilter about this docstring explaining the `recvStateCount` that is used in the wrapping blocks.  And as mentioned elsewhere, the separation of send and recv felt a bit strange to me...

I was poking at it as I was trying to understand this PR, and I wound up with this implementation of `visitSelect`.  What do you think of the readability of extracting this counter as a bit of preprocessing, and combing the actual assessment of each `case` within a single loop?

```golang
func (prop *Propagation) visitSelect(sel *ssa.Select, maxInstrReached map[*ssa.BasicBlock]int, lastBlockVisited *ssa.BasicBlock) {
	// Comment about extracted tuple and why count starts at 2.  Reference docs.
	// This could also be extracted as a separate method if that would further improve readability
	var recvInds = map[int]int{}
	count := 2
	for i, s := range sel.States {
		if s.Dir == types.RecvOnly {
			recvInds[i] = count
			count++
		}
	}

	for i, s := range sel.States {
		switch {
		// High-level comment about tainting chans
		case s.Dir == types.SendOnly && prop.marked[s.Send.(ssa.Node)]:
			prop.dfs(s.Chan.(ssa.Node), maxInstrReached, lastBlockVisited, false)

		// High-level comment about recieving tainted value via chans
		case s.Dir == types.RecvOnly && prop.marked[s.Chan.(ssa.Node)]:
			if sel.Referrers() == nil {
				continue
			}
			// Implementation comment about extraction referrer
			for _, r := range *sel.Referrers() {
				e, ok := r.(*ssa.Extract)
				if !ok || e.Index != recvInds[i] {
					continue
				}
				prop.dfs(e, maxInstrReached, lastBlockVisited, false)
			}
		}
	}
}
```

It puts the nil-safety check nearer its usage, too, so even if it ends up being unreachable in practice, your argument for keeping it there as safety is more apparent in the implementation.

Thoughts?

internal/pkg/levee/testdata/src/example.com/tests/select/tests.go
```diff
@@ -122,3 +122,63 @@ func TestTaintedInForkedCall(objects chan interface{}) {
...
+	case y := <-i2:
+		i3 <- y
+	case x := <-i3:
+		i4 <- x
+	}
+	core.Sink(srcs) // want "a source has reached a sink"
+	core.Sink(i1)   // want "a source has reached a sink"
+	core.Sink(i2)   // TODO(#211) want "a source has reached a sink"
+	core.Sink(i3)   // TODO(#211) want "a source has reached a sink"
+	core.Sink(i4)   // TODO(#211) want "a source has reached a sink"
```
PurelyApplied, line 165: I maybe misunderstood the intent, since `i4` was previously emitting diagnostic.  Reading #211, the objective would be this current behavior where `i2, i3, i4` are _not_ considered tainted, since that would be outside the scope of this once select.  So these shouldn't be `TODO`s, since they're currently behaving as desired.

Am I misunderstanding #211 here?

If the `select` were within a `for` loop, then they should emit.  (Testing locally, indeed they do.)  That test-case could be added here as well.

internal/pkg/levee/propagation/propagation.go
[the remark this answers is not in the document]
PurelyApplied: If we're keeping it for the nil-safety, I think it should either self-document by being before usage, or having a comment.  Although this thread is a bit elided by my below comment...

internal/pkg/levee/testdata/src/example.com/tests/select/tests.go
mlevesquedion: You're right, this is different from #211.

I've added the `for` loop test case.

internal/pkg/levee/propagation/propagation.go
mlevesquedion: I do like having the two cases in the same loop + switch. My perception that the sends have to be visited before the recvs may have been caused by the earlier `count++` bug that is now fixed.

I also like the idea of computing the indices in advance. I took a slightly different approach w.r.t. to the implementation details. Starting the count at 2 doesn't make sense to me.

internal/pkg/levee/propagation/propagation.go
```diff
@@ -248,6 +251,45 @@ func (prop *Propagation) visitOperands(n ssa.Node, maxInstrReached map[*ssa.Basi
...
+	// 0-based, the Extract corresponding to the current Recv state
+	// therefore has an Index that is 1 greater than the current recvCount
+	// when that state is reached.
+	// See the ssa package code for more details.
+	recvCount := 0
+	extractIndex := map[*ssa.SelectState]int{}
+	for _, ss := range sel.States {
+		if ss.Dir == types.RecvOnly {
+			recvCount++
+			extractIndex[ss] = recvCount + 1
```
PurelyApplied, line 267: The comment above feels like it's working to coerce 0-indexing into 1-indexing.  Consider flipping these instructions, i.e.

```
extractIndex[ss] = recvCount + 2
recvCount++
```
and using 0-indexing throughout.

The comment could be massaged at that point to something more like what you had prior.  "Select returns a tuple whose first two elements are unused for our analysis.  The remainder of the tuple corresponds one-to-one to channels in the Recv state."

PurelyApplied (approved):
LGTM.

internal/pkg/levee/propagation/propagation.go
mlevesquedion: I considered doing that, but I'm not sure it's more readable. On the one hand, not having the "2" anywhere in the code feels confusing. On the other hand, writing `... = recvCount + 2` feels like it's incorrect because it's using an outdated value of `recvCount`.

That being said, I tried changing `recvCount` to `recvIndex` and I think that this naming change, combined with your proposal, is what makes the most sense.

Thank you for your valuable input!

Issue #191, opened 2020-11-13
Determine which Basic types can't meaningfully be tainted


mlevesquedion:
Currently, we are avoiding `Booleans`, because it is quite clear that those can't meaningfully be tainted. The are other `Basic` types we may want to avoid though, among the list of `types.BasicKinds`:

```go
type BasicKind int

const (
	Invalid BasicKind = iota // type is invalid

	// predeclared types
	Bool
	Int
	Int8
	Int16
	Int32
	Int64
	Uint
	Uint8
	Uint16
	Uint32
	Uint64
	Uintptr
	Float32
	Float64
	Complex64
	Complex128
	String
	UnsafePointer

	// types for untyped values
	UntypedBool
	UntypedInt
	UntypedRune
	UntypedFloat
	UntypedComplex
	UntypedString
	UntypedNil

	// aliases
	Byte = Uint8
	Rune = Int32
)
```

(this is from `go/types/type.go`)

For additional context, see this discussion: https://github.com/google/go-flow-levee/pull/178#discussion_r522506449

mlevesquedion:
While not a `Basic` type, we want to avoid `Signatures`: #194.

PR #236, opened 2020-12-18
Handle slice instructions

mlevesquedion:
This PR is a part of #188 and constitutes a deep investigation into the `ssa.Slice` instruction.

Currently, the way we handle `Slice` instructions is that we visit every referrer and operand.

This can create false positives:

```go
func TestSliceBoundariesAreNotTainted(lo, hi, max int) {
	sources := [1]core.Source{{Data: "secret"}}
	slice := sources[lo:hi:max]
	core.Sink(lo)      // a false positive report is produced here
	core.Sink(hi)      // a false positive report is produced here
	core.Sink(max)  // a false positive report is produced here
	_ = slice
}
```

One could argue that in the above case, the incorrect behavior is actually that we are tainting integers. I encourage you to discuss this issue here: #191. In any case, within the context of a `Slice` instruction, it is incorrect to visit these `Operands`.

I originally thought that the only thing that should legitimately be visited is the `Referrers`. However, consider this case:

```go
func TestSlicedArrayIsTainted() {
	innocs := [1]interface{}{nil}
	slice := innocs[:]
	slice[0] = core.Source{Data: "secret"}
	core.Sink(innocs) // want "a source has reached a sink"
	_ = slice
}
```

In this case, the backing array of `slice` is written to when doing `slice[0] = core.Source{Data: "secret"}`, so to properly model the taint propagation, we need to visit the `Slice`'s `X` field (one of its `Operands`).


internal/pkg/levee/propagation/propagation.go
```diff
@@ -199,6 +199,12 @@ func (prop *Propagation) visit(n ssa.Node, maxInstrReached map[*ssa.BasicBlock]i
 	case *ssa.Send:
 		prop.dfs(t.Chan.(ssa.Node), maxInstrReached, lastBlockVisited, false)
 
+	case *ssa.Slice:
+		prop.visitReferrers(n, maxInstrReached, lastBlockVisited)
+		// This allows taint to propagate backwards into the sliced value
+		// when the resulting slice is tainted
```
PurelyApplied, line 205: Nit: the underlying collection is not necessarily a `Slice`.  Consider a more generic phrasing here.

PurelyApplied (approved):

PR #238, opened 2020-12-21
Loop control refactor

PurelyApplied:
This refactor spun out from discussion on #223.

The loop control in the switch during source identification can be hard to follow.  This PR refactors the switch to identify values entirely, rather than to defer to a snippet after the switch.  With this condensation, the switch itself can be extracted into a separate method, improving readability in the loop over instructions.


vinayakankugoyal (approved):

internal/pkg/source/source.go
```diff
@@ -119,73 +119,59 @@ func sourcesFromBlocks(fn *ssa.Function, conf *config.Config, taggedFields field
...
+			if n := instr.(ssa.Node); isSourceNode(n, conf, propagators, taggedFields) {
+				sources = append(sources, New(n))
 			}
 		}
 	}
 	return sources
 }
 
+func isSourceNode(n ssa.Node, conf *config.Config, propagators fieldpropagator.ResultType, taggedFields fieldtags.ResultType) bool {
+	switch v := n.(type) {
```
mlevesquedion, line 131: Save for `Extract`, all of these cases have an identical call: `IsSourceType(conf, taggedFields, n.(ssa.Value).Type())`. 

Consider extracting a variable before the switch, e.g. `nodeHasSourceType = IsSourceType(...)` to help readability.

PurelyApplied: > Save for Extract and TypeAssert, all of these cases have an identical call: IsSourceType(conf, taggedFields, n.(ssa.Value).Type()).

Crucially, the `default` case is doing some lifting here.  Non-`Value` instructions do not have `Type()`.  That's an easy enough check to add... But even so restricted, `Range` instructions cause panics.  From [their docs](https://pkg.go.dev/golang.org/x/tools/go/ssa#Range).

> Type() returns an opaque and degenerate "rangeIter" type.

So to extract that call would be to add the preamble:

```golang
func isSourceNode(n ssa.Node, conf *config.Config, propagators fieldpropagator.ResultType, taggedFields fieldtags.ResultType) bool {
	_, isVal := n.(ssa.Value)
	_, isRange := n.(*ssa.Range)
	if !isVal || isRange {
		return false
	}

	isSrcTyp := IsSourceType(conf, taggedFields, n.(ssa.Value).Type())
	switch v := n.(type) {
		// ...
	}
}
```

At that point, I'm not sure which is more readability.  Thoughts?

internal/pkg/source/source.go
```diff
@@ -119,73 +119,59 @@ func sourcesFromBlocks(fn *ssa.Function, conf *config.Config, taggedFields field
...
+
+	// Values produced by sanitizers are not sources.
+	// Values produced by field propagators are.
+	case *ssa.Call:
+		return !isProducedBySanitizer(v, conf) &&
+			(propagators.IsFieldPropagator(v) || IsSourceType(conf, taggedFields, n.(ssa.Value).Type()))
+
+	// A type assertion can assert that an interface is of a source type.
+	// Only panicky type asserts will refer to the source Value.
+	// _, ok type assertions are checked as a source type in the ssa.Extract instruction.
```
mlevesquedion, line 148: Nit: ... in the *case for the* ssa.Extract ...

internal/pkg/source/source.go
mlevesquedion: Ah, good point. On closer look, I think the repetition is fine. I think I got distracted by the call itself, which looks quite onerous.

mlevesquedion (approved):

internal/pkg/source/source.go
PurelyApplied: `case` specified.  Also removed how that case is doing its check - that belongs to that case's comment.

PR #240, opened 2020-12-21
Discussion: Use ResultType as the de facto interface.

PurelyApplied:
This PR intended as discussion for neighboring PR.

While plumbing arguments down in #239, I was blocked at the exported `IsSourceType`.  It made me wonder what the meaningful distinction between a top-level function and a result method is.

This PR shifts `source.IsSourceType` to `source.ResultType.IsSourceType`.  Doing so makes the `source.ResultType` the summary interface for the `source` package and analyzer.  This would cleanly separate our internal package usage from external package consumption.  That separation gives much more modular flexibility in implementation changes.

On the other hand, it is essentially a reproduction of the Exported / Unexported model.  The real solution may really be to be more careful about consumption of exported functions by unexported methods.

### DIFF
--- a/internal/pkg/levee/propagation/propagation.go
+++ b/internal/pkg/levee/propagation/propagation.go
@@ -43,7 +43,7 @@ type Propagation struct {
 
 // Dfs performs a depth-first search of the graph formed by SSA Referrers and
 // Operands relationships, beginning at the given root node.
-func Dfs(n ssa.Node, conf *config.Config, taggedFields fieldtags.ResultType) Propagation {
+func Dfs(n ssa.Node, conf *config.Config, taggedFields fieldtags.ResultType, sourceResult source.ResultType) Propagation {
 	record := Propagation{
 		root:         n,
 		marked:       make(map[ssa.Node]bool),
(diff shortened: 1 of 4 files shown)


internal/pkg/levee/propagation/propagation.go
```diff
@@ -151,14 +151,14 @@ func (prop *Propagation) visit(n ssa.Node, maxInstrReached map[*ssa.BasicBlock]i
 
 		// This is to avoid attaching calls where the source is the receiver, ex:
 		// core.Sinkf("Source id: %v", wrapper.Source.GetID())
-		if recv := t.Call.Signature().Recv(); recv != nil && source.IsSourceType(prop.config, prop.taggedFields, recv.Type()) {
+		if recv := t.Call.Signature().Recv(); recv != nil && sourceResult.IsSourceType(prop.config, prop.taggedFields, recv.Type()) {
```
mlevesquedion, line 154: This seems to be the only place where this function is used. I don't think this function should be exposed by the `source` analyzer. I think that function should be owned by a different package, which the `source` analyzer and other analyzers would use. (e.g. see my comment on #239).

As it is, having this function in the source package, and especially on its `ResultType`, feels confusing to me. Why would a `map` from functions to source nodes know how to identify source types? Why does it need to be fed the config and tagged fields? I think this should be an `IsSourceType` function on a `Classifier` that the `Propagation` would hold a reference to. This would simplify this and other calls.

PurelyApplied:
Summarizing relevant threads here and offline: while more attention could be pair to the interface exposed by the `ResultType` of various analyzers, that abstraction doesn't need to be formalized in an interface, nor act as the sole point of entry.

PR #241, opened 2020-12-21
Delete map key does not propagate source.

### LINKED ISSUE #186: Deleting a key from a tainted map should not taint the key
## Context

```go
func TestDeletingFromTaintedMapDoesNotTaintTheKey(key string, sources map[string]core.Source) {
	delete(sources, key)
	// TODO: no report should be produced here
	core.Sink(key) // want "a source has reached a sink"
}
```

## Discussion

The simplest way to handle this would be to check if the `delete`'s associated `Call`'s `StaticCallee` is the `Builtin` function `delete`, and avoid propagating if that is the case.

benhxy:
Fixes #186

- [x] Ran `go test`


go.sum
```diff
@@ -2,18 +2,24 @@ github.com/davecgh/go-spew v1.1.1 h1:vj9j/u1bqnvCEfJOwUhtlOARqs3+rkHYY13jYWTU97c
 github.com/davecgh/go-spew v1.1.1/go.mod h1:J7Y8YcW2NihsgmVo/mv3lAwl/skON4iLHjSsI+c5H38=
 github.com/google/go-cmp v0.5.2 h1:X2ev0eStA3AbceY54o37/0PQ/UWqKEiiO2dKL5OPaFM=
 github.com/google/go-cmp v0.5.2/go.mod h1:v8dTdLbMG2kIc/vJvl+f65V22dbkXbowE6jgT/gNBxE=
+github.com/yuin/goldmark v1.1.27 h1:nqDD4MMMQA0lmWq03Z2/myGPYLQoXtmi0rGVs95ntbo=
```
vinayakankugoyal, line 5: Why are these changes required?

benhxy: Hmm I have no idea why this is changed every time I build or run. I suspect it has something to do with our internal $GOPATH but I'm not sure.

benhxy: Reverted

internal/pkg/levee/propagation/propagation.go
```diff
@@ -145,6 +145,11 @@ func (prop *Propagation) visit(n ssa.Node, maxInstrReached map[*ssa.BasicBlock]i
 		}
 
 	case *ssa.Call:
+		// The builtin delete(m map[Type]Type1, key Type) func does not propagate source.
```
mlevesquedion, line 148: Nit: ... does not propagate *taint*. (If this is just a typo, ignore the following: a `source` is a value that is itself sensitive, it is the starting point of a `taint` propagation analysis; `taint` propagates from value to value, starting from a `source`)

mlevesquedion (approved):
LGTM, just a nit.

internal/pkg/levee/propagation/propagation.go
benhxy: Updated the doc.

PR #242, opened 2020-12-22
Refactor calls to determine if an instruction is tainted

mlevesquedion:
As a top-level, end-of-the-chain analyzer, `levee` should be able to just ask the `Propagation` if an instruction is tainted. `levee` shouldn't know about `sanitizer`s if it doesn't have to.

### DIFF
--- a/internal/pkg/levee/levee.go
+++ b/internal/pkg/levee/levee.go
@@ -78,7 +78,7 @@ func run(pass *analysis.Pass) (interface{}, error) {
 func reportSourcesReachingSink(pass *analysis.Pass, sources []*source.Source, instr ssa.Instruction, propagations map[ssa.Node]propagation.Propagation) {
 	for _, s := range sources {
 		prop := propagations[s.Node]
-		if prop.HasPathTo(instr.(ssa.Node)) && !prop.IsSanitizedAt(instr) {
+		if prop.IsTainted(instr) {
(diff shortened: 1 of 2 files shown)


vinayakankugoyal (approved):

PurelyApplied (approved):
* I agree `HasPathTo` is no longer a fitting name.
* For `report*` and the two touched methods, consider changing `instr` to `sink`.  These are specifically methods that are examining if a tainted value has reached a sink, so the generic `instr` could be improved.
* Given that, I'm not sure `IsTainted(sink)` really fits.  Perhaps `TaintCanReach(sink)`, although that's practically back to `HasPathTo`.  \shrug.
* Inasmuch as someone can have a _forte_, naming is very much my _piano_, so double-\shrug.

mlevesquedion:
> For report* and the two touched methods, consider changing instr to sink. These are specifically methods that are examining if a tainted value has reached a sink, so the generic instr could be improved.

I think in `levee`, `sink` is indeed the right name. This change was actually sneaked into #234. In the `propagation` package though, I think we should stick with `instr`, since this functionality could be used for more than just source to sink tracing. Indeed, if we wanted to see which `Return` instructions in a function are reached by which parameters (e.g. for IPA), this `IsTainted` instruction would again come in handy.

PR #243, opened 2020-12-23
Handle append and copy builtins

mlevesquedion:
The idea for this PR came from recent PR's implementing special handling for certain builtin instructions (specifically, `panic` #226 and `delete` #241). Currently, we handle other builtins like any other call, but since the builtins' behavior is fully known, we can do a bit better.

Consider this example for the `copy` builtin:
```go
func TestFalsePositive(s core.Source) {
	data := []byte(s.Data)
	redacted := []byte("<redacted>")
	copy(data, redacted)
	core.Sink(redacted) // a false positive report is produced here
}
```

Here, `redacted` is tainted via a backwards flow from the destination of the copy to the source of the copy. For slices, such an inference doesn't make sense: the taint can only propagate from the source to the destination.

I couldn't construct false positive cases for the `append` builtin, thanks to other mechanisms we have for guarding against incorrect propagations. Still, I think it's good to have better test coverage.

Because the `Call` case in the switch was getting unwieldy, I extracted it to its own function. I also extracted a function that handles visiting call arguments, since this requires a special check (the args have to be reference-like) and this check was being duplicated.

## Future work

Go has 15 builtins:
```go
func append(slice []Type, elems ...Type) []Type
func cap(v Type) int
func close(c chan<- Type)
func complex(r, i FloatType) ComplexType
func copy(dst, src []Type) int
func delete(m map[Type]Type1, key Type)
func imag(c ComplexType) FloatType
func len(v Type) int
func make(t Type, size ...IntegerType) Type
func new(Type) *Type
func panic(v interface{})
func print(args ...Type)
func println(args ...Type)
func real(c ComplexType) FloatType
func recover() interface{}
```

Of those, the following could still require some special handling:
* `len` and `cap` will return a "tainted" integer if the given collection is tainted. This ties into #191.
* `print` and `println` may be considered sinks.


PurelyApplied (approved):
I agree with your assessment of the other builtins.  Personally, I'd consider `len` and `cap` as metadata rather than source, but I'll comment in the Issue about it.

internal/pkg/levee/propagation/propagation.go
```diff
@@ -261,6 +241,50 @@ func (prop *Propagation) visitOperands(n ssa.Node, maxInstrReached map[*ssa.Basi
...
+func (prop *Propagation) visitCall(call *ssa.Call, maxInstrReached map[*ssa.BasicBlock]int, lastBlockVisited *ssa.BasicBlock) {
+	// Some builtins require special handling
+	if builtin, ok := call.Call.Value.(*ssa.Builtin); ok {
+		switch builtin.Name() {
+		// Only the slice being appended to and the referrers can be tainted.
+		// (The elements being appended cannot be tainted.)
+		case "append":
+			prop.visitCallArg(call.Call.Args[0], maxInstrReached, lastBlockVisited)
+			prop.visitReferrers(call, maxInstrReached, lastBlockVisited)
+			return
```
PurelyApplied, line 253: `append` does a lot...

* If the capacity of the slice - that is, the size of the underlying array - passed at argument 0 can fit the item(s) to be appended, they are written to the underlying array.
* If the capacity of the slice cannot accommodate the new items, a new array is allocated and elements are copied.
* In either case, the resized (with possibly new-underlying-array) slice is returned.

So strictly speaking, I think `append` should taint only its return argument and the underlying array, not the slice itself.  The arg0 slice only has access to the tainted value if it had capacity and is later re-sliced to expand the capacity.  See this [GoPlayground example](https://play.golang.org/p/SDuDtdz_Te1).

I don't know that we should try to do all that, but it would warrant a comment.

PurelyApplied (commented):
[I had thought my previous comment would update PR state `Approval -> Comment`, since I wanted to talk about the `append`s.  Double-commenting to get state change.

internal/pkg/levee/propagation/propagation.go
mlevesquedion: The "if there's capacity, the underlying array will be written to" is why I wanted to traverse to the slice that is being appended to (the `arg0` slice).

> I think append should taint only its return argument and the underlying array, not the slice itself.

I agree, but I don't think we can do that. As you mention, we don't want to try modelling the underlying array's capacity. In general, we also don't have a way to get at the underlying array (consider a slice received as a parameter), so I think tainting the slice is a reasonable approximation.

I've added a comment to explain why we want to traverse to the `arg0` slice.

internal/pkg/levee/propagation/propagation.go
```diff
@@ -261,6 +241,53 @@ func (prop *Propagation) visitOperands(n ssa.Node, maxInstrReached map[*ssa.Basi
...
+		// Only the first argument (dst) can be tainted. (The src cannot be tainted.)
+		case "copy":
+			prop.visitCallArg(call.Call.Args[0], maxInstrReached, lastBlockVisited)
+			return
+
+		// The builtin delete(m map[Type]Type1, key Type) func does not propagate a taint.
+		case "delete":
+			return
+		}
+	}
```
vinayakankugoyal, line 267: I think we should break this out into its own function like visitCallToBuiltin or visitBuiltinCall or visitBuiltin.

mlevesquedion: Agreed. Done.

vinayakankugoyal (approved):

PurelyApplied (approved):

PR #246, opened 2020-12-30
Handle pointer methodsets in fieldpropagator

mlevesquedion:
Consider this snippet from the [Go types tutorial](https://github.com/golang/example/tree/master/gotypes):

> A variable of type T has access to all the methods of type *T as well, due to the implicit address-of operation during method calls, but those extra methods are not part of the method set of T.

Currently, the `fieldpropagators` analyzer does not take this into account: it only analyzes the method set of `T`. However, if a method has a pointer receiver, it will be in `*T`'s method set, which means the analyzer will not even consider it.

This PR rectifies this situation by analyzing the method set of `*T` as well.


internal/pkg/fieldpropagator/analyzer.go
```diff
@@ -87,6 +84,26 @@ func run(pass *analysis.Pass) (interface{}, error) {
...
+	// The method sets of T and *T are disjoint.
+	// In Go code, a variable of type T has access to all
+	// the methods of type *T due to an implicit address operation.
+	methods = append(methods, methodValues(ssaProg, t)...)
+	return append(methods, methodValues(ssaProg, types.NewPointer(t))...)
+}
+
+func methodValues(ssaProg *ssa.Program, t types.Type) []*ssa.Function {
+	var methodValues []*ssa.Function
+	mset := ssaProg.MethodSets.MethodSet(t)
```
vinayakankugoyal, line 98: Dos this no support a ranged for loop?

vinayakankugoyal (approved):
LGTM

internal/pkg/fieldpropagator/analyzer.go
mlevesquedion: Sadly, no. A `MethodSet` is a wrapper over a slice, but the slice isn't exposed so we have to access it through these methods.

PR #252, opened 2021-01-05
Allow traversal through source methods

### LINKED ISSUE #249: Handle propagation via source methods
Consider the following (slightly contrived) test case:

```go
func TestPropagationViaSourceMethod(s core.Source) {
	str := s.Propagate(s.Data)
	core.Sink(str) // TODO want "a source has reached a sink"
}
```

Currently, we do *not* produce a report in the above case. This is because of the following snippet in `propagation.go`:

```go
if recv := t.Call.Signature().Recv(); recv != nil && sourcetype.IsSourceType(prop.config, prop.taggedFields, recv.Type()) {
	return
}
```

As far as I know, this check is to avoid incorrectly propagating through safe getters such as `Source.GetID()`. If a getter is not safe, it should be identified by the `fieldpropagators` analyzer, which will cause the `Call` to be identified as a source in `source.go`.

However, as seen in the test case at the beginning of this issue, this does not take into accounts methods on source types that can propagate taint received via one of their arguments. This could cause false negatives, and is inconsistent with how we handle calls in general : if a tainted argument goes in, we assume that whatever comes out is tainted.

mlevesquedion:
Fixes #249.

Allowing traversal through source methods is trickier than I expected. Consider this example:

```go
func TestPropagationViaSourceMethod(s core.Source) {
	tainted := s.Propagate(s.Data)
	core.Sink(tainted) // want "a source has reached a sink"
}
```

image

Under the current propagation logic, the call to `Propagate` will be reached first via the `Alloc --> UnOp --> Call` chain. Thus when the call is first visited, the propagation does not yet know that taint can reach the call via an argument.

The solution proposed by the current PR is to unmark the node if it is visited from the Source receiver. This allows a future traversal to taint the call via an argument.


internal/pkg/propagation/propagation.go
```diff
@@ -253,10 +253,21 @@ func (prop *Propagation) visitCall(call *ssa.Call, maxInstrReached map[*ssa.Basi
...
-	// This is to avoid attaching calls where the source is the receiver, ex:
-	// core.Sinkf("Source id: %v", wrapper.Source.GetID())
+	// Do not traverse through a method call if it is being reached via a Source receiver.
+	// Do traverse if the call is being reached through a tainted argument.
+	// Source methods that return tainted values regardless of their arguments should be identified by the fieldpropagator analyzer.
 	if recv := call.Call.Signature().Recv(); recv != nil && sourcetype.IsSourceType(prop.config, prop.taggedFields, recv.Type()) {
-		return
+		visitingFromArg := false
+		for _, a := range call.Call.Args[1:] {
+			if prop.marked[a.(ssa.Node)] {
```
vinayakankugoyal, line 262: if prop.tainted... here and other places.

vinayakankugoyal (commented):
LGTM, but you might need to make changes inline with #225

internal/pkg/propagation/propagation.go
mlevesquedion: Weird. I didn't get a merge conflict so I assumed everything was fine. Sorry about that.

vinayakankugoyal (approved):

PurelyApplied:
> Consider this example

The code for this example would be helpful.

mlevesquedion:
Sorry about that. I've added it to the description.

PurelyApplied (approved):
I had some longer rambling comments about how this didn't sit well with me, but it's probably fine.  The thrust was that we had talked about updating the fieldpropagator analyzer to identify "blessed" methods rather than "known bad" methods.  I might push on prioritizing that, since I think this method would read more cleanly in that context.

mlevesquedion:
I've thought some more about the "known safe" vs "known bad" approaches. I'm not convinced that switching to "known safe" would give us much, because it boils down to the same problem, i.e. due to analysis limitations we may incorrectly claim that a function is safe, just as currently we may miss an unsafe function. Now that the `fieldpropagators` analyzer uses the same logic as the main analysis, I feel more confident about the quality of its results.

I'll do some experimenting to see how this code would read if we took the "known safe" approach and I'll let you know how it goes.

PR #255, opened 2021-01-07
Avoid collision of cache config.

PurelyApplied:
* A collision can occur if tests are run using multiple configurations
within a single package.  This commit attaches source file to config
caching to avoid this issue.

-----

While iterating on #103 and adding tests for format of reporting with or without `ReportMessage` configured, I ran into an issue involving the config cache persisting across multiple tests within a single package.  This might not be the case in all environments, but I hit it on the command-line and in IntelliJ.

This PR separates cached config by the filename of the config's source to avoid this collision.

-----


internal/pkg/config/config.go
```diff
@@ -284,27 +284,59 @@ func (fm funcMatcher) MatchFunction(path, receiver, name string) bool {
...
+}
 
-// ReadConfig reads, parses, and validates config file.
-func ReadConfig() (*Config, error) {
-	readFileOnce.Do(func() {
+// configCache safely stores readConfigResults for concurrent access.
+type configCache struct {
+	mu    sync.Mutex
+	cache map[string]*readConfigResults
+}
```
PurelyApplied, line 298: This could probably be replaced by `sync.Map` if anyone has a strong preference.  However, from the `sync` documentation...

> The Map type is specialized. Most code should use a plain Go map instead, with separate locking or coordination, for better type safety and to make it easier to maintain other invariants along with the map content.

> The Map type is optimized for two common use cases: (1) when the entry for a given key is only ever written once but read many times, as in caches that only grow, or (2) when multiple goroutines read, write, and overwrite entries for disjoint sets of keys. In these two cases, use of a Map may significantly reduce lock contention compared to a Go map paired with a separate Mutex or RWMutex.

While we satisfy the first point in the optimization, we expect multiple reads to be occurring on the same key throughout each package's analysis passes.  As such, I've deferred to the recommendation of a plain Go map with separate lock coordination.

vinayakankugoyal: No strong preference but use case (1) fits very well for us.

mlevesquedion:
Could you expand on how you discovered this, specifically what tests were failing and how? The issue seems to be that multiple tests run in the same process and therefore only the "first" config is loaded because of the cache. This is making me wonder why we haven't run into this issue before, since e.g. `levee_test.go` has two test functions that use different configs. What is different about your new tests that caused them to fail?

internal/pkg/config/config.go
```diff
@@ -284,27 +284,59 @@ func (fm funcMatcher) MatchFunction(path, receiver, name string) bool {
...
+	}
+}
+
+var cache = configCache{
+	mu:    sync.Mutex{},
+	cache: make(map[string]*readConfigResults),
+}
+
+// ReadConfig reads, parses, and validates config file.
+func ReadConfig() (*Config, error) {
```
mlevesquedion, line 338: Nit: consider moving this function above the others so that the code can be read top-to-bottom.

internal/pkg/config/config.go
```diff
@@ -284,27 +284,59 @@ func (fm funcMatcher) MatchFunction(path, receiver, name string) bool {
...
+	w.mu.Lock()
+	defer w.mu.Unlock()
+
+	if w.cache[file] == nil {
+		w.cache[file] = new(readConfigResults)
+	}
+}
+
+var cache = configCache{
+	mu:    sync.Mutex{},
```
mlevesquedion, line 333: Nit: this can be omitted.

internal/pkg/config/config.go
```diff
@@ -284,27 +284,59 @@ func (fm funcMatcher) MatchFunction(path, receiver, name string) bool {
...
-	readFileOnce.Do(func() {
+// configCache safely stores readConfigResults for concurrent access.
+type configCache struct {
+	mu    sync.Mutex
+	cache map[string]*readConfigResults
+}
+
+// Instantiates readConfigResults if file has not yet been loaded
+func (w *configCache) read(file string) (*Config, error) {
+	w.instantiateCacheForFile(file)
```
mlevesquedion, line 302: Consider returning the cached config from this method (the naming may need to be updated, e.g. `get...` instead of `instantiate...`).

internal/pkg/config/config.go
```diff
@@ -284,27 +284,59 @@ func (fm funcMatcher) MatchFunction(path, receiver, name string) bool {
...
+
+	cc := w.cache[file]
+	cc.once.Do(func() {
```
mlevesquedion, line 305: Nit: this feels like it should be a method on `readConfigResults`, but if you want `readConfigResults` to be a simple bag of fields, that's fine too.

internal/pkg/config/config.go
```diff
@@ -284,27 +284,59 @@ func (fm funcMatcher) MatchFunction(path, receiver, name string) bool {
 	return fm.Package.MatchString(path) && fm.Receiver.MatchString(receiver) && fm.Method.MatchString(name)
 }
 
-var readFileOnce sync.Once
-var readConfigCached *Config
-var readConfigCachedErr error
+// readConfigResults reduces disk access across multiple ReadConfig calls.
```
mlevesquedion, line 287: Nit: `readConfigResults` is fine, but I think `cachedConfig` may read slightly better.

internal/pkg/config/config.go
```diff
@@ -284,27 +284,59 @@ func (fm funcMatcher) MatchFunction(path, receiver, name string) bool {
...
+}
+
+// Instantiates readConfigResults if file has not yet been loaded
+func (w *configCache) read(file string) (*Config, error) {
+	w.instantiateCacheForFile(file)
+
+	cc := w.cache[file]
+	cc.once.Do(func() {
 		c := new(Config)
 		bytes, err := ioutil.ReadFile(configFile)
```
mlevesquedion, line 307: Seems like `configFile` should be `file`?

internal/pkg/config/config.go
```diff
@@ -284,27 +284,59 @@ func (fm funcMatcher) MatchFunction(path, receiver, name string) bool {
...
-	return readConfigCached, readConfigCachedErr
+
+	return cc.conf, cc.err
+}
+
+func (w *configCache) instantiateCacheForFile(file string) {
+	w.mu.Lock()
+	defer w.mu.Unlock()
+
+	if w.cache[file] == nil {
```
mlevesquedion, line 327: Nit: consider using `_, ok := w.cache[file]; ok`, as it is agnostic to the type of value in the map.

internal/pkg/config/config.go
PurelyApplied: I had thought the style was to specify all keys during initialization if you define any?  I can't seem to find where I read that, though, so I might be misremembering.

PurelyApplied:
> Could you expand on how you discovered this, specifically what tests were failing and how? 

Double whoops.  I meant to say working on PR #204, which would address issue #103.  I've since pushed the addition of the tests that uncovered this issue.


> This is making me wonder why we haven't run into this issue before, since e.g. `levee_test.go` has two test functions that use different configs. What is different about your new tests that caused them to fail?

I think we _have_ been hitting this, but quietly.  `levee_test.go` is the only place where we have the same package invoke `Flags.Set` twice on different files.  Test order seems to be sequential as written in file.  That means `TestLeveeDoesNotCreateReportsForPanicIfPanicingOnTaintedValuesIsAllowed` is using the wrong config.

The `nopanic` doesn't have any positive assertion of `// want` in its test file, so the fact that the wrong configuration is loaded allows a false pass.  This can be observed either by removing the 
```golang
	if err := Analyzer.Flags.Set("config", dataDir+"/allowpanicontaintedvalues-config.yaml"); err != nil {
		t.Error(err)
	}
```
entirely and observing that the test passes when run at the package level.  Alternatively, set appropriate breakpoints and observe that the `config` refers to the same address.

internal/pkg/config/config.go
mlevesquedion: I don't remember reading that in the style guides I've read. That feels like it could get annoying quickly if you have a struct with a lot of fields.

For context, there are places in the code where we don't respect this potential guideline, e.g. in `propagation.go:Taint`.

You may be thinking of the fact that if you key a field, the compiler won't let you have unkeyed fields in the same literal?

In any case. Explicitly setting defaults feels unnecessary to me, but this is not a big deal.

mlevesquedion:
Good investigation, thanks for explaining.

I've opened a fix for the broken tests here: #257.

internal/pkg/config/config.go
PurelyApplied: I actually had something similar first, but didn't like the near-collision with `type configCache`.  It's also not just the config anymore, but the config, error, and sync.  I could go with the more verbose `configCacheElement` I suppose?

internal/pkg/config/config.go
PurelyApplied: Done.

internal/pkg/config/config.go
PurelyApplied: Oof.  Bad overlook and good catch!

internal/pkg/config/config.go
PurelyApplied: Done.

internal/pkg/config/config.go
PurelyApplied: Moved, and added source file to the cache to avoid passing it twice to the `readOnce()` signature.

internal/pkg/config/config.go
PurelyApplied: Done.

internal/pkg/config/config.go
PurelyApplied: Dunno!  Either way, zero-value assignment pruned.

mlevesquedion (approved):
LGTM.

I think we could still get bit if tests run in parallel, e.g.
```go
A: sets flag to "a-config.yaml"
B: sets flag to "b-config.yaml"
A: calls ReadConfig(), gets B's config
```

But we can deal with that if/when it happens.

vinayakankugoyal:
Like @mlevesquedion mentioned this won't really work for parallel tests. If we will need to run the tests sequentially then won't it be easier to just convert `var readFileOnce sync.Once` to `var *readFileOnce sync.Once` and reset it in the tests where we need to update the config files?

PurelyApplied:
> Like @mlevesquedion mentioned this won't really work for parallel tests. If we will need to run the tests sequentially then won't it be easier to just convert `var readFileOnce sync.Once` to `var *readFileOnce sync.Once` and reset it in the tests where we need to update the config files?

I don't see how that avoids our race condition.  We'd still have

* A sets config file
* B resets config file
* A reads (reset) config file

I agree this isn't a perfect solution.  I'm really just trying to unblock other work for the moment.  If we need to start running tests in parallel, we can ~~burn~~cross that bridge when we get to it.

PR #257, opened 2021-01-07
Validate that nopanic test can catch leaks

mlevesquedion:
This PR builds on #255, which enables tests using different configs to run in the same test process.

This PR adds a leak to the `nopanic` tests to show that the test is actually running on the `core.go` file and using the proper configuration. Prior to #255, these tests were using the wrong config (`test-config.yaml` instead of `allowpanic*.yaml`).


internal/pkg/levee/levee_test.go
```diff
@@ -32,7 +32,7 @@ func TestLevee(t *testing.T) {
 	if *debugging {
 		Analyzer.Requires = append(Analyzer.Requires, debug.Analyzer)
 	}
-	analysistest.Run(t, dataDir, Analyzer, "./...")
+	analysistest.Run(t, dataDir, Analyzer, "./src/example.com/...")
```
mlevesquedion, line 35: This is needed because running on files under `nopanic.com` will produce a "no diagnostic was reported" error.

vinayakankugoyal (approved):

PurelyApplied (approved):

PR #259, opened 2021-01-13
Rework debug graph edges

mlevesquedion:
This PR makes a few changes to the way the debug SSA graphs are rendered:

1. Remove referrers edges
2. Flip operands edges

The reason referrers edges can (and should, I believe) be removed is:

```
	// Value.Referrers is a subset of the inverse of this
	// relation.  (Referrers are not tracked for all types of
	// Values.)
```

This is from the `ssa` package docs for the `Operands` method. What this boils down to is that for the purposes of rendering the SSA graph, `Referrers` edges are redundant.

I think it's still useful to keep the *O*range = *O*perand mnemonic, so I've kept the coloring.

Now, the reason why edges should be flipped is that I believe it makes the graph more readable. Up until this PR, referrers and operands edges pointed in different directions. This made for a lot of noise in the graph. The edges were also incoherent: referrers edges were oriented in a way that you could read an edge from `A` to `B` as `A is referred to by B`, whereas for operands edges the relationship could be read (backwards) as `B is an operand of A`. By flipping the edge, we get the more natural relationship `A is an operand of B`, which coincides with the referrers relationship.

---

Here's a before/after comparison:

## Before

image

## After

image

## Future work

In a future PR, I think it could be a good idea to remove the nodes/edges for `Const`s and `Function`s, as I don't believe they are particularly useful. Thoughts?


vinayakankugoyal (approved):
Great change!

PR #261, opened 2021-01-13
Fix panic in visitCall

mlevesquedion:
I'm still investigating the cause of this, just fixing the panic for now.

### DIFF
--- a/internal/pkg/propagation/propagation.go
+++ b/internal/pkg/propagation/propagation.go
@@ -258,6 +258,9 @@ func (prop *Propagation) visitCall(call *ssa.Call, maxInstrReached map[*ssa.Basi
 	// Source methods that return tainted values regardless of their arguments should be identified by the fieldpropagator analyzer.
 	if recv := call.Call.Signature().Recv(); recv != nil && sourcetype.IsSourceType(prop.config, prop.taggedFields, recv.Type()) {
 		visitingFromArg := false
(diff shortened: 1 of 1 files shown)


internal/pkg/propagation/propagation.go
```diff
@@ -258,6 +258,9 @@ func (prop *Propagation) visitCall(call *ssa.Call, maxInstrReached map[*ssa.Basi
 	// Source methods that return tainted values regardless of their arguments should be identified by the fieldpropagator analyzer.
 	if recv := call.Call.Signature().Recv(); recv != nil && sourcetype.IsSourceType(prop.config, prop.taggedFields, recv.Type()) {
 		visitingFromArg := false
+		if len(call.Call.Args) == 0 {
```
vinayakankugoyal, line 261: Any bug fix should have a unit test? Could we add a unit test for this particular case?

mlevesquedion:
This has happened too many times now, I will look into using Github Actions to run this check as a presubmit.

internal/pkg/propagation/propagation.go
mlevesquedion: I agree. My intent here is to fix the analyzer ASAP, because currently we have a panicking analyzer on `master`. There will be a follow-up PR with a detailed investigation + tests.

vinayakankugoyal (approved):

PR #262, opened 2021-01-13
Fix incorrect source identification logic

mlevesquedion:
Follow-up to #261.

## What's the bug?

The bug involves incorrect source identification logic in `sourcetype.IsSourceType`. In the following snippet taken from `propagation.visitCall`:

```go
    ...
    for _, a := range call.Call.Args[1:] {
```

I was expecting the slice operation on `Args` to be fine, because I thought `sourcetype.IsSourceType` would prevent the `if` from being entered when the receiver had an interface type. Indeed, interfaces are not supposed to be sources. However, the current code does not actually prevent interface types from being configured as sources:

```go
case *types.Named:
    return c.IsSourceType(utils.DecomposeType(tt)) || IsSourceType(c, tf, tt.Underlying())
```

If `tt` has `types.Interface` as `Underlying` type, and it matches one of the configured matchers, this will return `true`.

So in the snippet from `visitCall`, if an interface was configured as a source, and the method being called had no arguments, `call.Call.Args` would be empty, because `call.Call.Args` only contains the receiver if the receiver's type is statically known (which is not the case when the receiver has interface type).

## How did this panic come about?

The panic was unknowingly introduced in #252. The behavior that caused a panic wasn't exercised by our tests, nor by running the analyzer on kubernetes. Therefore, this error wouldn't have been caught by running a check on kubernetes in CI. (I am still planning to investigate GitHub Actions, because I think it would be good to have anyway). When I ran across this bug (while running on a different codebase, i.e. not kubernetes), my first thought was that I must have gone through the following sequence of actions:

1. Write #252 and open the PR. Run the check on kubernetes to confirm that the analyzer doesn't crash.
2. Review process occurs. I make a change that I (wrongly) think has no impact on the behavior and I forget to run the check again.

But I didn't make any such changes, so I didn't need to run the check again, and had I run it, it wouldn't have caught the bug.

## Next steps

We should consider making it "structurally" impossible for interface types to be configured as sources. There isn't much gain to being able to say that every field on a struct is sensitive by not providing a field matcher, and in practice, we don't use this facility in kubernetes, so we should consider making it mandatory to provide a field matcher. That way, it would be impossible to define a source matcher that matches an interface type, because interfaces have no fields.

### DIFF
--- a/internal/pkg/propagation/propagation.go
+++ b/internal/pkg/propagation/propagation.go
@@ -258,7 +258,6 @@ func (prop *Propagation) visitCall(call *ssa.Call, maxInstrReached map[*ssa.Basi
 	// Source methods that return tainted values regardless of their arguments should be identified by the fieldpropagator analyzer.
 	if recv := call.Call.Signature().Recv(); recv != nil && sourcetype.IsSourceType(prop.config, prop.taggedFields, recv.Type()) {
 		visitingFromArg := false
-		// This condition should never be true, because:
 		// Interface types cannot be sources. If the receiver is not a source, the
 		// above condition will be false, so this code won't be executed.
 		// When the receiver's type is statically known (it isn't an interface type),
--- a/internal/pkg/propagation/propagation.go
+++ b/internal/pkg/propagation/propagation.go
@@ -258,6 +258,15 @@ func (prop *Propagation) visitCall(call *ssa.Call, maxInstrReached map[*ssa.Basi
 	// Source methods that return tainted values regardless of their arguments should be identified by the fieldpropagator analyzer.
 	if recv := call.Call.Signature().Recv(); recv != nil && sourcetype.IsSourceType(prop.config, prop.taggedFields, recv.Type()) {
 		visitingFromArg := false
+		// This condition should never be true, because:
+		// Interface types cannot be sources. If the receiver is not a source, the
+		// above condition will be false, so this code won't be executed.
+		// When the receiver's type is statically known (it isn't an interface type),
+		// it will be the first element of the Args slice.
+		if len(call.Call.Args) == 0 {
+			fmt.Printf("call.Call.Args was empty; please report this issue")
+			return
+		}
 		for _, a := range call.Call.Args[1:] {
 			if prop.tainted[a.(ssa.Node)] {
 				visitingFromArg = true
--- a/internal/pkg/propagation/propagation.go
+++ b/internal/pkg/propagation/propagation.go
@@ -258,9 +258,6 @@ func (prop *Propagation) visitCall(call *ssa.Call, maxInstrReached map[*ssa.Basi
 	// Source methods that return tainted values regardless of their arguments should be identified by the fieldpropagator analyzer.
 	if recv := call.Call.Signature().Recv(); recv != nil && sourcetype.IsSourceType(prop.config, prop.taggedFields, recv.Type()) {
 		visitingFromArg := false
-		if len(call.Call.Args) == 0 {
-			return
-		}
 		for _, a := range call.Call.Args[1:] {
 			if prop.tainted[a.(ssa.Node)] {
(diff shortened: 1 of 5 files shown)


vinayakankugoyal (approved):

vinayakankugoyal (approved):

internal/pkg/sourcetype/sourcetype.go
```diff
@@ -26,14 +26,16 @@ import (
...
-// - A composite type that contains a Source Type
+// - A Named Struct Type that is configured as a Source
 // - A Struct Type that contains a tagged field
+// - A composite type that contains a Source Type
 func IsSourceType(c *config.Config, tf fieldtags.ResultType, t types.Type) bool {
 	deref := utils.Dereference(t)
 	switch tt := deref.(type) {
 	case *types.Named:
-		return c.IsSourceType(utils.DecomposeType(tt)) || IsSourceType(c, tf, tt.Underlying())
+		// only named structs can be *configured* sources
```
PurelyApplied, line 36: Is this an assertion we really want to make?

PurelyApplied, line 36: Is this an assertion we really want to make?

internal/pkg/sourcetype/sourcetype.go
```diff
@@ -26,14 +26,16 @@ import (
...
+		_, ok := tt.Underlying().(*types.Struct)
```
PurelyApplied, line 37: This feels more like something that belongs to `utils.Dereference`.  I'm not sure how well this will play with, e.g., `type SourceRef *Source`.

internal/pkg/propagation/propagation.go
```diff
@@ -258,7 +258,13 @@ func (prop *Propagation) visitCall(call *ssa.Call, maxInstrReached map[*ssa.Basi
 	// Source methods that return tainted values regardless of their arguments should be identified by the fieldpropagator analyzer.
 	if recv := call.Call.Signature().Recv(); recv != nil && sourcetype.IsSourceType(prop.config, prop.taggedFields, recv.Type()) {
 		visitingFromArg := false
+		// This condition should never be true, because:
```
PurelyApplied, line 261: Nit: This line feels superfluous.

PurelyApplied (commented):
The explanation in the body of this PR is exemplary.  The commit summary is a bit vague, though.  I'd prefer a commit along the lines of

```
Avoid panic in source identification.
* A dynamically dispatched ssa.Call can have empty Args.
```

mlevesquedion:
Thank you for your kind words! I did make a special effort to be especially thorough on this one.

I agree that the commit message could be improved, I'll be sure to integrate your suggestion when I squash.

internal/pkg/propagation/propagation.go
mlevesquedion: Done.

internal/pkg/sourcetype/sourcetype.go
mlevesquedion: Good point. I've integrated `utils.Dereference` to remove any pointer indirection on the `Underlying()` type.

internal/pkg/sourcetype/sourcetype.go
mlevesquedion: Good call, I believe the answer is no. I've reworked this snippet to deny interfaces rather than only accepting structs. Indeed, the following is a named basic type, which could also be configured as a source:

```go
type Password string
```

PurelyApplied (approved):
I'm not sure how I feel about asserting that Interface should not be used to identify sources.  I think it's fine for now, but I'd prefer to keep that as something we could back-peddle on if necessary, if there ends up being a push from users to enable that.

So LGTM overall, with one more drop in the bucket of future concerns that I shouldn't prematurely borrow from.

PR #265, opened 2021-01-28
Add quickstart user guide

mlevesquedion:
This PR adds a user guide that shows how to configure and run the analyzer. The intent is to provide a "quickstart" tutorial that allows users to quickly grasp how they can start applying the analyzer.


guides/quickstart.md
```diff
@@ -0,0 +1,175 @@
...
+		return nil, err
+	}
+	return response, nil
+}
+```
+
+This code is an instance of [CWE-522](https://cwe.mitre.org/data/definitions/522.html), "Insufficiently Protected Credentials".
+Indeed, the `Password` field on the `Authentication` struct contains credentials, which should not be allowed to reach logs.
+
+This example may seem a bit silly, but the analyzer has actually caught many similar cases in the wild.
```
vinayakankugoyal, line 33: This example may seem a bit silly -> This example may seem contrived

guides/quickstart.md
```diff
@@ -0,0 +1,175 @@
...
+	if err != nil {
+		log.Printf("unable to make authenticated request: incorrect authentication? %v\n", auth)
+		return nil, err
+	}
+	return response, nil
+}
+```
+
+This code is an instance of [CWE-522](https://cwe.mitre.org/data/definitions/522.html), "Insufficiently Protected Credentials".
+Indeed, the `Password` field on the `Authentication` struct contains credentials, which should not be allowed to reach logs.
```
vinayakankugoyal, line 31: Remove indeed?

guides/quickstart.md
```diff
@@ -0,0 +1,175 @@
...
+
+This example may seem a bit silly, but the analyzer has actually caught many similar cases in the wild.
+
+## Configuration
+
+Let's see how we can configure the analyzer to automatically detect this incorrect handling of credentials.
+
+In order to do its job, the analyzer needs 2 pieces of information:
+1. Which types hold sensitive data?
```
vinayakankugoyal, line 40: Which types hold sensitive data aka sources.

guides/quickstart.md
```diff
@@ -0,0 +1,175 @@
...
+2. Which functions should not be called with sensitive data?
```
vinayakankugoyal, line 41: Which functions should not be called with sensitive data? The are called sinks.

guides/quickstart.md
```diff
@@ -0,0 +1,175 @@
...
+### Configuring a Source type
+
+In taint propagation lingo, a value that contains data of interest is called a Source.
+In the current example, a Source is a value that contains sensitive data, e.g. a password.  
+
+There are 2 ways to define a Source type:
+1. Define a field tag and apply it to sensitive fields
+2. Describe a type using its name, the name of its package, and the relevant field(s)
+
+#### Using field tags
```
vinayakankugoyal, line 54: move this under 1.?

guides/quickstart.md
```diff
@@ -0,0 +1,175 @@
...
+```go
+type Authentication struct {
+  Username string 
+  Password string `datapolicy:"secret"`
+}
+```
+
+If you do not wish to define your own field tag, you may use the built-in `levee:"source"` tag.
+
+#### Using type descriptions
```
vinayakankugoyal, line 77: Move this section under 2.?

guides/quickstart.md
```diff
@@ -0,0 +1,175 @@
...
+
+## Conclusion
+
+In this example, we showed how the `go-flow-levee` analyzer can detect incorrect handling of
+credentials. However, note that the analyzer can be used for any kind of taint propagation,
+such as detecting code that is vulnerable to SQL injection.
+
+Also note that when the analyzer does not produce any reports on a codebase, it does *not* mean
+that the code base is "safe". There are inherent limitations to static analysis, so we recommend that
+you also use a form of dynamic analysis, such as a sanitizing logger. Furthermore, you should verify
```
vinayakankugoyal, line 169: no new line

guides/quickstart.md
```diff
@@ -0,0 +1,175 @@
...
+your configuration using simple examples like the one presented here to make sure that a lack of reports
```
vinayakankugoyal, line 170: no new line

guides/quickstart.md
```diff
@@ -0,0 +1,175 @@
...
+is not caused by a configuration error.
```
vinayakankugoyal, line 171: no new line

guides/quickstart.md
```diff
@@ -0,0 +1,175 @@
...
+
+Finally, in some cases the analyzer may produce reports that are in fact incorrect. Indeed, the analyzer
+does not attempt to model your program's behavior completely, and its analysis is currently limited to one
```
vinayakankugoyal, line 174: no new line

guides/quickstart.md
mlevesquedion: My intention was to list the options first, then explain them. WDY?T

guides/quickstart.md
mlevesquedion: I am not sure what you mean. :thinking:

guides/quickstart.md
```diff
@@ -0,0 +1,175 @@
...
+
+`auth.go`
+```go
+		log.Printf("unable to make authenticated request: incorrect authentication?\n")
+```
+
+After making this change, the analyzer no longer produces a report. There are many ways to
+address reports, the important thing is to modify the code so that the Source can no longer
+reach the Sink. For example, you may wish to define a "Sanitizer" function that removes the
+password, making the `auth` struct safe to log. See [the documentation](../configuration/README.md)
```
vinayakankugoyal, line 158: `Authentication`

guides/quickstart.md
vinayakankugoyal: I thought there was an unnecessary new line, but now I realized that there isn't one.

guides/quickstart.md
vinayakankugoyal: That works too.

guides/quickstart.md
```diff
@@ -0,0 +1,175 @@
+# Quickstart: How to configure and run the analyzer
```
vinayakankugoyal, line 1: I think this is also a good opportunity to introduce the levee analogy. WDYT?

mlevesquedion: Maybe. :thinking: I want this doc to be as concise as possible, and I'm not sure how to work that in.

It's a bit subtle, but the analogy is introduced in the `README`, at the top:

```
This static analysis tool works to ensure your program's data flow does not spill beyond its banks.
```

Do you have a suggestion?

guides/quickstart.md
mlevesquedion: Are you referring to `auth`? I think I prefer `auth` here, because that is IMO the most specific label, being the name of the variable in the example. WDYT?

guides/quickstart.md
vinayakankugoyal: Something like 
> before we get into the working of levee we want to explain why we decided to call it so. If you think of you programs data as water that flows from one place to another then like a levee ensures that the water flow does not spill beyond its banks, go-flow-levee ensures that your program's data flow does not spill beyond its banks.

vinayakankugoyal (approved):

guides/quickstart.md
vinayakankugoyal: It says making the `auth` struct safe to log. But there is no `auth` struct there is an `Authentication` struct.

guides/quickstart.md
mlevesquedion: Thank you. I'm not sure how that fits into "how to configure and run the analyzer", though. The intent for this guide is to show users how to get up and running quickly, so I don't think users will be interested in reading about the origin of the name of the tool.

@PurelyApplied, @benhxy, thoughts?

guides/quickstart.md
mlevesquedion: In my understanding, the `auth` variable is a `struct` of type `Authentication`. I see your point, though. The sanitizer would work with values of type `Authentication`.

I've reworded it. WDYT?

vinayakankugoyal: I am thinking do we really need the line
`For example, you may wish to define ...`

guides/quickstart.md
```diff
@@ -0,0 +1,175 @@
...
+		return nil, err
+	}
+	return response, nil
+}
+```
+
+This code is an instance of [CWE-522](https://cwe.mitre.org/data/definitions/522.html), "Insufficiently Protected Credentials".
+The `Password` field on the `Authentication` struct contains credentials, which should not be allowed to reach logs.
+
+This example may seem contrived, but the analyzer has actually caught many similar cases in the wild.
```
PurelyApplied, line 33: Nit: Consider "this analyzer" over "the analyzer."

guides/quickstart.md
PurelyApplied: I feel like we would expect a user to arrive here from the root `README.md`, which already contains that analogy in the header.  I'm not sure repeating it here is necessary.  \</2¢\>

guides/quickstart.md
```diff
@@ -0,0 +1,175 @@
...
+Given this information, the analyzer will perform taint propagation to determine whether sensitive values can reach calls to functions that they shouldn't reach.
+
+### Configuring a Source type
+
+In taint propagation lingo, a value that contains data of interest is called a Source.
+In the current example, a Source is a value that contains sensitive data, e.g. a password.  
+
+There are 2 ways to define a Source type:
+1. Define a field tag and apply it to sensitive fields
+2. Describe a type using its name, the name of its package, and the relevant field(s)
```
PurelyApplied, line 52: Nit: Markdown rendering ignores enumeration values and automatically renders `1. \\ 2. \\ `.  I believe the canonical approach is to use `1.` for all enumeration entries so that you don't have to manually re-enumerate them if an item is added or moved, and to let the renderer handle the actual incrementing of values.

guides/quickstart.md
```diff
@@ -0,0 +1,175 @@
...
+
+#### Using field tags
+
+In the analyzer's configuration, define the field tag:
+
+`analyzer_configuration.yaml`
+```yaml
+FieldTags:
+  - Key: datapolicy
+    Val: secret
```
PurelyApplied, line 62: While we're currently supporting `Val` so we don't have to rewrite existing configs, I for one would prefer to use `Value` as the canonical config key.

guides/quickstart.md
```diff
@@ -0,0 +1,175 @@
...
+If you do not wish to define your own field tag, you may use the built-in `levee:"source"` tag.
+
+#### Using type descriptions
+
+In the analyzer's configuration, identify the sensitive data:
+
+`analyzer_configuration.yaml`
+```yaml
+Sources:
+  - Package: auth
```
PurelyApplied, line 84: This feels a little misleading, because `Package` expects a complete path to the package, not just the tail of it.  Using this in the example config instead of the `FieldTags` does not produce a diagnostic.

I know you link the `configuration/README.md` right after, but consider expanding this to `Package: complete/path/to/auth`.  I would prefer to leave the partial / regexp matching to be explained in the `configuration/README.md` and not complicate that here.

This is a weird one, since we give the `package auth` in the example above, but without context of surrounding `GOPATH` / `go.mod`, the package path is incomplete.  I suppose you could use the complete path `github.com/google/go-flow-levee-guide/guides/quickstart/auth`, since the code is being introduce here.  If you go that route, maybe add the path as a comment to the `package auth`, i.e. 
```golang
// complete path github.com/ ...
package auth
```

guides/quickstart.md
```diff
@@ -0,0 +1,175 @@
...
+    Type: Authentication
+    Field: Password
+```
+
+This method does not require changes to the code.
```
PurelyApplied, line 89: Consider dereferencing the antecedant, i.e. "This method" -> "This method of configuration."

guides/quickstart.md
```diff
@@ -0,0 +1,175 @@
...
+### Configuring a Source type
+
+In taint propagation lingo, a value that contains data of interest is called a Source.
+In the current example, a Source is a value that contains sensitive data, e.g. a password.  
+
+There are 2 ways to define a Source type:
+1. Define a field tag and apply it to sensitive fields
+2. Describe a type using its name, the name of its package, and the relevant field(s)
+
+#### Using field tags
```
PurelyApplied, line 54: I might plug something like "This style of configuration is recommended for its maintainability."

guides/quickstart.md
```diff
@@ -0,0 +1,175 @@
...
+
+(This command assumes that it is being run in the `go-flow-levee/guides/quickstart` directory.)
+
+Running on the example code, you should see output similar to the following:
+
+```
+# github.com/google/go-flow-levee/guides/quickstart
+./auth.go:13:13: a source has reached a sink
+ source: ./auth.go:10:19
+```
```
PurelyApplied, line 143: I don't think there's anything for it, but this documentation is going to go stale pretty quickly.  Something to keep in mind down the road.

guides/quickstart.md
```diff
@@ -0,0 +1,175 @@
...
+
+The code for this example is available in the `go-flow-levee/guides/quickstart` directory.
+
+## Context
+
+Suppose you have the following piece of code:
+
+`auth.go`
+```go
+package auth
```
PurelyApplied, line 11: Nit: For filenames in Markdown, I feel like I've typically seen it as a leading comment, i.e.

```golang
// auth.go
package auth
// ...
```

The spacing staggers a bit otherwise.

PurelyApplied (changes requested):
Looks pretty good.  I've got a handful of nits, but this is a good addition.

guides/quickstart.md
mlevesquedion: I hope it does :sweat_smile:
We'll update it when it does

mlevesquedion:
Thank you for your comments! PTAAL.

mlevesquedion:
I've changed the config to use `Value` instead of `Val` when configuring the field tag, as per discussion here: https://github.com/google/go-flow-levee/pull/266#discussion_r567106506

I'm marking this PR a draft until that's merged. (Currently, using `Value` causes the configuration to be invalid, i.e. the tagged field won't be detected.)

PurelyApplied (approved):

PR #266, opened 2021-01-28
Enforce presence of Key and Value keys on field tag definition

mlevesquedion:
I ran into this issue while writing the quickstart user guide (#265).

Originally when I was writing the config, I wrote:

```go
FieldTags:
  - Key: datapolicy
    Value: password
```

Currently, we "allow" `Value` as a "valid" key, but in reality, if you use `Value` instead of `Val`, the `Val` will be empty and your matcher won't actually match anything. In fact, we don't check that the fields are present, only that if some fields are present, they are one of `key, val, value`.

So e.g. if you write:

```yaml
FieldTags:
  - Foo: Bar
```

the code will complain about `Foo` being an invalid key, but if you write this:

```yaml
FieldTags:
  - Key: Key
```

it won't complain about the missing `Val` field, but your matcher won't be able to match anything. (Unless your value really is empty, but I don't think we should support that use case.)

This PR fixes the issue by checking that the parsed `Key` and `Val` are non-empty strings, and producing an error if that isn't the case.


internal/pkg/config/config.go
```diff
@@ -147,8 +147,9 @@ type fieldTagMatcher struct {
 
 // this type uses the default unmarshaller and mirrors configuration key-value pairs
 type rawFieldTagMatcher struct {
```
vinayakankugoyal, line 149: Wont this allow to be parsed?
FieldTags:
  - Key: datapolicy
    Val: cert
    Value: password

internal/pkg/config/matcher_test.go
```diff
@@ -287,19 +287,38 @@ func TestFieldTagMatcherUnmarshalling(t *testing.T) {
...
+		},
+		{
+			desc: "valid field tag config using val",
 			yaml: `
 key: good
 val: AlsoGood`,
 			wantErr: false,
 		},
+		{
+			desc: "valid field tag config using value",
```
vinayakankugoyal, line 316: can we add a case where both Val and Value are set?

internal/pkg/config/config.go
mlevesquedion: That is correct. In that case we should probably do like we are doing with e.g. `Package` and `PackageRE`, and only allow one or the other to be set. Although, to be honest I see little value in having both `Val` and `Value`, so I think we may actually want to keep only one of them. WDYT?

vinayakankugoyal: Yeah lets just keep `Value` and treat setting `Val` as an error. Being strict while parsing the config is totally fine IMO.

internal/pkg/config/matcher_test.go
vinayakankugoyal: This test wont be necessary if we only allow `Value` and not `Val`.

internal/pkg/config/config.go
mlevesquedion: @PurelyApplied, @benhxy, thoughts on this?

PurelyApplied: Oh, I had a WIP branch that resolved some of this with `RawMessage` and `switch` statements.  \*dig dig dig\*

#271 opened as draft.  I hadn't added tests yet, but it hits some of the concerns you've voiced.  There's also the issue where, if we're case insensitive, you could theoretically add all of `key`, `Key`, `KEY`, etc and have collisions that way while still being valid YAML.  #271 pokes at the other matchers, too, since I wasn't a huge fan of that Valid Keys slice.

> That is correct. In that case we should probably do like we are doing with e.g. Package and PackageRE, and only allow one or the other to be set. Although, in my opinion there is little value in having both Val and Value, so I think we may actually want to keep only one of them. WDYT?

I'd be +1 to only retaining `Value`.  I think we'd kept `Val` because we didn't want to rewrite existing test configs.  But we're v0.x.  We can make a config-breaking change here.

.github/workflows/ci.yaml
```diff
@@ -53,7 +53,8 @@ jobs:
           staticcheck ./...
 
       - name: Check YAML
-        uses: ibiqlik/action-yamllint@v3
+        run: |
+          yamllint .
```
mlevesquedion, line 57: I had to change this because the action was reporting an error, while running `yamllint` locally did not. As it turns out, Github Actions runners have yamllint installed by default, so we don't need to use a third-party action.

PurelyApplied (commented):
LGTM.  There's still plenty of cleanup that could be done, but that could wait for a future PR / until I clean up #271.

We still have a case insensitivity issue that could use a test, e.g.
```yaml
FieldTags:
  - Key: levee
    key: also-levee
    Value: source
    value: ExtraSourcey
```

PurelyApplied (approved):

PR #271, opened 2021-01-29
Refactor config Unmarshal methods.

PurelyApplied:
* Use json.RawMessage to clean logic distinguishing literal and regexp matchers
* Add tag validation that both key and value are set.

### DIFF
--- a/internal/pkg/config/config.go
+++ b/internal/pkg/config/config.go
@@ -145,38 +145,33 @@ type fieldTagMatcher struct {
 	Val string
 }
 
-// this type uses the default unmarshaller and mirrors configuration key-value pairs
-type rawFieldTagMatcher struct {
-	Key string
-	Val string
-}
-
 func (ft *fieldTagMatcher) UnmarshalJSON(bytes []byte) error {
-	validFieldTagMatcherFields := []string{"key", "val", "value"}
-	if err := validateFieldNames(&bytes, "fieldTagMatcher", validFieldTagMatcherFields); err != nil {
-		return err
-	}
-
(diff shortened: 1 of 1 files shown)


internal/pkg/config/config.go
```diff
@@ -145,38 +145,33 @@ type fieldTagMatcher struct {
...
-}
+	for key, val := range raw {
+		switch strings.ToLower(key) {
+		case "key":
+			if ft.Key != "" {
+				return fmt.Errorf("Multiple values given for 'key'.")
+			}
+			ft.Key = val
+		case "val", "value":
+			if ft.Key != "" {
```
PurelyApplied, line 161: Obvious typo

internal/pkg/config/config.go
```diff
@@ -188,46 +183,69 @@ type sourceMatcher struct {
...
-
 func (s *sourceMatcher) UnmarshalJSON(bytes []byte) error {
-	validSourceMatcherFields := []string{"package", "packageRE", "type", "typeRE", "field", "fieldRE"}
-	if err := validateFieldNames(&bytes, "sourceMatcher", validSourceMatcherFields); err != nil {
-		return err
-	}
+	// Unspecified matchers match vacuously
+	s.Package = vacuousMatcher{}
+	s.Type = vacuousMatcher{}
+	s.Field = vacuousMatcher{}
```
PurelyApplied, line 190: Might read better as `= nil`, use ` != nil` checks below, and update `nil` to `vacuous` during final validation.  But it's a draft.

PR #278, opened 2021-02-03
Introduce analysistest testdata convention.

### LINKED ISSUE #274: Proposal for testdata convention - spoof source root with go.mod to assist IDEs
## Problem statement

`analysistest` expects `testdata` to be a fake ~~GOROOT~~GOPATH.  We currently have various tests at `.../testdata/src/example.com/...` to test the various analyzers.

IntelliJ expects our project to be configured with Go modules.  This fake root is opaque to IntelliJ.  As a consequence, the IDE can't resolve the linkage for packages within `testdata`.  E.g.,

image

This conflict makes development in of tests sticker than it needs to be.  Additionally, while I believe I'm the only one who does this, but it also prevents general execution of an analyzer from the command-line.  I am partial to using `ssadump` when investigating issues, but this is prevented in any file with an unresolved import.

## Proposal

I propose adding a trivial `go.mod` to each `testdata/src/<some shared root>` so that our IDEs can resolve these paths.  See #273 and associated branch for an example within `config/testdata`.  This allows proper linking within IntelliJ while retaining test compatability.

For compatability with the builder within `analysistest`, we would each `testdata/src/...` to be self-contained.  Cross-package imports are permitted, but not cross-module imports.  This requires a shared root within `testdata/src/`.  #273 demonstrates this as `testdata/src/github.com/`.  In practice, we should avoid collision of go modules and adopt adopt `testdata/src/<identifier>`, e.g. `testdata/src/levee_configtest/...`.

I'd love to hear your thoughts about adding this as a convention going forward.  At your convenience, please pull my `pseudo-module-testing` branch and see if this works with your own workflow.  (I had to execute `config_test` to trigger compiling and linkage.)

PurelyApplied:
```
* Add Project Conventions to CONTRIBUTING.md and document this testdata convention.
* Move analysistest testdata from **/testdata/src/** to **/testdata/src/*_analysistest/**.
* Add go.mod for each *_analysistest
* Update relevant analysistest configurations.
* Update relevant *_test.go analysistest.Run package target strings.
* Update impacted imports.

Fixes #274
```

---

    -   Currently failing on `master` with my current config.  Need to revise after the `FieldTags` only accepting `Value` I think.
- [x] Appropriate changes to ~~README~~ CONTRIBUTING are included in PR


vinayakankugoyal (approved):

mlevesquedion:
Why remove the "unnecessary" config element in this PR? This removal does not seem related to the other changes in the PR.

The config element is actually necessary, let's discuss on #279.

mlevesquedion (commented):
LGTM, apart from the removal of the `SourceManipulator` config entry, which doesn't belong in this PR. Let's discuss in #279.

mlevesquedion (approved):

mlevesquedion:
Thank you for setting this up, this is great! :tada:

Issue #155, opened 2020-10-19
Fix bugs involving sanitization


mlevesquedion:
Consider the following case:
```go
func TestTaintedInIfButSanitizedBeforeIfExit() {
	var e interface{}
	if false {
		e = core.Source{}
		e = core.Sanitize(e)[0]
	}
	core.Sink(e)
}
```

With the current code, `e` is not considered sanitized when it reaches `core.Sink`. This is because:
1. Traversal goes _through_ the sanitizer, which gives a path to `core.Sink`.
2. The call to `core.Sanitize` does not dominate the call to `core.Sink`, because one can get from `var e interface{}` to `core.Sink(e)` without going through the loop. However, in that path `e` is not tainted.

For this first case, correct behavior can be obtained by stopping traversal upon reaching a sanitizer. The fact that the sanitizer has been reached still needs to be recorded though, so that for other cases involving pointers, domination can be checked.

However, even if we don't traverse _through_ the sanitizer, the following case still poses an issue:
```go
func TestPointerTaintedInIfButSanitizedBeforeIfExit() {
	var e interface{}
	if false {
		s := &core.Source{}
		core.SanitizePtr(s)
		e = s
	}
	core.Sink(e)
}
```
In this case, even if we stop traversing when we hit `SanitizePtr`, there is still a path from `s`, to `e` (via `e = s`), to `core.Sink`. Then, because the call to `core.SanitizePtr` does not dominate the call to `core.Sink`, `e` is not considered sanitized.

PR #280, opened 2021-02-06
Do not propagate taint through sanitizers.

PurelyApplied:
Incomplete fix for #155, found during iteration for taint tracing.


mlevesquedion:
Good find!

I vaguely remember trying a similar fix a while ago, and I vaguely remember running into some issues with it, but I must be remembering the specifics wrong. I am fairly confident in our test coverage here.

LGTM.

mlevesquedion (approved):

PR #283, opened 2021-02-16
Allow report suppression via comments

### LINKED ISSUE #198: Allow suppression of false positives via code comments.
This is a continuation of #87.

At time of writing, report of diagnostics can be suppressed at a function-level granularity.  While functional, this has potential to suppress a wider breadth of reports than intended.

Suppression via code comment, e.g. `nolint:levee <Justification>` would allow fine-grained suppression by immediately preceding a sink call.  Optionally, we may consider suppression of source identification or taint propagation.

mlevesquedion:
Fixes #198.

I went with `levee.DoNotReport` as a suppression string. I think the name is quite clear and the syntax is Go-like (it looks sort of like calling a public function from the `levee` package, without the parens), which may help developers remember it.

## Future work

I think it would be good to produce a warning when a suppression doesn't actually suppress anything. I see two main reasons to do this:
  * If the erroneous suppression isn't removed, and the code changes such that there is now actually a report being suppressed, the developers won't know about it.
  * From a maintenance/readability point of view, a suppression comment that doesn't suppress anything is misleading and unnecessarily clutters the code.


internal/pkg/suppression/analyzer.go
```diff
@@ -0,0 +1,72 @@
...
+	for _, line := range strings.Split(commentGroup.Text(), "\n") {
+		trimmed := strings.TrimSpace(strings.TrimPrefix(strings.TrimPrefix(line, "//"), "/*"))
+		if strings.HasPrefix(trimmed, suppressionString) {
+			return true
+		}
+	}
+	return false
+}
+
+const suppressionString = "levee.DoNotReport"
```
vinayakankugoyal, line 72: move to top?

internal/pkg/suppression/analyzer.go
```diff
@@ -0,0 +1,72 @@
...
+			for _, cg := range commentGroups {
+				if isSuppressingCommentGroup(cg) {
+					result[node] = true
+					pass.Reportf(node.Pos(), "suppressed")
+				}
+			}
+		}
+	}
+
+	return ResultType(result), nil
```
cindy52, line 57: Why cast the result type here instead of initialing the result to be ResultType in line 44?

internal/pkg/suppression/analyzer.go
mlevesquedion: I know constants are typically around the top of the file, but I kinda like having it here because it is a deep implementation detail. The exact value of this constant is not very important. WDYT?

internal/pkg/suppression/analyzer.go
mlevesquedion: My thinking is that the exact place we do it doesn't matter *too* much, but it has to be done before returning, so why not do it right at the end. This is how it's done in the existing fieldtags analyzer too: 
```go
	// return all facts accumulated down the current path in the dependency graph
	result := map[types.Object]bool{}
	for _, f := range pass.AllObjectFacts() {
		result[f.Object] = true
	}

	return ResultType(result), nil
```

Do you see any pros/cons to doing it when the `result` variable is initialized on line 44?

internal/pkg/suppression/analyzer.go
PurelyApplied: I'd see it as one of the following:

* If it's a package `const`, convention holds that these appear early in the file.
* If it's an implementation detail, it could be in-lined in its sole consumption.  The compiler knows to `const` a string literal.

I'd personally lean towards the former, but either way would consider an alternative to having a mid-file package-level `const`.

configuration/README.md
```diff
@@ -92,6 +92,27 @@ Since no receiver matcher was provided, it will match any method beginning with
...
+mysinks.SinkF("here's a value: %v", safeValue)
+```
+
+You may include a justification in the suppression comment, e.g.:
+
+```
+// levee.DoNotReport: I have verified that "safeValue" cannot actually contain sensitive data (mlevesquedion, Feb 05 2021).
+```
+
+In fact, as long as the exact string `levee.DoNotReport` appears at the beginning of a line in a comment above or directly on the line where a report would be produced, the report will be suppressed.
```
PurelyApplied, line 110: Elaborate about not handling suppression comments in the middle of multi-line calls.  See test comment.

internal/pkg/levee/levee.go
```diff
@@ -69,15 +77,32 @@ func run(pass *analysis.Pass) (interface{}, error) {
 	return nil, nil
 }
 
-func reportSourcesReachingSink(conf *config.Config, pass *analysis.Pass, propagations map[*source.Source]propagation.Propagation, sink ssa.Instruction) {
+func reportSourcesReachingSink(conf *config.Config, pass *analysis.Pass, suppressedNodes suppression.ResultType, propagations map[*source.Source]propagation.Propagation, sink ssa.Instruction) {
```
PurelyApplied, line 80: This could be work for a future PR, but it could be worth printing or emitting a `Diagnostic` if we hit a point where suppression was not needed.

internal/pkg/levee/levee.go
```diff
@@ -69,15 +77,32 @@ func run(pass *analysis.Pass) (interface{}, error) {
...
 			break
 		}
 	}
 }
 
+func isSuppressed(pos token.Pos, suppressedNodes suppression.ResultType, pass *analysis.Pass) bool {
+	for _, f := range pass.Files {
+		if pos < f.Pos() || f.End() < pos {
+			continue
+		}
```
PurelyApplied, line 93: Prefer `pass.Fset.Position(pos).Filename` to get the filename to avoid the loop.  Check for `pos != 0`.

internal/pkg/levee/levee.go
```diff
@@ -69,15 +77,32 @@ func run(pass *analysis.Pass) (interface{}, error) {
...
+		path, _ := astutil.PathEnclosingInterval(f, pos, pos)
+		if len(path) < 2 {
+			return false
+		}
```
PurelyApplied, line 97: This logic was not obvious to me before pulling up the `astutil` documentation.  Consider adding the comment `// Get all ast ancestors of this pos`.  Consider renaming `path` to `ancestors` for consistency with doc.

internal/pkg/suppression/analyzer.go
PurelyApplied: I'd say the distinction is philosophical.  `ResultType` is essentially the exposed API.  It is not meant to be consumed as a `map`.  The underlying type as a `map` is an implementation detail.

If we _were_ to intend the consumption to be done as a `map`, I'd say that we should either not declare `ResultType` at all, or declare it as an alias, i.e. `type ResultType = map [...]...`.  At which point, the declaration one way or the other would be immaterial.

If we are considering it an implementation detail, then I'd say it comes down to preference as to whether you think those details matter in the moment.  In this method, we are using its map-yness, so it makes sense to me to declare it as a `map` and wrap it in the API layer when `return`ing.  On the other hand, that implementation detail is will in scope within this package, and declaring it as a `ResultType` doesn't prohibit you from using it like a `map`.

Personally, I'd prefer the `result := make(ResultType)` declaration and drop the cast on `return`, but I can see it go either way.

PurelyApplied (commented):
Good work.  I should really get more comfortable with the directly examining the `ast`, because I probably would've made this a lot harder than necessary if I were to have done it myself.

internal/pkg/levee/levee.go
mlevesquedion: I don't think we can avoid the loop here. If we get the `Filename` then we can use that instead of  `f.Pos()` and `f.End()` to find the appropriate `*ast.File`, but AFAIK there isn't an existing function to get the `*ast.File` from the filename. :thinking:

mlevesquedion:
Thank you! A prior iteration of this code was fairly gnarly, I tried doing it without `ast.NewCommentMap` at first, which turned out to be trickier than expected.

internal/pkg/levee/levee.go
PurelyApplied: Whoops.  Don't know why I thought those were `pass.Files` were `string`s.  Yeah, disregard.

internal/pkg/levee/testdata/src/levee_analysistest/example/suppression/tests.go
```diff
@@ -0,0 +1,85 @@
...
+}
+
+func TestReportsSeparatedByGeneralComment(s core.Source) {
+	core.Sink(s) /*
+		levee.DoNotReport
+	*/
+	core.Sink(s) // want "a source has reached a sink"
+}
+
+func TestLineCommentBeforeGeneralComment(s core.Source) {
```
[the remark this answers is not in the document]
mlevesquedion: Good test cases, thanks! I did some more digging around this and I ran into some weird things, e.g.:

```go
	core.Sink( // levee.DoNotReport
		"arg 1",
		s,
		"arg 3")

```

What node is the comment associated with in this example? That's right! The `Sink` identifier! Which makes sense, but as a user I would expect this to suppress the report.

It turns out it's straightforward to support that case with the information that we have in the `ast.CallExpr` (first node returned by `PathEnclosingInterval`).

We could support suppression on arguments as well, but I feel like that might be misleading, e.g. a user might think "here, let me suppress *just* this specific argument" and not realize that it suppresses the entire call. Correctly handling suppression in `Args` requires inspecting the entire tree below the `ast.CallExpr`. Indeed, consider this case:

```go
core.Sink(
  []interface{}{
    s, // levee.DoNotReport
  })
```

(In the example above, it's not sufficient to just look at the nodes in `ast.CallExpr.Args`, and in the general case we really need to dig into the tree.)

So we could definitely suppress the entire call when the suppression is associated with a specific argument, but that feels misleading to me. The following case is a bit unfortunate, though:

```go
core.Sink(s,  // levee.DoNotReport
  "foo", "bar")
```

In the above case, the comment is associated with `s`, not with the call, so currently this doesn't suppress the report. (But some might expect it to.)

I think as long as this is clearly documented, this is reasonable behavior. WDYT?

internal/pkg/levee/levee.go
mlevesquedion: Agreed! (This possibility is in fact mentioned in the PR description.)

internal/pkg/levee/levee.go
mlevesquedion: I personally prefer `path` because the function is called `Path...`, so it feels like a natural name.

I've added a comment which I think describes the behavior in sufficient detail for our use case.

WDYT?

internal/pkg/levee/testdata/src/levee_analysistest/example/suppression/tests.go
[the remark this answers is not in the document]
PurelyApplied: > I think as long as this is clearly documented, this is reasonable behavior. WDYT?

I think that's one of the two guiding stars.

I'd say the other one is to start simple.  I'm perfectly happy to have a more restrictive use pattern and, should we get requests for finer-grained control, iterate then.

Here I'm asking because I'm still not familiar with the `ast` graph, but how would it look if we looked at all the ancestry and descendants at a given sink call?  (Would that even be possible without a huge rework?)  To my crude understanding, we'd end up with three cases:
* The comment is on an descendant argument, so we exclude the call entirely (and document that it doesn't matter which arg it's on)
* The comment is on the call itself, so is naturally excluded
* The comment is on an ancestor, implying broader exclusion.

That last bullet makes me think we could get from a comment the sort of exclusion possible in configuration now - a function-level comment would exclude all contained calls, and possible a package-level comment could exclude everything there.  That's probably opening a granularity can of worms, so I could be absolutely fine not going that route.

It also brings into question nested calls.  What happens in the following case?

```golang
func SinkAndReturnError(s core.Source) error {  // <- This is declared as a sink
  return fmt.Errorf("%v", s)
}

func notASink(interface{}) {}

func TestInnerSink(s core.Source) {
  notASink(
    SinkAndReturnError(s)  // levee.DoNotReport
  )
}
```

Ignoring the taint propagation aspect, that might be a case we want to cover.  Also reasonable as a candidate for iteration, but something we should document one way or t'other.

configuration/README.md
mlevesquedion: Done.

internal/pkg/suppression/analyzer.go
mlevesquedion: That all makes sense to me. If there are no strong feelings here, I'll keep it the way it currently is since that's also how it is in `fieldtags/analyzer.go`.

internal/pkg/levee/testdata/src/levee_analysistest/example/suppression/tests.go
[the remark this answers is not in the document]
mlevesquedion: We definitely can handle both args (visit the tree below the call) and ancestors (use `astutil.PathEnclosingInterval`), but for now in my opinion we don't want either of those.

Concerning nested calls, good point. The results are interesting.

This works:

```go
	fmt.Println(
		// levee.DoNotReport
		core.SinkAndReturn(s),
	)
```

This doesn't work:

```
	fmt.Println(
		core.SinkAndReturn(s), // levee.DoNotReport
	)
```

In the above, the comment is associated, rather suprisingly, with the `s` identifier (`s` is a `core.Source` value).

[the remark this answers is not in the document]
PurelyApplied: > For now, I think it would be reasonable to open an issue for that second case, and include it in the tests with an appropriate TODO referencing the issue. WDYT?

Iterating on this sounds good to me.

internal/pkg/levee/levee.go
PurelyApplied: > I personally prefer path because the function is called Path..., so it feels like a natural name.

Fair enough.  The clarifying comment was the real ask, and we have one now.  👍

internal/pkg/levee/testdata/src/levee_analysistest/example/suppression/tests.go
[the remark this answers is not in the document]
mlevesquedion: Good. I've added an issue.

internal/pkg/levee/levee.go
```diff
@@ -69,15 +78,53 @@ func run(pass *analysis.Pass) (interface{}, error) {
...
+		// position, from the leaf node that directly contains it up to the ast.File node
+		path, _ := astutil.PathEnclosingInterval(f, pos, pos)
+		if len(path) < 2 {
+			return false
+		}
+		// Given the position of a call, path[0] holds the ast.CallExpr and
+		// path[1] holds the ast.ExprStmt. A suppressing comment may be associated
+		// with the name of the function being called (Ident, SelectorExpr), with the
+		// call itself (CallExpr), or with the entire expression (ExprStmt).
+		if ce, ok := path[0].(*ast.CallExpr); ok {
```
PurelyApplied, line 105: Can this type assertion fail?  What does it mean if it does?

mlevesquedion: Right now, in my understanding, no. Since sinks are functions and we report on calls to those functions, the first element of `path` should be an `*ast.CallExpr`. I guess I'm just trying to be defensive here and avoid a panic in case those assumptions ever change.

PurelyApplied: Gotcha.  I agree we probably shouldn't panic, but it might be good to add a warning in that unexpected case.  We have a few of those throughout the codebase, e.g. `propagation.go:215: fmt.Printf("unexpected node received: %T %v; please report this issue\n", n, n)`

PurelyApplied (approved):

PR #286, opened 2021-02-25
Always run all lint checks

mlevesquedion:
I got the idea for this PR from #285.

When running lints in Github Actions, currently the `CI / Lint` job stops on the first failure. I think it would be preferable to always run all checks. That way if there are errors in multiple lints, they can all be observed and corrected with a single run, instead of having to repeat the "push --> check lint --> fix" cycle multiple times.

I tried using `GO111MODULE=off` on `go get` and `go install` invocations to ensure `go.{mod,sum}` wouldn't be modified when installing `addlicense` and `staticcheck`, but it doesn't work in GithubActions and I don't believe it's worth debugging now. Github Actions steps run serially, and there is currently no way to run them in parallel (cf: https://github.community/t/steps-in-parallel/16343). As long as the `Check go mod` step is before the steps that install `addlicense` and `staticcheck`, this won't cause a spurious failure.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -27,14 +27,16 @@ jobs:
         uses: actions/checkout@v1
 
       - name: Check goimports
+        if: ${{ always() }}
         run: |
-          go install golang.org/x/tools/cmd/goimports
+          GO111MODULE=off go install golang.org/x/tools/cmd/goimports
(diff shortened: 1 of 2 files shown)


vinayakankugoyal (approved):

PurelyApplied (approved):
Ahh, glad they have "still don't pass" alternative to `continue-on-error`.

PR #287, opened 2021-03-03
Do not visit referrers of non-string basic-typed values

### LINKED ISSUE #191: Determine which Basic types can't meaningfully be tainted

mlevesquedion:
Fixes #191.

We've known that this could cause problems for a while, but we never hit an actual false positive because of it until recently. Motivating example:

```go
core.Sink(len(source.Data))
```

(where `source` has type `core.Source`)

See #191 for the list of basic types. My assessment is that `string`s are the only basic type that can meaningfully be tainted. Note that e.g. `[]byte` is not a basic type, its type is `Slice`, and its `Elem()` is `byte`. This does mean however that we won't catch pathological cases such as the following:

```go
for _, b := range "password" {
        core.Sink(b)
}
```

(Imagine that `core.Sink` behaves like `fmt.Print`, e.g.)

I've never observed anything like the above in the wild, so I think avoiding non-string basics is reasonable.

Something about this PR may surprise you: why move the `hasTaintableType` check from `shouldNotTaint` to `taintReferrers`? Consider this example:

```
func TestCopyPropagatesTaintFromSrcToDst(s core.Source) {
	b := make([]byte, len(s.Data))
	bytesCopied := copy(b, s.Data)
	core.Sink(bytesCopied)
	core.Sink(b) // want "a source has reached a sink"
}
```

What is the type of the `Call` to `copy` in the above? That's right, `int`. If we check whether the type is taintable in `shouldNotTaint`, we'll conclude that we shouldn't taint the `Call`, which means we won't propagate any taint through `copy`. This can occur for any other function that returns a non-string basic type. I think it's fine to taint a `bool` or an `int` (e.g.), as long as we don't then propagate to that `bool` or `int`'s referrers.

### DIFF
--- a/internal/pkg/propagation/propagation.go
+++ b/internal/pkg/propagation/propagation.go
@@ -102,10 +102,6 @@ func (prop *Propagation) shouldNotTaint(n ssa.Node, maxInstrReached map[*ssa.Bas
 		return true
 	}
 
-	if !hasTaintableType(n) {
-		return true
-	}
-
 	if instr, ok := n.(ssa.Instruction); ok {
 		instrIndex, ok := indexInBlock(instr)
 		if !ok {
@@ -225,6 +221,10 @@ func (prop *Propagation) taintField(n ssa.Node, maxInstrReached map[*ssa.BasicBl
 }
 
 func (prop *Propagation) taintReferrers(n ssa.Node, maxInstrReached map[*ssa.BasicBlock]int, lastBlockVisited *ssa.BasicBlock) {
+	if !hasTaintableType(n) {
+		return
+	}
+
 	if n.Referrers() == nil {
 		return
 	}
@@ -417,7 +417,7 @@ func hasTaintableType(n ssa.Node) bool {
 	if v, ok := n.(ssa.Value); ok {
 		switch t := v.Type().(type) {
 		case *types.Basic:
-			return t.Info() != types.IsBoolean
+			return (t.Info() & types.IsString) != 0
 		case *types.Signature:
 			return false
 		}
--- a/internal/pkg/levee/testdata/src/levee_analysistest/example/tests/basictypes/tests.go
+++ b/internal/pkg/levee/testdata/src/levee_analysistest/example/tests/basictypes/tests.go
@@ -1,0 +1,38 @@
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package basic
+
+import (
+	"levee_analysistest/example/core"
+)
+
+func BooleansDontPropagateTaint(s *core.Source) {
+	is := isSourcey(s)
+	core.Sink(is)
+}
+
+func isSourcey(x interface{}) bool {
+	// not taking any chances
+	return true
+}
+
+func IntegersDontPropagateTaint(sources []core.Source) {
+	sourceCount := countSources(sources)
+	core.Sink(sourceCount)
+}
+
+func countSources(sources []core.Source) int {
+	return len(sources)
+}
--- a/internal/pkg/levee/testdata/src/levee_analysistest/example/tests/basictypes/tests.go
+++ b/internal/pkg/levee/testdata/src/levee_analysistest/example/tests/basictypes/tests.go
@@ -29,10 +29,5 @@ func isSourcey(x interface{}) bool {
 }
 
 func IntegersDontPropagateTaint(sources []core.Source) {
-	sourceCount := countSources(sources)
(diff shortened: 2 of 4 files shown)


internal/pkg/propagation/propagation.go
```diff
@@ -225,6 +221,10 @@ func (prop *Propagation) taintField(n ssa.Node, maxInstrReached map[*ssa.BasicBl
 }
 
 func (prop *Propagation) taintReferrers(n ssa.Node, maxInstrReached map[*ssa.BasicBlock]int, lastBlockVisited *ssa.BasicBlock) {
+	if !hasTaintableType(n) {
+		return
+	}
+
```
PurelyApplied, line 227: What is the motivation of moving this block from `shouldNotTaint`?

internal/pkg/propagation/propagation.go
```diff
@@ -417,7 +417,7 @@ func hasTaintableType(n ssa.Node) bool {
 	if v, ok := n.(ssa.Value); ok {
 		switch t := v.Type().(type) {
 		case *types.Basic:
-			return t.Info() != types.IsBoolean
+			return (t.Info() & types.IsString) != 0
```
PurelyApplied, line 420: Not to gild a rose here, but should we check non-basic types?  While we won't taint, e.g., `float32`, we would still taint a `[]float32`, correct?

I don't remember where we left type inference, but that could be extended to include exemption for all non-interface-or-string-like-reachable structs, e.g. `type Coordinate struct { x, y int }` and `map[int]int` would be implicitly untaintable, being composed only of untaintable data.  That could be work for future PRs, though.  Just wondering your thoughts and if it's worth opening an `Issue` for it.

internal/pkg/propagation/propagation.go
PurelyApplied: I swear I clicked on `Conversation` and it said `0 comments`.  Let me go read the PR description...

PurelyApplied: Now that I'm caught up...

For clarity, we have an existing bug where a call that returns a boolean that should propagate taint between arguments would not have done so before this PR.  E.g. 
```golang
func foo(x, y interface[]) (ok bool) { 
  // ...
}

// ...
func bar(s Source) {
  var tainted interface{}
  if ok := foo(s, tainted) ; ok {
    sink(tainted) // TODO want
  }
}
```

If it's not too much of a hassle, I'd prefer to see a bug fix for this as a separate PR.  (Actually, I _really_ just want it visible as a separate commit in history, but since we squash rather than merge that necessitates a separate PR.)

internal/pkg/propagation/propagation.go
mlevesquedion: You are correct. I'm personally ok with waiting for e.g. `[]float32` to cause problems, like we did for e.g. `int` in the case of this PR.

I think what you bring up concerning type inference is interesting. We would need to be careful to not exclude e.g. `[]byte`, as that it basically the same thing as a `string`.

IMO this is definitely work for future PRs. I'm not sure it's worth its own issue, but it should at the very least be a comment on the existing source inference issue(s).

PurelyApplied: > IMO this is definitely work for future PRs. [...]

Sounds good to me.

And yeah, there's all the special handling around the general castability between `string`, `[]byte` (alias for `[]uint8`), and `[]rune` (alias for `[]int32`).  Because why should anything be clean cut and simple?

internal/pkg/propagation/propagation.go
mlevesquedion: Thank you for pointing out this test case.

The move from `shouldNotTaint` to `taintReferrers` originally came from an existing failing test case involving the `copy` builtin. So to implement this change without failing tests, I would need to make the move first (justifying it with e.g. the test case you gave), and then make the change to only taint strings.

I can definitely do that, but I'm not sure I see the benefit (a slightly cleaner git history?), so to me this does feel like a hassle. If you feel strongly about it though, I'll do it.

I'm not sure that this requires a separate PR, as I could just undo the commit, add the boolean test case, commit the boolean test case + move in a first commit, then do the int change in a second commit, and finally force push. If that sounds gross to you though, we can go with the 2 PRs route.

mlevesquedion:
See #289, as this PR will depend on it. Converting to draft until #289 is merged.

mlevesquedion:
Now that #289 is merged, this is once again ready for review.

internal/pkg/propagation/propagation.go
mlevesquedion: Done.

vinayakankugoyal (approved):

PurelyApplied (approved):

PR #289, opened 2021-03-09
Fix bug: do not skip instructions that return types that aren't taintable

mlevesquedion:
This fixes a bug discovered in #287. See that PR for details.

TLDR: When checking whether we should avoid a given `SSA` node based on its type (e.g., `bool` values can't be tainted), we should still do propagate to the operands if we normally would. We should avoid the node's referrers though, because that can lead to incorrect propagations. See the test case added by this PR for an example.


PurelyApplied (approved):
Thanks for extracting this.

PR #297, opened 2021-04-07
Allow config to be set directly.

mlevesquedion:
This change enables config bytes to be provided directly, instead of passing a file path on the command line. This is useful when running in an environment where file access is inconvenient.

Originally, I wanted to provide an API to set the config using a `Config` object. However, I think that this exposes too much of the implementation details of the `config` package. It is also a cumbersome change, requiring all types below `Config` to be made public and adding aliases for them in `pkg/levee`. I think it is safer and simpler to only provide an API to set the bytes. This is sufficient for the use case that prompted this change, and it allows the config to be provided in the usual format.

The tests for this API are isolated in their own file to avoid interfering with existing tests of the `config` package.

I am on the fence about whether this needs an explicit callout in the documentation. This feels like an advanced feature, which is unlikely to be relevant to most users. Let me know what you think.


mlevesquedion:
Checks are failing with a rather opaque error message ("This check failed" - that's the whole message.)

GitHub status is green.

Re-running the jobs did not help.

Leaving this as a draft for now, will try again tomorrow.

go.mod
```diff
@@ -5,5 +5,6 @@ go 1.14
 require (
 	github.com/google/go-cmp v0.5.2
 	golang.org/x/tools v0.0.0-20200416214402-fc959738d646
+	gopkg.in/yaml.v2 v2.2.8
```
mlevesquedion, line 8: This is used in the test for `SetConfigBytes` (specifically, it provides the `yaml.Marshal` function).

PurelyApplied:
> Checks are failing with a rather opaque error message ("This check failed" - that's the whole message.)

Next to `CI`, there's a `!` message balloon.  Clicking on that, I get `GitHub Actions has encountered an internal error when running your job.`  Github status reports green, so hopefully it'll sort itself out in the near future.

internal/pkg/config/set_config_bytes_test.go
```diff
@@ -0,0 +1,43 @@
...
+import (
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
+	"gopkg.in/yaml.v2"
+)
+
+func TestSetConfigBytes(t *testing.T) {
+	set := &Config{ReportMessage: "test"}
```
PurelyApplied, line 26: Nit: Consider setting the bytes directly, e.g. `b := []byte(`ReportMessage: "test"`)`

internal/pkg/config/config.go
```diff
@@ -33,12 +33,23 @@ var (
...
 )
 
 func init() {
 	FlagSet.StringVar(&configFile, "config", "config.yaml", "path to analysis configuration file")
 }
 
+// SetConfigBytes allows the contents of a configuration file
+// to be provided directly. This is useful when running the tool
+// in an environment where file access is inconvenient.
+func SetConfigBytes(b []byte) {
```
PurelyApplied, line 49: Nit: `config.SetConfigBytes` seems a bit against the "no repetitions" naming conventions of Go.  Consider just `config.SetBytes`?  Or expose `configBytes` directly in lieu of an explicit setter.

internal/pkg/config/config.go
```diff
@@ -295,12 +306,29 @@ func (fm funcMatcher) MatchFunction(path, receiver, name string) bool {
...
+	if configBytes != nil {
+		return readConfigBytes()
+	}
+
 	return cache.read(configFile)
 }
 
+func readConfigBytes() (*Config, error) {
+	if configFromBytes == nil {
+		configFromBytes = new(Config)
```
PurelyApplied, line 322: It's been a while since I poked at the particulars of the Go memory model, but this seems thread-unsafe.  `configFromBytes` should either be `nil` or the final value to avoid instability.  Use an intermediate value for marshalling.  I'm assuming reassignment to the same value is thread safe.

Alternatively, use locks and a `cache` pattern similar to the read-from-file.

PurelyApplied (changes requested):

go.mod
PurelyApplied: The tests appear to pass if you use `sigs.k8s.io/yaml` in the test rather than the `gopkg`.

mlevesquedion: You are correct. Not sure what happened here. Removed.

internal/pkg/config/set_config_bytes_test.go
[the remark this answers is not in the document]
mlevesquedion: Good idea. Done.

internal/pkg/config/config.go
mlevesquedion: Good point. I like `SetBytes`. Done.

internal/pkg/config/config.go
mlevesquedion: Good catch. I think in this case an intermediate value + a comment makes the most sense. Done.

PurelyApplied (approved):

PR #298, opened 2021-04-07
Allow summaries to be added externally.

mlevesquedion:
This change exposes the summaries maps so that additional summaries may be added if needed.

Like #297, this is an advanced feature and I am not sure if or how we should document it.

### DIFF
--- a/internal/pkg/propagation/summary/api.go
+++ b/internal/pkg/propagation/summary/api.go
@@ -30,10 +30,10 @@ import (
 // For returns the summary for a given call if it exists,
 // or nil if no summary matches the called function.
 func For(call *ssa.Call) *Summary {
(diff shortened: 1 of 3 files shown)


PurelyApplied (approved):

PurelyApplied (approved):
LGTM.  I see you edited the commit message to go from `set` to `added`.  I'm not sure if you intended that distinction to exist in the code, but direct exposure of the maps would allow for explicit setting, including removal of existing summaries.  If we wanted to only expose addition of new summaries, we'd want to expose some `Add(Interface)Summary` functions rather than the maps themselves.

I don't have a strong opinion one way or t'other.

mlevesquedion:
Good point. I'll edit the commit message when merging.

PR #300, opened 2021-04-09
Expose Summary data type.

mlevesquedion:
This is an oversight that should have been included in #298. It is currently impossible to add summaries to the summary package's exposed maps because the `Summary` type itself cannot be imported by external code.

### DIFF
--- a/pkg/levee/levee.go
+++ b/pkg/levee/levee.go
@@ -27,6 +27,10 @@ var Analyzer = levee.Analyzer
 // SetBytes is a wrapper around the config package's SetBytes function.
 var SetBytes = config.SetBytes
 
+// Summary is a wrapper around the propagation/summary
+// package's Summary type.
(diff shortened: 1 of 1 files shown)


PurelyApplied (approved):
This is feeling ad hoc.  If we're going to expose much more, we should probably distinguish separate `config/api.go` / `analyzer/api.go` / `whateverelse/api.go`.

My immediate thought looking more closely at the summaries is that a user wishing to define their own needs to intuit the correct function to key translation.  We should probably expose that also.  And with it being an exposed API, I'm going to reiterate that it chaffs a bit to have a mix-and-match of a bitmask and an integer slices.

Just some things to keep in mind going forward.

mlevesquedion:
Agreed on all counts. I think this is ok for now, but if we need to keep making changes here then we'll need to think more deeply about the structure.

PR #305, opened 2021-04-15
Propagate through go calls more selectively.

mlevesquedion:
This PR fixes some inconsistencies in how Call and Call-like instructions (Defer, Go) are handled.

The case that handles `Defer` says:
```
	// These nodes cannot propagate taint.
	case ..., *ssa.Defer, ...:
```

This is unfortunately inaccurate. For the most part, a `Defer` behaves like a `Call` (the main difference being *when* the call occurs.)

The case that handles `Go` says:
```
	// These nodes don't have referrers; they are Instructions, not Values.
	case *ssa.Go:
		prop.taintOperands(n, maxInstrReached, lastBlockVisited)
```

While it is true that a `Go` instruction can't have referrers, the way it propagates taint through its operands should be the same as for regular calls.

These inconsistencies can cause both false negatives and false positives. See the test cases for details.

I have moved the identification of sanitizers to the `Call` instruction's branch, because I don't think we want to consider `defer sanitize(...)` and `go sanitize(...)` to be valid sanitizations.

### DIFF
--- a/internal/pkg/levee/testdata/src/levee_analysistest/example/tests/callcommon/tests.go
+++ b/internal/pkg/levee/testdata/src/levee_analysistest/example/tests/callcommon/tests.go
@@ -1,0 +1,57 @@
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package callcommon contains tests for Defer and Go instructions,
+// which should be treated similarly to Call instructions.
+package callcommon
+
+import (
+	"fmt"
+	"io"
+	"levee_analysistest/example/core"
+)
+
+func TestGoBuiltin(dst, src []interface{}, s core.Source) {
+	src[0] = s
+	go copy(dst, src)
+	core.Sink(dst) // want "a source has reached a sink"
+}
+
+func TestDeferBuiltin(dst, src []interface{}, s core.Source) {
+	src[0] = s
+	defer copy(dst, src)
+	core.Sink(dst) // want "a source has reached a sink"
+}
+
+func TestGoStdlib(w io.Writer, s core.Source) {
+	go fmt.Fprint(w, s.Data)
+	core.Sink(w) // want "a source has reached a sink"
+}
+
+func TestDeferStdlib(w io.Writer, s core.Source) {
+	defer fmt.Fprint(w, s.Data)
+	core.Sink(w) // want "a source has reached a sink"
+}
+
+func TestGoUnknownFunction(i *core.Innocuous, s core.Source) {
+	go baz(i, s)
+	core.Sink(i)
+}
+
+func TestDeferUnknownFunction(i *core.Innocuous, s core.Source) {
+	defer baz(i, s)
+	core.Sink(i)
+}
+
+func baz(a, b interface{}) {}
--- a/internal/pkg/propagation/propagation.go
+++ b/internal/pkg/propagation/propagation.go
@@ -145,7 +145,7 @@ func (prop *Propagation) taintNeighbors(n ssa.Node, maxInstrReached map[*ssa.Bas
 
 	// The Go instruction is a wrapper around an implicit Call instruction.
 	case *ssa.Go:
-		prop.taintStdlibCall(t, t.Call, maxInstrReached, lastBlockVisited)
+		prop.taintStdlibCall(t, maxInstrReached, lastBlockVisited)
 
 	case *ssa.Field:
 		prop.taintField(n, maxInstrReached, lastBlockVisited, t.X.Type(), t.Field)
@@ -252,7 +252,7 @@ func (prop *Propagation) taintCall(call *ssa.Call, maxInstrReached map[*ssa.Basi
 		return
 	}
 
-	prop.taintStdlibCall(call, call.Call, maxInstrReached, lastBlockVisited)
+	prop.taintStdlibCall(call, maxInstrReached, lastBlockVisited)
 }
 
 func (prop *Propagation) taintBuiltin(call *ssa.Call, builtinName string, maxInstrReached map[*ssa.BasicBlock]int, lastBlockVisited *ssa.BasicBlock) {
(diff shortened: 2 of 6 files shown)


internal/pkg/levee/testdata/src/levee_analysistest/example/tests/callcommon/tests.go
```diff
@@ -0,0 +1,57 @@
...
+	src[0] = s
+	go copy(dst, src)
+	core.Sink(dst) // want "a source has reached a sink"
+}
+
+func TestDeferBuiltin(dst, src []interface{}, s core.Source) {
+	src[0] = s
+	defer copy(dst, src)
+	core.Sink(dst) // want "a source has reached a sink"
+}
```
PurelyApplied, line 35: This isn't expected to actually propagate.  Please add a ` // TODO (opened issue number)`.

internal/pkg/levee/testdata/src/levee_analysistest/example/tests/callcommon/tests.go
```diff
@@ -0,0 +1,57 @@
...
+	"fmt"
+	"io"
+	"levee_analysistest/example/core"
+)
+
+func TestGoBuiltin(dst, src []interface{}, s core.Source) {
+	src[0] = s
+	go copy(dst, src)
+	core.Sink(dst) // want "a source has reached a sink"
+}
```
PurelyApplied, line 29: `go` is indeterminate.  Please add a test swapping the first to lines of this test, as that would also potentially taint the sink.

internal/pkg/levee/testdata/src/levee_analysistest/example/tests/callcommon/tests.go
```diff
@@ -0,0 +1,57 @@
...
+
+func TestGoStdlib(w io.Writer, s core.Source) {
+	go fmt.Fprint(w, s.Data)
+	core.Sink(w) // want "a source has reached a sink"
+}
+
+func TestDeferStdlib(w io.Writer, s core.Source) {
+	defer fmt.Fprint(w, s.Data)
+	core.Sink(w) // want "a source has reached a sink"
+}
```
PurelyApplied, line 45: Ditto `// TODO (issue)`

PurelyApplied (changes requested):
> For the most part, a Defer behaves like a Call (the main difference being when the call occurs.)

That "when" is important, though.  A deferred copy doesn't execute until after the sink.  A goroutine copy may or may not propagate taint before a sink, irrespective of whether the argument was tainted when the goroutine was declared.

If we wanted to do this "right," we could properly post-process all `defer` and `go` calls after the propagation has spread through the containing blocks, in order to have a most-complete picture of the possible taint for when that instruction executes.  Although it's a little moot at present; since we are now explicitly enumerating which calls can propagate, all of these scenarios are a little pathological.  Why would someone `defer` a `copy`?  I don't see a real-world use of `go` or `defer` that directly invokes something from the standard library (other than closing a channel or canceling a context, neither of which propagate).

If the scope of this PR is to just flesh out our stance on `go` and `defer`, we could certainly flesh out testing to include some anonymous functions.

internal/pkg/propagation/propagation.go
```diff
@@ -240,34 +247,28 @@ func (prop *Propagation) taintOperands(n ssa.Node, maxInstrReached map[*ssa.Basi
 	}
 }
 
-func (prop *Propagation) taintCall(call *ssa.Call, maxInstrReached map[*ssa.BasicBlock]int, lastBlockVisited *ssa.BasicBlock) {
+func (prop *Propagation) taintCall(call ssa.Node, callCommon ssa.CallCommon, maxInstrReached map[*ssa.BasicBlock]int, lastBlockVisited *ssa.BasicBlock) {
```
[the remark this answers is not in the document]
mlevesquedion: (I don't have a hard preference here, just sharing my perception.) AFAIK we've been following the types of whatever `ssa` makes available to us? So e.g. all concrete nodes are pointers, but this CallCommon thing is a value, so I used it as is instead of passing a reference.

WDYT?

mlevesquedion:
You make some great points here, I think I took on too much in this PR. What originally prompted me to make a change was that I encountered a false positive similar to this:

```go
func TestGoUnknownFunction(i *core.Innocuous, s core.Source) {
	go baz(i, s)
	core.Sink(i) // a report is produced here, but to be coherent with how we handle regular calls, it shouldn't be
}
```

This is because currently, with `Go` instructions we propagate taint to all reference-like operands, which is incoherent with the way calls are handled since #292.

I agree that deferring/going a builtin or standard library function is likely to be very rare. I did a quick search in `kubernetes` and I didn't find much. There are a few uses of `go io.Copy(..., ...)`, and a few uses of `defer klog.Infof(...)`.

I think for now, I'll just change the propagation for the `Go` instruction to avoid the false positives and create an issue outlining future work.

internal/pkg/propagation/propagation.go
```diff
@@ -141,7 +141,18 @@ func (prop *Propagation) taintNeighbors(n ssa.Node, maxInstrReached map[*ssa.Bas
...
+			return
+		}
+		prop.taintCall(t, t.Call, maxInstrReached, lastBlockVisited)
+
+	// These nodes are wrappers around an implicit Call instruction.
+	// (The cases can't be merged because we need to access the .Call field.)
+	case *ssa.Defer:
+		prop.taintCall(t, t.Call, maxInstrReached, lastBlockVisited)
+	case *ssa.Go:
+		prop.taintCall(t, t.Call, maxInstrReached, lastBlockVisited)
```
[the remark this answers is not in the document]
mlevesquedion: I took a closer look at the interfaces in the `ssa` package and it turns out there's a `CallInstruction` interface with a `Common()` method. This interface is only implemented by `*ssa.{Call,Defer,Go}` and `Common()` returns the `CallCommon`. I've changed the code to use this instead, I think it's a significant improvement. PTAL.

internal/pkg/propagation/propagation.go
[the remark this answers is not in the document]
PurelyApplied: [outdated, resolving]

internal/pkg/propagation/propagation.go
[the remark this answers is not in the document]
PurelyApplied: `CallInstruction` cleans this all up very nicely.  👍

internal/pkg/levee/testdata/src/levee_analysistest/example/tests/callcommon/tests.go
PurelyApplied: [Outdated.  Resolving.]

internal/pkg/levee/testdata/src/levee_analysistest/example/tests/callcommon/tests.go
PurelyApplied: [Outdated.  Resolving.]

PurelyApplied (approved):
The thinner scope looks good to me.

PR #308, opened 2021-04-24
EAR pointer analysis: add inter-procedural support

guodongli-google:

### DIFF
--- a/internal/pkg/earpointer/analysis.go
+++ b/internal/pkg/earpointer/analysis.go
@@ -22,13 +22,12 @@ import (
 	"strconv"
 	"strings"
 
-	"golang.org/x/tools/go/types/typeutil"
-
 	"github.com/google/go-flow-levee/internal/pkg/config"
 
 	"golang.org/x/tools/go/analysis"
 	"golang.org/x/tools/go/analysis/passes/buildssa"
 	"golang.org/x/tools/go/ssa"
+	"golang.org/x/tools/go/types/typeutil"
 )
 
 // Analyzer traverses the packages and constructs an EAR partitions
@@ -434,8 +433,13 @@ func (vis *visitor) visitBuiltin(builtin *ssa.Builtin, instr ssa.Instruction) {
 // Collect unification constraints corresponding to a call.
 // This generates constraints for unifying parameters, free variables, and return values.
 func (vis *visitor) collectCalleeConstraints(common *ssa.CallCommon, fn *ssa.Function, instr ssa.Instruction) (map[ssa.Value][]ssa.Value, map[ssa.Value][][]ssa.Value) {
-	// Handle undefined/unknown functions.
-	if fn == nil || len(fn.Blocks) == 0 || len(fn.Params) != len(common.Args) {
+	// Handle undefined/unlinked functions.
+	if fn == nil || len(fn.Blocks) == 0 {
+		return nil, nil
+	}
+	// TODO: in some rare cases, SSA may generate an imported function with no arguments while this function actually
+	//  takes arguments. Skip such functions here.
+	if len(fn.Params) != len(common.Args) {
 		return nil, nil
 	}
 
@@ -443,8 +447,13 @@ func (vis *visitor) collectCalleeConstraints(common *ssa.CallCommon, fn *ssa.Fun
 	// Add caller_reg -> {callee_reg} constraints for parameters.
 	// For example, for g(a, b) and func g(x, y), add <a, g.x> and
 	// <b, g.y> into the constraints.
-	// Disable the unification of receivers here to avoid too many FPs.
-	// This may lead to under-approximation and FNs of the checkers.
+	//
+	// Unifying the receivers may cause too many FPs.
+	// For example, func (t T) f() { ... }; o1.f() and o2.f(), the unification
+	// results in {o1, o2, t}, while o1 and o2 should not be unified.
+	// So the receiver unification is disabled here, which may lead to under-approximation
+	// and FNs of the checkers.
+	// TODO: support context-sensitivity to mitigate this issue.
 	start := 0
 	if fn.Signature.Recv() != nil {
 		start = 1
--- a/internal/pkg/earpointer/analysis.go
+++ b/internal/pkg/earpointer/analysis.go
@@ -52,8 +52,11 @@ type visitor struct {
 func run(pass *analysis.Pass) (interface{}, error) {
 	ssainput := pass.ResultOf[buildssa.Analyzer].(*buildssa.SSA)
 	vis := visitor{state: NewState(), contexts: []*Context{&emptyContext}}
-	vis.initReferences(ssainput)
-	for _, fn := range ssainput.SrcFuncs {
+	// Analyze all the functions and methods in the package,
+	// not just those in ssainput.ScrFuncs.
+	fns := make(map[*ssa.Function]bool) // storing all the functions and methods
+	vis.initReferences(ssainput, fns)
+	for fn := range fns {
 		for _, blk := range fn.Blocks {
 			for _, i := range blk.Instrs {
 				vis.visitInstruction(i)
@@ -64,7 +67,8 @@ func run(pass *analysis.Pass) (interface{}, error) {
 }
 
 // Insert into the state all the local references and global references.
-func (vis *visitor) initReferences(ssainput *buildssa.SSA) {
+// Argument "fns" stores the functions and methods found during this process.
+func (vis *visitor) initReferences(ssainput *buildssa.SSA, fns map[*ssa.Function]bool) {
 	state := vis.state
 	prog := ssainput.Pkg.Prog
 	for _, member := range ssainput.Pkg.Members {
@@ -75,11 +79,11 @@ func (vis *visitor) initReferences(ssainput *buildssa.SSA) {
 				state.Insert(MakeGlobal(m))
 			}
 		case *ssa.Function:
-			vis.initFunction(m)
+			vis.initFunction(m, fns)
 		case *ssa.Type:
 			for _, meth := range typeutil.IntuitiveMethodSet(m.Type(), &prog.MethodSets) {
 				if mv := prog.MethodValue(meth); mv != nil {
-					vis.initFunction(mv)
+					vis.initFunction(mv, fns)
 				}
 			}
 		case *ssa.NamedConst:
@@ -88,7 +92,8 @@ func (vis *visitor) initReferences(ssainput *buildssa.SSA) {
 	}
 }
 
-func (vis *visitor) initFunction(fn *ssa.Function) {
+func (vis *visitor) initFunction(fn *ssa.Function, fns map[*ssa.Function]bool) {
+	fns[fn] = true
 	state := vis.state
 	for _, param := range fn.Params {
 		if typeMayShareObject(param.Type()) {
@@ -124,7 +129,7 @@ func (vis *visitor) initFunction(fn *ssa.Function) {
 	}
 	// Recursively process the embedded functions
 	for _, anon := range fn.AnonFuncs {
-		vis.initFunction(anon)
+		vis.initFunction(anon, fns)
 	}
 }
 
@@ -358,8 +363,11 @@ func (vis *visitor) visitCall(call *ssa.CallCommon, instr ssa.Instruction) {
 	// Collect unification constraints
 	// Here only static calls are supported.
 	// A future extension is to use a more advanced call graph.
-	paramCstrs, retCstrs :=
-		vis.collectCalleeConstraints(call, call.StaticCallee(), instr)
+	fn := call.StaticCallee()
+	if fn == nil {
+		return
+	}
+	paramCstrs, retCstrs := vis.collectCalleeConstraints(call, fn, instr)
 	// Unify caller arguments and callee parameters with suitable context.
 	for arg, params := range paramCstrs {
 		for _, param := range params {
--- a/internal/pkg/earpointer/analysis.go
+++ b/internal/pkg/earpointer/analysis.go
@@ -22,12 +22,13 @@ import (
 	"strconv"
 	"strings"
 
+	"golang.org/x/tools/go/ssa/ssautil"
+
 	"github.com/google/go-flow-levee/internal/pkg/config"
 
 	"golang.org/x/tools/go/analysis"
 	"golang.org/x/tools/go/analysis/passes/buildssa"
 	"golang.org/x/tools/go/ssa"
-	"golang.org/x/tools/go/types/typeutil"
 )
 
 // Analyzer traverses the packages and constructs an EAR partitions
@@ -52,10 +53,13 @@ type visitor struct {
 func run(pass *analysis.Pass) (interface{}, error) {
 	ssainput := pass.ResultOf[buildssa.Analyzer].(*buildssa.SSA)
 	vis := visitor{state: NewState(), contexts: []*Context{&emptyContext}}
+	vis.initGlobalReferences(ssainput)
 	// Analyze all the functions and methods in the package,
 	// not just those in ssainput.ScrFuncs.
-	fns := make(map[*ssa.Function]bool) // storing all the functions and methods
-	vis.initReferences(ssainput, fns)
+	fns := ssautil.AllFunctions(ssainput.Pkg.Prog)
+	for fn := range fns {
+		vis.initFunction(fn)
+	}
 	for fn := range fns {
 		for _, blk := range fn.Blocks {
 			for _, i := range blk.Instrs {
@@ -66,34 +70,23 @@ func run(pass *analysis.Pass) (interface{}, error) {
 	return vis.state.ToPartitions(), nil
 }
 
-// Insert into the state all the local references and global references.
-// Argument "fns" stores the functions and methods found during this process.
-func (vis *visitor) initReferences(ssainput *buildssa.SSA, fns map[*ssa.Function]bool) {
+// Insert into the state all the global references.
+func (vis *visitor) initGlobalReferences(ssainput *buildssa.SSA) {
 	state := vis.state
-	prog := ssainput.Pkg.Prog
 	for _, member := range ssainput.Pkg.Members {
-		switch m := member.(type) {
-		case *ssa.Global:
-			if typeMayShareObject(m.Type()) &&
-				!strings.HasPrefix(m.Name(), "init$") { // skip some synthetic variables
-				state.Insert(MakeGlobal(m))
+		if g, ok := member.(*ssa.Global); ok {
+			if typeMayShareObject(g.Type()) &&
+				// skip some synthetic variables
+				!strings.HasPrefix(g.Name(), "init$") {
+				state.Insert(MakeGlobal(g))
 			}
-		case *ssa.Function:
-			vis.initFunction(m, fns)
-		case *ssa.Type:
-			for _, meth := range typeutil.IntuitiveMethodSet(m.Type(), &prog.MethodSets) {
-				if mv := prog.MethodValue(meth); mv != nil {
-					vis.initFunction(mv, fns)
-				}
-			}
-		case *ssa.NamedConst:
-			// Global constants are not within the scope of the pointer analysis.
 		}
+		// Global constants are not within the scope of the pointer analysis.
 	}
 }
 
-func (vis *visitor) initFunction(fn *ssa.Function, fns map[*ssa.Function]bool) {
-	fns[fn] = true
+// Insert into the state all the local references.
+func (vis *visitor) initFunction(fn *ssa.Function) {
 	state := vis.state
 	for _, param := range fn.Params {
 		if typeMayShareObject(param.Type()) {
@@ -129,7 +122,7 @@ func (vis *visitor) initFunction(fn *ssa.Function, fns map[*ssa.Function]bool) {
 	}
 	// Recursively process the embedded functions
 	for _, anon := range fn.AnonFuncs {
-		vis.initFunction(anon, fns)
+		vis.initFunction(anon)
 	}
 }
 
--- a/internal/pkg/earpointer/analysis.go
+++ b/internal/pkg/earpointer/analysis.go
@@ -22,6 +22,8 @@ import (
 	"strconv"
 	"strings"
 
+	"golang.org/x/tools/go/types/typeutil"
+
 	"github.com/google/go-flow-levee/internal/pkg/config"
 
 	"golang.org/x/tools/go/analysis"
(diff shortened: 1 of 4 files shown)


internal/pkg/earpointer/analysis.go
```diff
@@ -65,33 +67,61 @@ func run(pass *analysis.Pass) (interface{}, error) {
...
 				!strings.HasPrefix(m.Name(), "init$") { // skip some synthetic variables
 				state.Insert(MakeGlobal(m))
 			}
+		case *ssa.Function:
+			vis.initFunction(m)
+		case *ssa.Type:
+			for _, meth := range typeutil.IntuitiveMethodSet(m.Type(), &prog.MethodSets) {
+				if mv := prog.MethodValue(meth); mv != nil {
+					vis.initFunction(mv)
+				}
```
PurelyApplied, line 84: We're a bit out of depth here, but is there any concern about double-initializing something that embeds?  I'm not sure how that method set resolves in, e.g.,

```golang
type Bar struct {}
func (Bar) Method() {}

type Foo struct {
  Bar
}
```
Will both `Foo.Method()` and `Bar.Method()` be produced by `IntuitiveMethodSet` / produce the same `*ssa.Function` from `prog.MethodValue`?  Is it a problem in `initFunction` if they do?  I see that you have handling in `initFunction` explicitly for anonymous fields, so I'm guessing you've already considered this.

Would be a good test-case to include, in any event.

internal/pkg/earpointer/analysis.go
```diff
@@ -65,33 +67,61 @@ func run(pass *analysis.Pass) (interface{}, error) {
 // Insert into the state all the local references and global references.
 func (vis *visitor) initReferences(ssainput *buildssa.SSA) {
 	state := vis.state
+	prog := ssainput.Pkg.Prog
 	for _, member := range ssainput.Pkg.Members {
-		if m, ok := member.(*ssa.Global); ok {
+		switch m := member.(type) {
```
PurelyApplied, line 72: We're now covering every `Member` type except `*ssa.NamedConst`.  It could be good to make it clear that wasn't an oversight, either with an explicit "do nothing" case or just a comment here at the switch.

Not necessary if it's coming in a near-future PR.

internal/pkg/earpointer/analysis.go
```diff
@@ -65,33 +67,61 @@ func run(pass *analysis.Pass) (interface{}, error) {
...
+	}
+	for _, fv := range fn.FreeVars {
+		if typeMayShareObject(fv.Type()) {
+			state.Insert(MakeLocal(&emptyContext, fv))
 		}
-		for _, fv := range fn.FreeVars {
-			if typeMayShareObject(fv.Type()) {
-				state.Insert(MakeLocal(&emptyContext, fv))
+	}
+	// Usual tuple-type registers are skipped since they will be processed by subsequent Extract instructions,
```
PurelyApplied, line 102: Nit: I try to respect 80-width, since I often have my display split vertically.

internal/pkg/earpointer/analysis.go
```diff
@@ -315,7 +345,134 @@ func (vis *visitor) visitSelect(s *ssa.Select) {
...
+				}
+			}
+		}
+	}
+}
+
+// Process calls to builtin functions; return true if the callee is a builtin
+// function.
+// Builtin functions: https://golang.org/pkg/builtin/
+func (vis *visitor) processBuiltinCall(builtin *ssa.Builtin, instr ssa.Instruction) {
```
PurelyApplied, line 399: You seem to have a well-defined sense of `visit` vs `process`, but that distinction is a bit opaque to me.  I might've guessed this should be named `visitBuiltinCall`, since it seems to be at that higher visitor level.

If there is a consistent theme in the verbiage for these methods, I'd find a top-level summary comment near the `package` declaration useful.

internal/pkg/earpointer/analysis_test.go
```diff
@@ -625,13 +625,6 @@ func TestMakeInterface(t *testing.T) {
 func TestUnimplemented(t *testing.T) {
 	code := `package p
 	func f() {
-		defer func() {
-			print("defer")
-		}()
-		go func() {
-			print("go")
-		}()
-
```
PurelyApplied, line 634: Is this `Test*` just a scratch pad for the things you are still intending to cover?  This block of deletion caught me off guard, and I note tests still pass with or without these lines.

internal/pkg/earpointer/analysis.go
```diff
@@ -315,7 +345,134 @@ func (vis *visitor) visitSelect(s *ssa.Select) {
...
+	if builtin.Name() == "append" { // func append(slice []Type, elems ...Type) []Type
```
mlevesquedion, line 400: Nit: this would read better as a `switch`.

internal/pkg/earpointer/analysis_test.go
```diff
@@ -651,3 +644,247 @@ func TestUnimplemented(t *testing.T) {
...
+			t5 = print(d1, t3)                         ()
+			return
+	*/
+	state, err := runCode(code)
+	if err != nil {
+		t.Fatal(err)
+	}
+	// f.s and f.d1 are separate due to the "copy";
+	// f.s, f.t2 and f.r3 ("d2") are unified due to the "append".
+	want := "{f.d1[.],f.t1}: --> f.x, {f.d1}: [0->f.t1], {f.s,f.t0,f.t2,f.t3}: [0->f.t1], {f.x}: []"
```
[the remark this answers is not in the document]
guodongli-google: This is due to unify-by-value: here f.s is cloned to f.d1 using "copy(d1, s)", which indicates that all the elements of s are copied to d1, but s and d1 are different slices. This is different from "d1 = s". In an approximation, we may unify d1 and s but that would not be the original semantics of "copy()".

internal/pkg/earpointer/analysis.go
```diff
@@ -315,7 +345,134 @@ func (vis *visitor) visitSelect(s *ssa.Select) {
...
+func (vis *visitor) collectCalleeConstraints(common *ssa.CallCommon, fn *ssa.Function, instr ssa.Instruction) (map[ssa.Value][]ssa.Value, map[ssa.Value][][]ssa.Value) {
+	if fn == nil || len(fn.Blocks) == 0 { // undefined functions
+		return nil, nil
+	}
+
+	paramCstrs := make(map[ssa.Value][]ssa.Value)
+	// Add caller_reg -> {callee_reg} constraints for parameters which are references.
+	// For example, for g(a, b) and func g(x, y), <a, g.x> and <b, g.y> into the constraints.
+	l1, l2 := len(common.Args), len(fn.Params)
+	if l2 < l1 {
```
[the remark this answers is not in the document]
guodongli-google: Good point! This is a way to detect "anomaly" in SSA. For some reason SSA doesn't get the signature right for some cross-package functions. In general, most imported functions have empty bodies, but they still have the right signatures. So being cross-package is not the right criterion. 
Anyway, len(fn.Params) != len(common.Args) indicates something wrong.  Now the function detects this earlier.

internal/pkg/earpointer/analysis.go
```diff
@@ -315,7 +345,134 @@ func (vis *visitor) visitSelect(s *ssa.Select) {
...
+				vis.unifyByValue(MakeLocal(context, dst), MakeLocal(context, src))
+			}
+		}
+	}
+}
+
+// Collect unification constraints corresponding to a call.
+// This generates constraints for unifying parameters, free variables, and return values.
+func (vis *visitor) collectCalleeConstraints(common *ssa.CallCommon, fn *ssa.Function, instr ssa.Instruction) (map[ssa.Value][]ssa.Value, map[ssa.Value][][]ssa.Value) {
+	if fn == nil || len(fn.Blocks) == 0 { // undefined functions
```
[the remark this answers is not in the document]
guodongli-google: This will require a more advanced call graph, e.g. the VTA that DeepGo is building. Such support is future work.

internal/pkg/earpointer/analysis.go
```diff
@@ -315,7 +345,134 @@ func (vis *visitor) visitSelect(s *ssa.Select) {
...
+	}
+
+	paramCstrs := make(map[ssa.Value][]ssa.Value)
+	// Add caller_reg -> {callee_reg} constraints for parameters which are references.
+	// For example, for g(a, b) and func g(x, y), <a, g.x> and <b, g.y> into the constraints.
+	l1, l2 := len(common.Args), len(fn.Params)
+	if l2 < l1 {
+		l1 = l2
+	}
+	// Disable the unification of receivers here to avoid too many FPs. This may lead to
```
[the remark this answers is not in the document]
guodongli-google: Based on the results of a DeepGo checker that uses a similar pointe analysis, unifying the receivers with #context K=0 will increase the FP rate a lot (>30%). A better handling is to support K>0. What is in my next step is to make K=1 as default, and this TODO should be fixed then.

internal/pkg/earpointer/analysis.go
```diff
@@ -315,7 +345,134 @@ func (vis *visitor) visitSelect(s *ssa.Select) {
...
+// function.
+// Builtin functions: https://golang.org/pkg/builtin/
+func (vis *visitor) processBuiltinCall(builtin *ssa.Builtin, instr ssa.Instruction) {
+	if builtin.Name() == "append" { // func append(slice []Type, elems ...Type) []Type
+		// Propagage the arguments to the return value.
+		if dst, ok := instr.(ssa.Value); ok {
+			ops := instr.Operands(nil)
+			for i := 1; i < len(ops); i++ {
+				arg := *ops[i]
+				if mayShareObject(arg) {
```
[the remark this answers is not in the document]
guodongli-google: The context is considered within UnifyLocals. Revised the code accordingly.

internal/pkg/earpointer/analysis.go
```diff
@@ -315,7 +345,134 @@ func (vis *visitor) visitSelect(s *ssa.Select) {
...
+
+// Process calls to builtin functions; return true if the callee is a builtin
+// function.
+// Builtin functions: https://golang.org/pkg/builtin/
+func (vis *visitor) processBuiltinCall(builtin *ssa.Builtin, instr ssa.Instruction) {
+	if builtin.Name() == "append" { // func append(slice []Type, elems ...Type) []Type
+		// Propagage the arguments to the return value.
+		if dst, ok := instr.(ssa.Value); ok {
+			ops := instr.Operands(nil)
+			for i := 1; i < len(ops); i++ {
```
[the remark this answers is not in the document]
guodongli-google: We need to skip ops[0] which is the function itself.

internal/pkg/earpointer/analysis_test.go
guodongli-google: They are now covered by "TestGoDefer(t *testing.T)". 
Previously I added them purely to obtain better code coverage.

internal/pkg/earpointer/analysis.go
guodongli-google: Good point! I added a test.
In general this is not a real issue because (1) SSA will produce Method() for Foo and Bar individually; so Method() won't be reused. (2) "state.Insert(ref)" will insert "ref" into the state only when ref is not already in the state. 
A more serious issue is that when a reference is not in a state while it is used. So here we want to instantiate all the local/global variables.

guodongli-google:
Somehow I can't add new reviewer into this PR. Can we add Tim as a reviewer?

mlevesquedion:
It seems I am unable to add him either. :thinking: If you send him a link to the PR he should be able to review it though.

PurelyApplied:
> Somehow I can't add new reviewer into this PR. Can we add Tim as a reviewer?

I believe GitHub doesn't allow you to add someone to a PR until they are a contributor in that repo's history - usually as a contributor of at least one commit.  Mentions should still show up in their notifications, though (maybe?).  @timothy-king

internal/pkg/earpointer/analysis.go
```diff
@@ -315,7 +350,139 @@ func (vis *visitor) visitSelect(s *ssa.Select) {
...
+			}
+		}
+	}
+}
+
+// Collect unification constraints corresponding to a call.
+// This generates constraints for unifying parameters, free variables, and return values.
+func (vis *visitor) collectCalleeConstraints(common *ssa.CallCommon, fn *ssa.Function, instr ssa.Instruction) (map[ssa.Value][]ssa.Value, map[ssa.Value][][]ssa.Value) {
+	// Handle undefined/unknown functions.
+	if fn == nil || len(fn.Blocks) == 0 || len(fn.Params) != len(common.Args) {
```
timothy-king, line 438: When can `len(fn.Params) != len(common.Args)`?

internal/pkg/earpointer/analysis.go
```diff
@@ -315,7 +350,139 @@ func (vis *visitor) visitSelect(s *ssa.Select) {
...
+		return nil, nil
+	}
+
+	paramCstrs := make(map[ssa.Value][]ssa.Value)
+	// Add caller_reg -> {callee_reg} constraints for parameters.
+	// For example, for g(a, b) and func g(x, y), add <a, g.x> and
+	// <b, g.y> into the constraints.
+	// Disable the unification of receivers here to avoid too many FPs.
```
timothy-king, line 446: Why would this lead to too many FPs?

internal/pkg/earpointer/analysis_test.go
```diff
@@ -651,3 +644,303 @@ func TestUnimplemented(t *testing.T) {
...
+		0:                                                     entry P:0 S:0
+			t0 = &ks[0:int]                                    **int
+			t1 = *t0                                           *int
+			return t1
+	*/
+	state, err := runCode(code)
+	if err != nil {
+		t.Fatal(err)
+	}
+	want := "{f.a,f.t4,g.t1}: [], {f.b}: [], {f.t0,f.t3,g.ks}: " +
```
timothy-king, line 894: I am pleasantly surprised we have enough precision to split f.a and f.b. What enabled this?

internal/pkg/earpointer/analysis_test.go
```diff
@@ -651,3 +644,303 @@ func TestUnimplemented(t *testing.T) {
...
+		"[0->g.t0, 1->f.t2], {f.t1,g.t0}: --> g.t1, {f.t2}: --> f.b"
```
timothy-king, line 895: Is it possible to make the want statements less prone to line breaks between the keys and values?

internal/pkg/earpointer/analysis.go
```diff
@@ -315,7 +350,139 @@ func (vis *visitor) visitSelect(s *ssa.Select) {
...
+
+// Handle calls to builtin functions: https://golang.org/pkg/builtin/.
+func (vis *visitor) visitBuiltin(builtin *ssa.Builtin, instr ssa.Instruction) {
+	switch builtin.Name() {
+	case "append": // func append(slice []Type, elems ...Type) []Type
+		// Propagage the arguments to the return value.
+		if dst, ok := instr.(ssa.Value); ok {
+			ops := instr.Operands(nil)
+			// ops[0] is the "append" function itself.
+			for i := 1; i < len(ops); i++ {
```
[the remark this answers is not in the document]
guodongli-google: SSA does useful work here to avoid unify(l1, x), unify(l1, y), ...
For example, l1 = append(l0, x) is translated to:
  
  t1 = &t0[0:int]                            **int   // t0 is l0
  *t1 = x
  t2 = slice t0[:]                           []*int
  t3 = append(s, t2...)                []*int.  // t3 is l1

So a slice is first created and "x" is assigned to its first element, then this slice is unified with l1. Specifically, for your example, the heap is:
  {l1, l0} [0 -> x, 1 -> y]

internal/pkg/earpointer/analysis.go
guodongli-google: See Micheal's comment -- he did an experiment on k8s. Roughly, it is about some imported functions (perhaps synthesized functions?). This shall be considered as an SSA bug (which rarely shows up), and we are defensive here.

internal/pkg/earpointer/analysis.go
guodongli-google: Add comments. This will be addressed in the follow-up PR(s).

internal/pkg/earpointer/analysis_test.go
guodongli-google: Again SSA does some useful work here by first creating an array and storing arguments "a" and "b" into the array. Now f.a and f.b are the elements of this array (which is unified with the function argument).

  t0 = new [2]*int (varargs)                         *[2]*int
  t1 = &t0[0:int]                                    **int
  *t1 = a
  t2 = &t0[1:int]                                    **int
  *t2 = b

internal/pkg/earpointer/analysis.go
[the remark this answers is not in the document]
guodongli-google: See below. It is something else, possibly due to a bug in SSA.
Finding the root cause of this issue is out of this PR's scope though.

internal/pkg/earpointer/analysis_test.go
timothy-king: How are
```
  t1 = &t0[0:int] **int
  t2 = &t0[1:int] **int
```
separate? Are we keeping different 0 and 1 indices? If so where?

internal/pkg/earpointer/analysis_test.go
```diff
@@ -891,8 +891,9 @@ func TestVariadicCall(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
-	want := "{f.a,f.t4,g.t1}: [], {f.b}: [], {f.t0,f.t3,g.ks}: " +
-		"[0->g.t0, 1->f.t2], {f.t1,g.t0}: --> g.t1, {f.t2}: --> f.b"
+	want := "{f.a,f.t4,g.t1}: [], {f.b}: [], " +
```
timothy-king, line 894: This was not really the suggestion. Can we make a testing utility that takes a non-string type to do the comparison?  Like a map[string]string where the key is the "{f.a,f.t4,g.t1}" and value is the "[]"? That seems like it would simplify maintenance a lot.

mlevesquedion: `cmp.Diff` maybe?

guodongli-google: Now the non-trivial strings are replaced by a map from the key to the field string. Yes it is much more readable.

internal/pkg/earpointer/analysis_test.go
```diff
@@ -65,9 +77,15 @@ func TestFieldAddr(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
-	want := "{*f.a}: [x->f.t0], {f.a}: --> *f.a, {f.b}: [], {f.t0}: --> f.b"
-	if got := state.String(); got != want {
-		t.Errorf("got: %s\n want: %s", got, want)
+	want := map[string]string{
```
timothy-king, line 80: `want := partitionItemsToString(map[string]string{...})`

This gets rid of the pesky "wantstr" name.

internal/pkg/earpointer/analysis_test.go
```diff
@@ -47,6 +49,16 @@ func runCode(code string) (*earpointer.Partitions, error) {
 	return partitions.(*earpointer.Partitions), nil
 }
 
+// Concatenates the "members -> fields" map items into a string.
+func partitionItemsToString(m map[string]string) string {
```
timothy-king, line 53: "partitionItemsToString" the "ToString" suffix is not needed to disambiguate this given the signature.

(My hunch is that a short better name. I dont have a good suggestion though.)

internal/pkg/earpointer/analysis_test.go
```diff
@@ -120,9 +144,17 @@ func TestField(t *testing.T) {
 		t.Fatal(err)
 	}
 	// t2's field x points to t3
-	want := "{*f.t0}: [x->f.t1], {f.a}: [], {f.t0}: --> *f.t0, {f.t1}: --> f.a, {f.t2[.],f.t3}: --> f.a, {f.t2}: [x->f.t3]"
-	if got := state.String(); got != want {
-		t.Errorf("got: %s\n want: %s", got, want)
+	want := map[string]string{
+		"{f.t0}":         "--> *f.t0",
```
timothy-king, line 148: I think this better, but what do you think?

internal/pkg/earpointer/analysis_test.go
```diff
@@ -348,10 +439,20 @@ func TestStore(t *testing.T) {
...
+		"{*f.a}": "[x->f.t0, y->f.t2]",
+		"{f.b}":  "[10->f.t3]",
+		"{f.t0}": "--> f.t1",
+		"{f.t1}": "[]",
+		"{f.t2}": "--> f.t4",
+		"{f.t3}": "--> f.t4",
+		"{f.t4}": "[]",
+		"{z}":    "--> f.t1",
+	}
+	wantstr := partitionItemsToString(want)
```
timothy-king, line 453: All of these tests roughly have the same pattern. Seems like a good candidate for a table based test. Zvonimir had a decent setup you might want to take a look at. Used multiple files though.

internal/pkg/earpointer/heap.go
```diff
@@ -55,8 +55,14 @@ func (l Local) Value() ssa.Value {
 
 func (l Local) String() string {
 	reg := l.reg
-	if p := reg.Parent(); p != nil {
-		return l.context.String() + p.Name() + "." + reg.Name()
+	if f := reg.Parent(); f != nil {
+		fstr := f.Name()
+		if recv := f.Signature.Recv(); recv != nil {
+			if named, ok := recv.Type().(*types.Named); ok {
+				fstr = named.Obj().Name() + ":" + fstr
```
timothy-king, line 62: nit: return a string here to remove typestate and variable.

Also maybe fmt.Sprintf()? Not confident about that vs. `+` though.

timothy-king (changes requested):

internal/pkg/earpointer/analysis.go
```diff
@@ -315,7 +349,149 @@ func (vis *visitor) visitSelect(s *ssa.Select) {
...
+		// special handling of builtin functions
+		vis.visitBuiltin(builtin, instr)
+		return
+	}
+
+	// Collect unification constraints
+	// Here only static calls are supported.
+	// A future extension is to use a more advanced call graph.
+	paramCstrs, retCstrs :=
+		vis.collectCalleeConstraints(call, call.StaticCallee(), instr)
```
mlevesquedion, line 362: Consider returning early if `call.StaticCallee()` is `nil`. Without looking at the implementation of `collectCalleeConstraints`, it is not obvious that `paramCstrs` and `retCstrs` will be empty.

internal/pkg/earpointer/analysis_test.go
```diff
@@ -65,9 +77,15 @@ func TestFieldAddr(t *testing.T) {
...
-	if got := state.String(); got != want {
-		t.Errorf("got: %s\n want: %s", got, want)
+	want := map[string]string{
+		"{f.a}":  "--> *f.a",
+		"{*f.a}": "[x->f.t0]",
+		"{f.t0}": "--> f.b",
+		"{f.b}":  "[]",
+	}
+	wantstr := partitionItemsToString(want)
+	if got := state.String(); got != wantstr {
```
mlevesquedion, line 87: Consider using `cmp.Diff` on the strings for a human-readable diff. (We have a few example uses in this codebase if you want a quick reference.)

internal/pkg/earpointer/analysis_test.go
```diff
@@ -651,3 +767,347 @@ func TestUnimplemented(t *testing.T) {
...
+	if err != nil {
+		t.Fatal(err)
+	}
+	// f.t1, f.x and g.x are unified due to calling g().
+	// Note g.a is a struct receiver that won't be unified.
+	want := map[string]string{
+		"{f.a}":            "--> f.t0",
+		"{f.t0}":           "[]",
+		"{A:g.x,f.t1,f.x}": "[]",
+		"{A:g.a}":          "[]",
```
mlevesquedion, line 922: It feels like this should mirror `f.a --> f.t0` ?

internal/pkg/earpointer/analysis_test.go
```diff
@@ -651,3 +767,347 @@ func TestUnimplemented(t *testing.T) {
...
+		"{T1:f.arg0}":  "[]",
+		"{T1:f.x,g.x}": "[]",
+		"{T2:f.arg0}":  "[]",
+		"{T2:f.t0}":    "[]",
+		"{T2:f.t1}":    "[]",
+		"{T2:f.t2}":    "[]",
+		"{T2:f.x}":     "[]",
+		"{g.t0}":       "--> g.t1",
+		"{g.t1}":       "[]",
+		"{g.t2}":       "--> *g.t2",
```
mlevesquedion, line 1106: The name `t2` is a bit confusing here, because it's the name of both a param and a local temporary.

mlevesquedion (approved):
LGTM.

Maybe consider adding a test case with a recursive function.

internal/pkg/earpointer/analysis_test.go
timothy-king: " is that a short better name" -> "is that there exists a shorter better name"

internal/pkg/earpointer/analysis_test.go
guodongli-google: Yes writing the expected string into a text file or with some format will be surely more readable. But it requires a reader to open the file to compare. I think that a better one may be write the state as a Diagonistics and then everything is in the same source file. However vet's Diagonistics is basically one line so it may not be very readable if the state is large.
I am OK with the map-based solution for the moment. More cleaning up can be done in the future.

guodongli-google:
> LGTM.
> 
> Maybe consider adding a test case with a recursive function.

Such a test has been added. 
More tests about function/method calls will come in the next PR.

internal/pkg/earpointer/analysis.go
```diff
@@ -64,7 +67,8 @@ func run(pass *analysis.Pass) (interface{}, error) {
 }
 
 // Insert into the state all the local references and global references.
-func (vis *visitor) initReferences(ssainput *buildssa.SSA) {
+// Argument "fns" stores the functions and methods found during this process.
```
timothy-king, line 70: Why carry fns around as an output parameter? This is just all methods contained in ssainput.Pkg. I would do this after  of time in a run or generate them ahead of time. This keeps the functions simpler. If you need to accumulate this, you can add a field to visitor.

timothy-king (changes requested):

timothy-king (approved):
LGTM

Issue #312, opened 2021-06-12
Handle standard library functions in the analysis engine


guodongli-google:
## Bug report

**Describe the bug**
Go has a large number of library functions in the [standard package](https://golang.org/pkg/), whose source code may be missing and the propagation analysis may have no clue on their semantics. 
While we can hard-code the support for some known ones, e.g. `fmt.Printf(...)`, it may be impractical to do that for all the functions. If we over-approximate their semantics, then false positives will be generated.
This bug is to track the progress of the support for these library functions.

mlevesquedion:
I see you have already added a hook `visitKnownFunction` for this purpose.

FWIW, the non-EAR version of the analysis already supports this to a large extent. See the `internal/pkg/propagation/summary` package.

Perhaps these summaries can be used by EAR as well.

PR #313, opened 2021-06-12
Various minor improvements of the EAR pointer analysis.

guodongli-google:
(1) Use unify-by-reference more regressively.
(2) Handle the map operations better.
(3) Support some known functions.

### DIFF
--- a/internal/pkg/earpointer/analysis.go
+++ b/internal/pkg/earpointer/analysis.go
@@ -440,18 +440,16 @@ func (vis *visitor) visitExtract(extract *ssa.Extract) {
 	case *ssa.Select:
 		// t1 = select blocking [c1<-t0, <-c2]  // (index int, recvOk bool, r_0 T_0, ... r_n-1 T_n-1)
 		// t2 = extract t1 #n
-		if extract.Index > 1 {
-			k := -1 // for locating the (extract.Index-2)^{th} receive
-			sts := base.States
-			for i := 0; i < len(sts); i++ {
-				if sts[i].Dir == types.RecvOnly {
-					if k++; k == extract.Index-2 {
-						// Propagate the (extract.Index-2)^{th} receive value to the LHS.
-						if val := sts[i].Send; val != nil {
-							vis.unifyLocals(extract, val)
-						}
-						break
-					}
+		if extract.Index <= 1 {
+			break
+		}
+		k := -1 // for locating the (extract.Index-2)^{th} receive
+		for _, st := range base.States {
+			if st.Dir == types.RecvOnly {
+				if k++; k == extract.Index-2 && mayShareObject(st.Chan) {
+					// Extract the receive value from the channel.
+					vis.processHeapAccess(extract, st.Chan, anyIndexField)
+					break
 				}
 			}
 		}
--- a/internal/pkg/earpointer/analysis.go
+++ b/internal/pkg/earpointer/analysis.go
@@ -64,22 +64,25 @@ type visitor struct {
 
 func run(pass *analysis.Pass) (interface{}, error) {
 	ssainput := pass.ResultOf[buildssa.Analyzer].(*buildssa.SSA)
-	p := analyze(ssainput.Pkg)
+	p := Analyze(ssainput)
 	return p, nil
 }
 
 // Analyzes an SSA program and build the partition information.
-func analyze(pkg *ssa.Package) *Partitions {
-	prog := pkg.Prog
+func Analyze(ssainput *buildssa.SSA) *Partitions {
+	prog := ssainput.Pkg.Prog
 	// Use the call graph to initialize the contexts.
 	// TODO: the call graph can be CHA, RTA, VTA, etc.
 	cg := static.CallGraph(prog)
 	vis := visitor{state: NewState(), callees: mapCallees(cg)}
 	vis.initContexts(cg)
-	vis.initGlobalReferences(pkg)
+	vis.initGlobalReferences(ssainput.Pkg)
 	// Analyze all the functions and methods in the package,
 	// not just those in ssainput.ScrFuncs.
 	fns := ssautil.AllFunctions(prog)
+	for _, fn := range ssainput.SrcFuncs {
+		fns[fn] = true
+	}
 	for fn := range fns {
 		vis.initFunction(fn)
 	}
@@ -227,6 +230,8 @@ func (vis *visitor) visitInstruction(instr ssa.Instruction) {
 		vis.visitSlice(i)
 	case *ssa.MakeInterface:
 		vis.visitMakeInterface(i)
+	case *ssa.MakeClosure:
+		vis.visitMakeClosure(i)
 	case *ssa.Select:
 		vis.visitSelect(i)
 
@@ -235,7 +240,6 @@ func (vis *visitor) visitInstruction(instr ssa.Instruction) {
 
 	case *ssa.Alloc,
 		*ssa.MakeChan,
-		*ssa.MakeClosure,
 		*ssa.MakeMap,
 		*ssa.MakeSlice:
 		// Local variables are created in the beginning of the pass by initReferences().
@@ -271,15 +275,40 @@ func (vis *visitor) visitFieldAddr(addr *ssa.FieldAddr) {
 }
 
 func (vis *visitor) visitField(field *ssa.Field) {
-	// "t1 = t0.x" is implemented by t0[x -> t1].
+	// "t1 = t0.x" is implemented by t0[x -> addr], addr --> t1,
+	// where addr is the reference to the field address.
 	obj := field.X
 	fd := makeField(obj.Type(), field.Field)
-	vis.processHeapAccess(field, obj, *fd)
+	state := vis.state
+	for _, c := range vis.getContexts(field) {
+		objv := MakeReference(c, obj)
+		fmap := state.PartitionFieldMap(state.representative(objv))
+		fref := MakeReference(c, field)
+		// If obj's partition already has a mapping for this field, unify fref
+		// with the suitable partition. Otherwise set fref as the ref.
+		if v, ok := fmap[*fd]; ok {
+			// Implement "fmap[fd -> addr], addr --> fref".
+			state.Unify(vis.getPointee(v), fref)
+		} else {
+			fmap[*fd] = fref // over-approximation
+		}
+	}
 }
 
 func (vis *visitor) visitIndexAddr(addr *ssa.IndexAddr) {
 	// "t2 = &t1[t0]" is handled through address unification.
-	vis.processIndex(addr, addr.X, addr.Index)
+	obj := addr.X
+	if c, ok := addr.Index.(*ssa.Const); ok {
+		field := Field{Name: c.Value.String()}
+		for _, c := range vis.getContexts(addr) {
+			vis.unifyFieldAddress(c, obj, &field, addr)
+		}
+	}
+	// Also update the AnyField information.
+	// Note that base[c -> data] implies base[AnyField -> data]
+	for _, c := range vis.getContexts(addr) {
+		vis.unifyFieldAddress(c, obj, &anyIndexField, addr)
+	}
 }
 
 func (vis *visitor) visitIndex(index *ssa.Index) {
@@ -347,10 +376,22 @@ func (vis *visitor) visitUnOp(op *ssa.UnOp) {
 func (vis *visitor) visitBinOp(op *ssa.BinOp) {
 	// Numeric binop will be skipped.
 	// The only binop that matters now is "+" that concatenates strings.
-	if op.Op == token.ADD {
-		if tp, ok := op.X.Type().(*types.Basic); ok && tp.Kind() == types.String {
-			vis.unifyLocals(op, op.X)
-			vis.unifyLocals(op, op.Y)
+	if op.Op != token.ADD {
+		return
+	}
+	// for r0 = r1 + r2, unifying r0, r1 and r2 may cause too many
+	// over-approximations. Here fields are introduced to make:
+	//     {r0}[0->r1, 1->r2]
+	// which avoids unifying r1 and r2.
+	if tp, ok := op.X.Type().(*types.Basic); ok && tp.Kind() == types.String {
+		// unify instance field
+		for _, c := range vis.getContexts(op) {
+			if mayShareObject(op.X) {
+				vis.unifyField(c, op, Field{Name: "0"}, op.X)
+			}
+			if mayShareObject(op.Y) {
+				vis.unifyField(c, op, Field{Name: "1"}, op.Y)
+			}
 		}
 	}
 }
@@ -387,6 +428,13 @@ func (vis *visitor) visitExtract(extract *ssa.Extract) {
 		if base.CommaOk && extract.Index == 0 {
 			vis.processIndex(extract, base.X, base.Index)
 		}
+	case *ssa.UnOp:
+		if base.Op == token.ARROW && base.CommaOk && extract.Index == 0 { // channel read
+			// Use array read to simulate channel read.
+			vis.processHeapAccess(extract, base.X, anyIndexField)
+		}
+	case *ssa.Select:
+		// TODO: to be implemented
 	default: // Other cases: model "r1 = extract r0 #n" as "r0["n"] = r1".
 		field := Field{Name: strconv.Itoa(extract.Index)}
 		vis.processHeapAccess(extract, base, field)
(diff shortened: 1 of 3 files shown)


internal/pkg/earpointer/analysis.go
```diff
@@ -506,12 +570,13 @@ func (vis *visitor) unifyCallWithContexts(arg ssa.Value, param ssa.Value, callsi
 
 // Handle calls to builtin functions: https://golang.org/pkg/builtin/.
 func (vis *visitor) visitBuiltin(builtin *ssa.Builtin, instr ssa.Instruction) {
+	// TODO: support more library functions (https://github.com/google/go-flow-levee/issues/312)
```
mlevesquedion, line 573: Consider using `TODO(#312)` instead of providing the full link.

internal/pkg/earpointer/analysis.go
```diff
@@ -532,6 +597,41 @@ func (vis *visitor) visitBuiltin(builtin *ssa.Builtin, instr ssa.Instruction) {
 	}
 }
 
+// Handle some known functions, e.g. in package "fmt".
+func (vis *visitor) visitKnownFunction(fn *ssa.Function, instr ssa.Instruction) bool {
```
mlevesquedion, line 601: IMO this should also have `TODO(#312)`.

internal/pkg/earpointer/analysis.go
```diff
@@ -532,6 +597,41 @@ func (vis *visitor) visitBuiltin(builtin *ssa.Builtin, instr ssa.Instruction) {
...
+		fd := Field{Name: strconv.Itoa(index)}
+		if mayShareObject(op) {
+			for _, c := range vis.getContexts(dst) {
+				vis.unifyField(c, dst, fd, op)
+			}
+		}
+	}
+	fname := fn.Name()
+	if strings.HasPrefix(fname, "Sprint") || strings.HasPrefix(fname, "Errorf") {
+		// Propagage the arguments to the return value using fields.
```
mlevesquedion, line 613:
```suggestion
		// Propagate the arguments to the return value using fields.
```

internal/pkg/earpointer/analysis.go
```diff
@@ -532,6 +597,41 @@ func (vis *visitor) visitBuiltin(builtin *ssa.Builtin, instr ssa.Instruction) {
...
+		if dst, ok := instr.(ssa.Value); ok {
+			ops := instr.Operands(nil)
+			// ops[0] is the "append" function itself.
```
mlevesquedion, line 616:
```suggestion
			// ops[0] is the function itself.
```

internal/pkg/earpointer/analysis.go
```diff
@@ -532,6 +597,41 @@ func (vis *visitor) visitBuiltin(builtin *ssa.Builtin, instr ssa.Instruction) {
...
+			for i := 1; i < len(ops); i++ {
+				addField(dst, *ops[i], i)
+			}
+		}
+		return true
+	} else if strings.HasPrefix(fname, "Fprint") {
+		// Propagage the arguments to the first argument using fields.
+		ops := instr.Operands(nil)
+		first := *ops[1]
+		// ops[0] is the "append" function itself.
```
mlevesquedion, line 626:
```suggestion
		// ops[0] is the function itself.
```

internal/pkg/earpointer/analysis.go
```diff
@@ -532,6 +597,41 @@ func (vis *visitor) visitBuiltin(builtin *ssa.Builtin, instr ssa.Instruction) {
...
+		if dst, ok := instr.(ssa.Value); ok {
+			ops := instr.Operands(nil)
+			// ops[0] is the "append" function itself.
+			for i := 1; i < len(ops); i++ {
+				addField(dst, *ops[i], i)
+			}
+		}
+		return true
+	} else if strings.HasPrefix(fname, "Fprint") {
+		// Propagage the arguments to the first argument using fields.
```
mlevesquedion, line 623:
```suggestion
		// Propagate the arguments to the first argument using fields.
```

internal/pkg/earpointer/analysis.go
```diff
@@ -532,6 +597,41 @@ func (vis *visitor) visitBuiltin(builtin *ssa.Builtin, instr ssa.Instruction) {
...
+	addField := func(dst ssa.Value, op ssa.Value, index int) {
+		fd := Field{Name: strconv.Itoa(index)}
+		if mayShareObject(op) {
+			for _, c := range vis.getContexts(dst) {
+				vis.unifyField(c, dst, fd, op)
+			}
+		}
+	}
+	fname := fn.Name()
+	if strings.HasPrefix(fname, "Sprint") || strings.HasPrefix(fname, "Errorf") {
```
mlevesquedion, line 612: Consider using a `switch` with these prefix checks in the `case` clauses.

(Maybe you could put `return false` in the `default` case and `return true` after the `switch` to avoid duplication. Some may find this control flow unintuitive though, so maybe not.)

internal/pkg/earpointer/analysis.go
```diff
@@ -532,6 +597,41 @@ func (vis *visitor) visitBuiltin(builtin *ssa.Builtin, instr ssa.Instruction) {
...
+			// ops[0] is the "append" function itself.
+			for i := 1; i < len(ops); i++ {
+				addField(dst, *ops[i], i)
+			}
+		}
+		return true
+	} else if strings.HasPrefix(fname, "Fprint") {
+		// Propagage the arguments to the first argument using fields.
+		ops := instr.Operands(nil)
+		first := *ops[1]
```
mlevesquedion, line 625: Consider swapping with next line for clarity.

internal/pkg/earpointer/analysis.go
```diff
@@ -592,10 +692,12 @@ func (vis *visitor) processIndex(data ssa.Value, base ssa.Value, index ssa.Value
 	var field Field
 	if c, ok := index.(*ssa.Const); ok {
 		field = Field{Name: c.Value.String()}
+		vis.processHeapAccess(data, base, field)
```
mlevesquedion, line 695: Same comment as for `IndexAddr`: why not do just the "any" field?

internal/pkg/earpointer/analysis.go
```diff
@@ -626,7 +728,8 @@ func (vis *visitor) processAddressToValue(addr ssa.Value, value ssa.Value) {
 			if isUnifyByReference(value.Type()) { // unify-by-reference
 				state.Unify(v, cvalue)
 			} else {
-				vis.unifyByValue(v, cvalue)
+				// TODO: handle the unify-by-value semantics
```
mlevesquedion, line 731: Consider creating an issue to track this work.

internal/pkg/earpointer/state.go
```diff
@@ -169,7 +171,10 @@ func (state *state) unifyReps(ref1 Reference, ref2 Reference) {
 	pinfo1 := state.partitionInfo(prep1)
 	pinfo2 := state.partitionInfo(prep2)
 	// swap so that "ref1" is always the smaller partition
-	if pinfo1.numMembers > pinfo2.numMembers {
+	// if their sizes are the same, use source position information to
+	// make the choice deterministic.
```
mlevesquedion, line 175: Nice!

internal/pkg/earpointer/analysis_test.go
```diff
@@ -1027,16 +999,56 @@ func TestCallBuiltin(t *testing.T) {
...
 	}
 }
 
+func TestKnownCall(t *testing.T) {
+	code := `package p
+	import (
+		"fmt"
+		"os"
+	)
+	func f(s []*int, x string) {
```
mlevesquedion, line 1019:
```suggestion
	func f(x string) {
```

internal/pkg/earpointer/analysis_test.go
```diff
@@ -1027,16 +999,56 @@ func TestCallBuiltin(t *testing.T) {
...
+		t.Fatal(err)
+	}
+	// f.x is pointed by a field of f.t0,
+	// and by a field of f.t1 (i.e. os.Stderr).
+	want := concat(map[string]string{
+		"{Stderr}":    "--> f.t1",
+		"{f.t0}":      "[1->f.x]",
+		"{f.t1,f.t2}": "[2->f.x]",
+		"{f.t3}":      "[]",
+		"{f.s}":       "[]",
```
mlevesquedion, line 1044:
```suggestion
		"{f.s}":       "[]",
```

internal/pkg/earpointer/analysis_test.go
mlevesquedion: (`s` is not used)

internal/pkg/earpointer/analysis_test.go
mlevesquedion: (to go with the above suggested change)

internal/pkg/earpointer/analysis_test.go
```diff
@@ -1068,15 +1080,23 @@ func TestVariadicCall(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
-	want := concat(map[string]string{
-		"{f.a,f.t4,g.t1}":  "[]",
-		"{f.b}":            "[]",
-		"{f.t0,f.t3,g.ks}": "[0->g.t0, 1->f.t2]",
-		"{f.t1,g.t0}":      "--> g.t1",
-		"{f.t2}":           "--> f.b",
+	// Handle the non-determinism when choosing the representative during unifying.
```
mlevesquedion, line 1083: Consider highlighting the part of the `want` that can change due to non-determinism with e.g. a comment.

internal/pkg/earpointer/state.go
```diff
@@ -308,10 +313,27 @@ func (state *state) ToPartitions() *Partitions {
...
 func (p *Partitions) finalize() {
-	// Construct the map from field references to their parents.
-	p.constructFieldParentMap()
-
-	// Construct the members. members maps partitions representatives to the
+	// (1) Construct a new parent map that maps a reference to its representative
+	//     so as to accelerate representative lookup.
+	// For example, supposed the old parent map contains r1->r2, r2->r3,
+	// r3->r4(representative), then the new parent map contains r1->r4.
+	// "r4" can be returned immediately without chasing the union-find relation
```
mlevesquedion, line 320: Won't chasing only occur on lookup? And then because of compression it won't happen again?

mlevesquedion (commented):
LGTM, just some minor comments and questions.

Test outputs make more sense to me now.

internal/pkg/earpointer/analysis.go
```diff
@@ -64,22 +64,25 @@ type visitor struct {
...
 	// TODO: the call graph can be CHA, RTA, VTA, etc.
 	cg := static.CallGraph(prog)
 	vis := visitor{state: NewState(), callees: mapCallees(cg)}
 	vis.initContexts(cg)
-	vis.initGlobalReferences(pkg)
+	vis.initGlobalReferences(ssainput.Pkg)
 	// Analyze all the functions and methods in the package,
 	// not just those in ssainput.ScrFuncs.
 	fns := ssautil.AllFunctions(prog)
+	for _, fn := range ssainput.SrcFuncs {
```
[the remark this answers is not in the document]
guodongli-google: Without this the following unit test will fail. Yes it is strange that "ssautil.AllFunctions(prog)" doesn't include all the functions in some rare cases. 
levee_analysistest/example/tests/receivers/tests.go
Anyway it doesn't hurt to add those in ssainput.SrcFuncs.

internal/pkg/earpointer/analysis.go
```diff
@@ -271,15 +275,40 @@ func (vis *visitor) visitFieldAddr(addr *ssa.FieldAddr) {
...
+		if v, ok := fmap[*fd]; ok {
+			// Implement "fmap[fd -> addr], addr --> fref".
+			state.Unify(vis.getPointee(v), fref)
+		} else {
+			fmap[*fd] = fref // over-approximation
+		}
+	}
 }
 
 func (vis *visitor) visitIndexAddr(addr *ssa.IndexAddr) {
```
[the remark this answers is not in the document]
guodongli-google: Adding the "constant" field is to provide some extra information about the field, e.g. 

m["a"] = r0
m["b"] = r1
r2 = m["a"]
r3 = m[x]

we know that r2 can only be r0, but r3 can be {r0, r1}. Of course, the current revision actually over-approximates r2 to be {r0, r1} as well since  m["a"] = r0 results in m[anyIndex -> {r0}]. However, it is good to keep the constant field information as we can refine this in the future. For example, in a must-may analysis, we know that r2 may be {r0, r1}, but r2 must be {r0}.

internal/pkg/earpointer/analysis.go
```diff
@@ -387,6 +428,13 @@ func (vis *visitor) visitExtract(extract *ssa.Extract) {
 		if base.CommaOk && extract.Index == 0 {
 			vis.processIndex(extract, base.X, base.Index)
 		}
+	case *ssa.UnOp:
+		if base.Op == token.ARROW && base.CommaOk && extract.Index == 0 { // channel read
```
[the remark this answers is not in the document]
guodongli-google: The non-CommaOk is processed within "visitUnOp()".

internal/pkg/earpointer/analysis.go
guodongli-google: See above discussions.

internal/pkg/earpointer/state.go
guodongli-google: Yes chasing happens when an EAR state is constructed:

func (state *state) representative(ref Reference) Reference {
	failureCallback := func(ref Reference) Reference {
		// Some global variables has no declarations in the SSA;
		// Insert their references dynamically.
		state.Insert(ref)
		return ref
	}
	return state.lookupPartitionRep(ref, failureCallback)
}

After the entire state has been constructed, we can accelerate the lookup and avoid chasing:

func (p *Partitions) Representative(ref Reference) Reference {
	return p.parents[ref]
}

Since "Representative()" is a popular operation, it is worth to optimizing it here.

internal/pkg/earpointer/analysis.go
```diff
@@ -387,6 +428,33 @@ func (vis *visitor) visitExtract(extract *ssa.Extract) {
...
+		// t1 = <-t0,ok
+		// t2 = extract t1 0
+		if base.Op == token.ARROW && base.CommaOk && extract.Index == 0 { // channel read
+			// Use array read to simulate channel read.
+			vis.processHeapAccess(extract, base.X, anyIndexField)
+		}
+	case *ssa.Select:
+		// t1 = select blocking [c1<-t0, <-c2]  // (index int, recvOk bool, r_0 T_0, ... r_n-1 T_n-1)
+		// t2 = extract t1 #n
+		if extract.Index > 1 {
```
mlevesquedion, line 443: There is a fair bit of nesting going on here. Consider breaking if the condition isn't met.

internal/pkg/earpointer/analysis.go
```diff
@@ -387,6 +428,33 @@ func (vis *visitor) visitExtract(extract *ssa.Extract) {
...
+			k := -1 // for locating the (extract.Index-2)^{th} receive
+			sts := base.States
+			for i := 0; i < len(sts); i++ {
+				if sts[i].Dir == types.RecvOnly {
+					if k++; k == extract.Index-2 {
+						// Propagate the (extract.Index-2)^{th} receive value to the LHS.
+						if val := sts[i].Send; val != nil {
```
mlevesquedion, line 450: AFAIU since we checked that the `Dir` is `RecvOnly` above, `Send` should always be nil. I believe this should be setting the `Chan`'s `AnyField` to point to the received value?

internal/pkg/earpointer/analysis.go
```diff
@@ -387,6 +428,33 @@ func (vis *visitor) visitExtract(extract *ssa.Extract) {
...
+			// Use array read to simulate channel read.
+			vis.processHeapAccess(extract, base.X, anyIndexField)
+		}
+	case *ssa.Select:
+		// t1 = select blocking [c1<-t0, <-c2]  // (index int, recvOk bool, r_0 T_0, ... r_n-1 T_n-1)
+		// t2 = extract t1 #n
+		if extract.Index > 1 {
+			k := -1 // for locating the (extract.Index-2)^{th} receive
+			sts := base.States
+			for i := 0; i < len(sts); i++ {
```
mlevesquedion, line 446: Consider using `range` since the index isn't needed.

internal/pkg/earpointer/analysis_test.go
```diff
@@ -697,6 +670,58 @@ func TestChannel(t *testing.T) {
...
+			t7 = make interface{} <- string ("blocking select m...":string) interface{}
+			panic t7
+	*/
+	state, err := runCodeK0(code)
+	if err != nil {
+		t.Fatal(err)
+	}
+	want := concat(map[string]string{
+		"{f.c1}":      "[AnyField->f.m2]",
+		"{f.c2}":      "[]",
```
mlevesquedion, line 715: I think this should be `"{f.c2}": "[AnyField->f.t4]"`

mlevesquedion (approved):
Few more comments on the select implementation for extracts, otherwise LGTM.

PurelyApplied (approved):
I believe all my previous concerns were split out into PR#310, which I will re-review Soon™️.

PR #316, opened 2021-06-17
Ear taint checker

guodongli-google:


guodongli-google:
This actually supersedes: https://github.com/google/go-flow-levee/pull/310, so #310 can be ignored.

internal/pkg/earpointer/analysis.go
```diff
@@ -592,6 +596,7 @@ func (vis *visitor) unifyCallWithContexts(arg ssa.Value, param ssa.Value, callsi
 
 // Handle calls to builtin functions: https://golang.org/pkg/builtin/.
 func (vis *visitor) visitBuiltin(builtin *ssa.Builtin, instr ssa.Instruction) {
+	// TODO: support more library functions (https://github.com/google/go-flow-levee/issues/312)
```
mlevesquedion, line 599:
```suggestion
	// TODO(#312): support more library functions
```

Suggesting, as this is our convention for referencing issues from TODOs.

internal/pkg/levee/levee.go
```diff
@@ -48,6 +51,15 @@ func run(pass *analysis.Pass) (interface{}, error) {
...
 	}
+	if conf.UseEAR() {
+		return runEAR(pass) // Use the EAR-pointer based taint analysis
+	} else {
+		return runPropagation(pass) // Use the propagation based taint analysis
+	}
+}
+
+func runPropagation(pass *analysis.Pass) (interface{}, error) {
+	conf, _ := config.ReadConfig()
```
mlevesquedion, line 62: Prefer passing the config as an argument, instead of reading it again and ignoring the error.

internal/pkg/levee/levee.go
```diff
@@ -78,6 +90,72 @@ func run(pass *analysis.Pass) (interface{}, error) {
 	return nil, nil
 }
 
+// Use the EAR pointer analysis as the propagation engine
+func runEAR(pass *analysis.Pass) (interface{}, error) {
+	conf, _ := config.ReadConfig()
```
mlevesquedion, line 95: Prefer passing the config as an argument, instead of reading it again and ignoring the error.

internal/pkg/levee/levee.go
```diff
@@ -78,6 +90,72 @@ func run(pass *analysis.Pass) (interface{}, error) {
...
+	if state == nil {
+		return nil, fmt.Errorf("no valid EAR partitions")
+	}
+	funcSources := pass.ResultOf[source.Analyzer].(source.ResultType)
+	taggedFields := pass.ResultOf[fieldtags.Analyzer].(fieldtags.ResultType)
+	suppressedNodes := pass.ResultOf[suppression.Analyzer].(suppression.ResultType)
+	// Return whether a field is tainted.
+	isTaintField := func(named *types.Named, index int) bool {
+		if tt, ok := named.Underlying().(*types.Struct); ok {
+			return conf.IsSourceField(utils.DecomposeField(named, index)) ||
```
mlevesquedion, line 106: This would read better on a single line.

internal/pkg/levee/levee.go
```diff
@@ -78,6 +90,72 @@ func run(pass *analysis.Pass) (interface{}, error) {
...
+		for _, s := range sources {
+			srcRefs[s] = earpointer.GetSrcRefs(s, isTaintField, state, callees)
+		}
+		// Traverse all the callee functions (not just the ones with sink sources)
+		for member := range callees {
+			for _, b := range member.Blocks {
+				for _, instr := range b.Instrs {
+					switch v := instr.(type) {
+					case *ssa.Call:
+						if callee := v.Call.StaticCallee(); callee != nil &&
```
mlevesquedion, line 128: Prefer keeping expressions on a single line.

internal/pkg/earpointer/taint.go
```diff
@@ -0,0 +1,191 @@
...
+	ht.getStructFieldRefs(rep, val.Type(), refs, isTaintField)
+	return refs
+}
+
+// Obtain the references associated with a taint source, with field sensitivity.
+// Composite types are examined recursively to identify the taint elements, e.g.
+// (1) when a map contains taint elements, these elements are examined to identify
+//     taint sources; and
+// (2) when a struct object contains taint fields, only these fields
+// 	   and all their subfields are included in taint sources, and the struct object
```
mlevesquedion, line 90:
```suggestion
// 	 and all their subfields are included in taint sources, and the struct object
```

mlevesquedion (approved):
LGTM.

mlevesquedion:
Requesting review from @timothy-king.

internal/pkg/earpointer/taint.go
```diff
@@ -0,0 +1,191 @@
...
+type heapTraversal struct {
+	partitions *Partitions
+	callees    map[*ssa.Function]bool // the functions containing the references of interest
+	visited    ReferenceSet           // the visited references during the traversal
+}
+
+// For a function, transitively get the functions called within this function.
+// Argument "depth" controls the depth of the call chain.
+// For example, return {g1,g2,g3} for "func f(){ g1(); g2() }, func g1(){ g3() }".
+func getCalleeFunctions(fn *ssa.Function, result map[*ssa.Function]bool, depth uint) {
```
timothy-king, line 34: What is the evidence we need a depth parameter?

internal/pkg/earpointer/taint.go
```diff
@@ -0,0 +1,191 @@
...
+						result[callee] = true
+						getCalleeFunctions(callee, result, depth-1)
+					}
+				}
+			}
+		}
+	}
+}
+
+func GetCalleeFunctions(fn *ssa.Function, depth uint) map[*ssa.Function]bool {
```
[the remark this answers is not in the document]
timothy-king: `TransitiveCallees` is misleading due to the depth bound. BoundedDepthCallees?

internal/pkg/earpointer/taint.go
```diff
@@ -0,0 +1,191 @@
...
+// Argument "depth" controls the depth of the call chain.
+// For example, return {g1,g2,g3} for "func f(){ g1(); g2() }, func g1(){ g3() }".
+func getCalleeFunctions(fn *ssa.Function, result map[*ssa.Function]bool, depth uint) {
+	if depth <= 0 {
+		return
+	}
+	for _, b := range fn.Blocks {
+		for _, instr := range b.Instrs {
+			if call, ok := instr.(*ssa.Call); ok {
+				// TODO: the call graph can be CHA, RTA, VTA, etc.
```
timothy-king, line 41: Addressing this will change the public signature. (I think this is fine but @mlevesquedion should make the call.)

internal/pkg/earpointer/taint.go
```diff
@@ -0,0 +1,191 @@
...
+				if callee := call.Call.StaticCallee(); callee != nil && len(callee.Blocks) > 0 {
```
timothy-king, line 42: I think you need a comment saying why len(callee.Blocks) > 0 matters. even better would be a function that named the property you are testing.

internal/pkg/earpointer/taint.go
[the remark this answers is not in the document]
timothy-king: Does `map[*ssa.Function]bool` or `[]*ssa.Function` make more sense as the output of this function?

internal/pkg/earpointer/taint.go
```diff
@@ -0,0 +1,191 @@
...
+
+import (
+	"go/types"
+
+	"github.com/google/go-flow-levee/internal/pkg/source"
+	"golang.org/x/tools/go/ssa"
+)
+
+// A data structure to assist the traversal of an EAR heap.
+type heapTraversal struct {
```
timothy-king, line 25: nit: Group the heapTraveral functions closer together. A good place to start is after GetSrcRefs where this starts to make sense.

internal/pkg/earpointer/taint.go
```diff
@@ -0,0 +1,191 @@
...
+package earpointer
+
+import (
+	"go/types"
+
+	"github.com/google/go-flow-levee/internal/pkg/source"
+	"golang.org/x/tools/go/ssa"
+)
+
+// A data structure to assist the traversal of an EAR heap.
```
timothy-king, line 24: What does this mean? To be blunt this is just re-stating the name of the type in a sentence.

internal/pkg/earpointer/taint.go
```diff
@@ -0,0 +1,191 @@
...
+}
+
+func GetCalleeFunctions(fn *ssa.Function, depth uint) map[*ssa.Function]bool {
+	result := make(map[*ssa.Function]bool)
+	result[fn] = true
+	getCalleeFunctions(fn, result, depth)
+	return result
+}
+
+func (ht *heapTraversal) isWithinFunctions(ref Reference) bool {
```
timothy-king, line 60: isWithinCallees?

internal/pkg/earpointer/taint.go
```diff
@@ -0,0 +1,191 @@
...
+		_, ok := ht.callees[fn]
+		return ok
+	}
+	// Globals and Builtins have no parents.
+	return true
+}
+
+// Obtain the references associated with a taint source, with field sensitivity.
+// Consider only the references located with the "callees".
+func GetSrcRefs(src *source.Source, isTaintField func(named *types.Named, index int) bool,
```
timothy-king, line 71: "SourceReferences"?

internal/pkg/earpointer/taint.go
```diff
@@ -0,0 +1,191 @@
...
+}
+
+// Obtain the references associated with a taint source, with field sensitivity.
+// Composite types are examined recursively to identify the taint elements, e.g.
+// (1) when a map contains taint elements, these elements are examined to identify
+//     taint sources; and
+// (2) when a struct object contains taint fields, only these fields
+// 	   and all their subfields are included in taint sources, and the struct object
+//     itself is a taint source. Other fields will not be tainted.
+func (ht *heapTraversal) getStructFieldRefs(rep Reference, tp types.Type, result ReferenceSet, isTaintField func(named *types.Named, index int) bool) {
```
timothy-king, line 92: Push "fixed for a given instance" fields into heapTraversal. This is [likely] mildly more efficient. More importantly it is less code and less complicated signatures. `result` and `isTaintField` look like they fall into this.

internal/pkg/levee/levee.go
```diff
@@ -78,6 +90,72 @@ func run(pass *analysis.Pass) (interface{}, error) {
...
+							conf.IsSink(utils.DecomposeFunction(callee)) {
+							sink := instr
+							for _, src := range sources {
+								if earpointer.IsEARTainted(sink, srcRefs[src], state, callees) &&
```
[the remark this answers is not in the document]
timothy-king: Also this is likely in the realm for introducing an auxiliary function.

internal/pkg/earpointer/taint.go
```diff
@@ -0,0 +1,191 @@
...
+	rep := p.Representative(ref)
+	for _, r := range p.PartitionFieldMap(rep) {
+		if _, ok := ht.visited[r]; !ok {
+			ht.getFieldReferences(r, result)
+		}
+	}
+}
+
+// Checks whether a taint sink can be reached by a taint source by examining the alias information.
+func IsEARTainted(sink ssa.Instruction, srcRefs ReferenceSet, state *Partitions, callees map[*ssa.Function]bool) bool {
```
timothy-king, line 168: Is there an expected or required relationship between any of these arguments?

internal/pkg/earpointer/taint.go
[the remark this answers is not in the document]
mlevesquedion: IMO the `map` makes more sense, since AFAICT we're only interested in membership tests.

internal/pkg/earpointer/taint.go
mlevesquedion: This looks fine to me also.

Consider creating an issue an referencing it in the TODO, though. (I think there are other similar TODOs lying around, if you want to create the issue please also update those comments).

internal/pkg/earpointer/taint.go
guodongli-google: Now `isTaintField` has been moved into `heapTraversal`.
`result` is kept as it is the output of this function.

internal/pkg/earpointer/taint.go
guodongli-google: These arguments are the required ones used by this public API. We may refactor it to be a public method of heapTraversal:

`func (ht *heapTraversal) IsEARTainted(sink ssa.Instruction, srcRefs ReferenceSet) { ... }`

This will require exposing the `heapTraversal`. Either way works for me.

internal/pkg/earpointer/taint.go
guodongli-google: As mentioned before, this is to control how far we go when searching the call chain. In the future CLs it will be used to control the searching of other cases. For example, one interesting case is to chase the chain of callers: f2->g2->h2.

```
func f2() string {
   a := new T{} //  taint source
   return fmt.Printf(a)
}

func g2() {
   return f2(a)
}

func h2() { 
   x := g2()
  sinkf(x) // taint sink
}
```

Another interesting case is:
```
func f3() string {
   a := new T{} //  taint source
   return fmt.Printf(a)
}

func g3(x string) {
     sinkf(x) // taint sink
}

func h3() { 
   x := f3()
   g3(x)
}
```

So connecting a source and a sink may span many functions. This `depth` is used to limit how many functions we should consider when connecting a source and a sink. Ideally we should not use such a limit, but since EAR uses aggressive merging, the FP rate may be high without applying this limit. Too many references may be involved as well if we don't bound it. If later on we have experimental data justifying that we don't need such a limit, then we can remove it.
I recall that HawkEye uses a similar limit.

mlevesquedion (approved):
LGTM.

I notice that the tests that were present in #310 are not in this PR. Are you planning to add those in a follow-up PR?

internal/pkg/earpointer/taint.go
timothy-king: The question was focusing on something slightly different.  We have 3 inter related function cals:
```
callees := earpointer.GetCalleeFunctions(fn, conf.EARCallDepth())
earpointer.GetSrcRefs(s, isTaintField, state, callees)
earpointer.IsEARTainted(sink, srcRefs[src], state, callees)
```
These seem to all be related. Does this function support any srcRefs, state or callees or is there an expected relationship that these are setup exactly like this?

internal/pkg/earpointer/taint.go
timothy-king: So the evidence at the moment is 1) theoretical 2) following practices from similar projects. Is this a fair summary? These are not terrible reasons, but they are not overwhelming either.

> If later on we have experimental data justifying that we don't need such a limit, then we can remove it.

It is easy to go from no limit to adding one too.

internal/pkg/earpointer/taint.go
guodongli-google: Yes `GetSrcRefs()` and `IsEARTainted()` can be combined, while `GetCalleeFunctions()` is also used in `runEAR()`. I actually plan to rewrite the main logic of connecting the sources and sinks to take into account the two examples I gave above. Roughly, given a set of source locations, and a set of sink locations, return the reachable source-sink pairs, possibly limited by the distance between the source and sink. This way we need only one API.
```
  func Reach(sources []Source, sinks []ssa.Instruction) map[Source]map[ssa.Instruction]
```
But this will be done in a future PR, which needs some new unit tests.
Regarding the current limit, we can simply set it to a large number to actually void its effect.

internal/pkg/earpointer/taint.go
guodongli-google: I believed that I've done some tests by disabling the limit or setting a large limit, there were more FPs showing up. 
Anyway, this can be gotten around by using a large limit so I don't think that it is a fundamental issue.

guodongli-google:
> LGTM.
> 
> I notice that the tests that were present in #310 are not in this PR. Are you planning to add those in a follow-up PR?

Forgot to add it last time. Now it is there.

internal/pkg/levee/levee_test.go
```diff
@@ -29,6 +29,9 @@ func TestLevee(t *testing.T) {
 	if err := Analyzer.Flags.Set("config", dataDir+"/test-config.yaml"); err != nil {
 		t.Error(err)
 	}
+	if err := Analyzer.Flags.Set("useEAR", "false"); err != nil {
```
mlevesquedion, line 32: The flag is `false` by default, so this may not be necessary? Same comment for other tests in this file.

internal/pkg/earpointer/taint.go
timothy-king: I am trying to ask whether these *must* be combined for the function to succeed.

Can I have sequences where these values are unrelated?
```
srcRefs := earpointer.GetSrcRefs(s, isTaintField, state0, callees0)
earpointer.IsEARTainted(sink, srcRefs, state0, callees1)
```
Is this meaningful? Will it crash? Have we baked in a precondition that state0 and state1 must be the same? callees0 and callees1?

internal/pkg/levee/levee_test.go
guodongli-google: The previous commit fails the tests in "levee_test.go". When "levee_ear_test.go" is run first, this flag is set to "true" there, and when the subsequent running of "levee_test.go" will fail. This is because "Analyzer.Flags" is a global.

mlevesquedion: Ah, I was wondering if there might be some issue involving test ordering. Thanks, resolved.

internal/pkg/earpointer/taint.go
guodongli-google: It won't crash if we use different `state`s or `callees`s, but the results may be not we expect.
The current expectation is: `state`s should be the same since the same heap is traversed for the sources and sinks. In contrast, `callees0` and `callees1` can be different: `callees0` controls the set of sources we are interested in, and `callees1` controls the set of sinks. In fact, using different `callees0` and `callees1` give different heuristics. Basically, `callees0` and `callees1` specify the regions in the heap where we search for the source or sink references.

timothy-king: It sounds like state needs additional documentation about what the expectation is then. It sounds like callees do not.

internal/pkg/levee/levee_ear_test.go
```diff
@@ -0,0 +1,69 @@
...
+	}
+	if *debugging {
+		Analyzer.Requires = append(Analyzer.Requires, debug.Analyzer)
+	}
+
+	analysistest.Run(t, dataDir, Analyzer, "./src/levee_analysistest/example/tests/arguments")
+	analysistest.Run(t, dataDir, Analyzer, "./src/levee_analysistest/example/tests/basictypes")
+	analysistest.Run(t, dataDir, Analyzer, "./src/levee_analysistest/example/tests/binop")
+	analysistest.Run(t, dataDir, Analyzer, "./src/levee_analysistest/example/tests/booleans")
+	// analysistest.Run(t, dataDir, Analyzer, "./src/levee_analysistest/example/tests/callorder")  // TODO: flow-sensitive
```
PurelyApplied, line 40: If these test differences are irreconcilable, we should partition them out.
 
Part of the appeal of the `...` format is that we were forgetting to add our test targets when we added new test packages.  We should refactor these tests soon so that the shared tests can be `example/tests/shared/...`, and approach-specific tests can be tested at `example/tests/propagation/...` and `example/tests/ear/...`.  That can happen in a future PR, though.

PurelyApplied (commented):
Broadly, LGTM.  My primary concerns are about globals and some nits about `map`s.

internal/pkg/levee/levee_test.go
timothy-king: For testing, an "easier" answer is to just have different analysis.Analyzer objects for testing. Sure there is a bit of a disconnect from the "production" object, but this should not be a big deal if there is enough separation between the modes.

internal/pkg/levee/levee_ear_test.go
guodongli-google: Good point. I added a TODO #318 for this.

internal/pkg/levee/levee.go
```diff
@@ -78,6 +90,77 @@ func run(pass *analysis.Pass) (interface{}, error) {
 	return nil, nil
 }
 
+// Use the EAR pointer analysis as the propagation engine
+func runEAR(pass *analysis.Pass, conf *config.Config) (interface{}, error) {
```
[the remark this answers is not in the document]
guodongli-google: Most of the EAR related code has been moved to `earpointer/taint.go`. 
This part will be revised significantly in the future CLs, e.g. to implement a better way to connect the sources and the sinks.

internal/pkg/levee/levee_test.go
guodongli-google: Now the flags have been moved to `config.yaml`, and specific yaml files will be used for EAR. There are no globals for these flags anymore.

internal/pkg/earpointer/taint.go
```diff
@@ -189,10 +189,73 @@ func IsEARTainted(sink ssa.Instruction, srcRefs ReferenceSet, heap *Partitions,
...
+		// This set is used to narrow down the set of references needed to be
+		// considered during EAR heap traversal. It can also help reducing the
+		// false positives and boosting the performance.
+		callees := boundedDepthCallees(fn, conf.EARTaintCallSpan)
+		srcRefs := make(map[*source.Source]ReferenceSet)
+		for _, s := range sources {
+			srcRefs[s] = srcAliasRefs(s, isTaintField, heap, callees)
+		}
+		// Return a source if it can reach the given sink.
+		reachAnySource := func(sink ssa.Instruction) *source.Source {
```
PurelyApplied, line 222: If this doesn't _need_ to be dynamically linked, I'd rather it were extracted to a top-level `func` definition with its inputs can be made explicit.

PurelyApplied (approved):

PurelyApplied (approved):

PR #319, opened 2021-07-09
EAR taint checker: traverse callers when connecting the sources and s…

guodongli-google:
…inks

### DIFF
--- a/internal/pkg/earpointer/taint.go
+++ b/internal/pkg/earpointer/taint.go
@@ -241,9 +241,6 @@ func SourcesToSinks(funcSources source.ResultType, isTaintField func(named *type
 				for _, instr := range b.Instrs {
 					switch v := instr.(type) {
 					case *ssa.Call:
-						if _, ok := traces[instr]; ok {
-							continue
-						}
 						callees := calleeMap[&v.Call]
 						if callees == nil { // the static call graph may ignore some cases
 							if sc := v.Call.StaticCallee(); sc != nil {
@@ -254,8 +251,11 @@ func SourcesToSinks(funcSources source.ResultType, isTaintField func(named *type
 						for _, callee := range callees {
 							if conf.IsSink(utils.DecomposeFunction(callee)) {
 								if src := ht.canReach(sink, sources, srcRefs); src != nil {
-									traces[sink] = &SourceSinkTrace{Src: src, Sink: sink}
-									break
+									// If a previous source has been found, be in favor of the source within the same
+									// function. This can be extended to be in favor of the source closest to the sink.
+									if _, ok := traces[instr]; !ok || src.Node.Parent() == sink.Parent() {
+										traces[sink] = &SourceSinkTrace{Src: src, Sink: sink}
+									}
 								}
 							}
 						}
--- a/internal/pkg/earpointer/taint.go
+++ b/internal/pkg/earpointer/taint.go
@@ -16,6 +16,7 @@ package earpointer
 
 import (
 	"go/types"
+
 	"golang.org/x/tools/go/callgraph"
 
 	"github.com/google/go-flow-levee/internal/pkg/config"
--- a/internal/pkg/earpointer/taint.go
+++ b/internal/pkg/earpointer/taint.go
@@ -28,15 +28,17 @@ import (
 
 // Bounded traversal of an EAR heap.
 type heapTraversal struct {
-	heap         *Partitions
-	callees      map[*ssa.Function]bool // the functions containing the references of interest
-	visited      ReferenceSet           // the visited references during the traversal
+	heap *Partitions
+	// The reachable functions containing the references of interest.
+	reachableFns map[*ssa.Function]bool
+	// The visited references during the traversal.
+	visited      ReferenceSet
 	isTaintField func(named *types.Named, index int) bool
 }
 
 func (ht *heapTraversal) isWithinCallees(ref Reference) bool {
 	if fn := ref.Value().Parent(); fn != nil {
-		return ht.callees[fn]
+		return ht.reachableFns[fn]
 	}
 	// Globals and Builtins have no parents.
 	return true
@@ -61,7 +63,8 @@ func (ht *heapTraversal) srcRefs(rep Reference, tp types.Type, result ReferenceS
 			ht.srcRefs(rep, tp.Underlying(), result)
 			return
 		}
-		result[rep] = true // the current struct object is tainted
+		// Mark the current struct object as tainted.
+		result[rep] = true
 		// Look for the taint fields.
 		for i := 0; i < tt.NumFields(); i++ {
 			f := tt.Field(i)
@@ -134,7 +137,7 @@ func (ht *heapTraversal) canReach(sink ssa.Instruction, sources []*source.Source
 	sinkedRefs := make(map[Reference]bool)
 	for _, op := range sink.Operands(nil) {
 		// Use a separate heapTraversal to search for the sink references.
-		sinkHT := &heapTraversal{heap: ht.heap, callees: ht.callees, visited: make(ReferenceSet)}
+		sinkHT := &heapTraversal{heap: ht.heap, reachableFns: ht.reachableFns, visited: make(ReferenceSet)}
 		v := *op
 		if isLocal(v) || isGlobal(v) {
 			ref := MakeLocalWithEmptyContext(v)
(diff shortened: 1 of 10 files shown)


guodongli-google:
Add @timothy-king as a reviewer.

internal/pkg/earpointer/taint.go
```diff
@@ -153,31 +155,34 @@ func (ht *heapTraversal) canReach(sink ssa.Instruction, sources []*source.Source
...
-// For example, return {g1,g2,g3} for "func f(){ g1(); g2() }, func g1(){ g3() }".
-func calleeFunctions(fn *ssa.Function, result map[*ssa.Function]bool, depth uint) {
-	if depth <= 0 {
+// For a function, transitively get the functions reachable from this function
+// according to the call graph. Both callers and callees are considered.
+// Argument "depth" controls the depth of the call chain. For example,
+//   func f(){ g1(); g2() }
+//   func g1(){ g3() }
+// for input "g1", return {f,g1,g2,g3} if depth>1, and return {f, g1} if depth=1.
+func reachableFunctions(fn *ssa.Function, result map[*ssa.Function]bool, cg *callgraph.Graph, depth uint) {
```
timothy-king, line 164: result is an output parameter. This probably should be the last or first argument and documented.

internal/pkg/earpointer/taint.go
```diff
@@ -153,31 +155,34 @@ func (ht *heapTraversal) canReach(sink ssa.Instruction, sources []*source.Source
...
-// Argument "depth" controls the depth of the call chain.
-// For example, return {g1,g2,g3} for "func f(){ g1(); g2() }, func g1(){ g3() }".
-func calleeFunctions(fn *ssa.Function, result map[*ssa.Function]bool, depth uint) {
-	if depth <= 0 {
+// For a function, transitively get the functions reachable from this function
+// according to the call graph. Both callers and callees are considered.
+// Argument "depth" controls the depth of the call chain. For example,
+//   func f(){ g1(); g2() }
+//   func g1(){ g3() }
+// for input "g1", return {f,g1,g2,g3} if depth>1, and return {f, g1} if depth=1.
```
timothy-king, line 163: "return" is bit misleading as it is an output parameter.

internal/pkg/levee/testdata/src/levee_analysistest/ear/tests/call/embedded-src.go
```diff
@@ -0,0 +1,39 @@
...
+func f1(s string) {
+	sinkf1(s)
+}
+
+func createData() string {
+	src := &core.Source{}
+	return src.Data
+}
+
+// Test the case where the source is introduced in an embedded function.
```
timothy-king, line 34: I don't know what embedded means in this context.

internal/pkg/earpointer/taint.go
```diff
@@ -206,33 +211,52 @@ type SourceSinkTrace struct {
...
+		// in search for connected sinks.
+		ht := &heapTraversal{heap: heap, callees: reachable, visited: make(ReferenceSet)}
+		for member := range reachable {
 			for _, b := range member.Blocks {
 				for _, instr := range b.Instrs {
 					switch v := instr.(type) {
 					case *ssa.Call:
+						callees := calleeMap[&v.Call]
+						if callees == nil { // the static call graph may ignore some cases
+							if sc := v.Call.StaticCallee(); sc != nil {
```
timothy-king, line 246: This seems unlikely to help. There is a static callee, but no edge in the static call graph with depth >= 1.

Can you point to a case it does help? If it does, this seems like a bug and probably deserves a todo.

internal/pkg/earpointer/taint.go
```diff
@@ -153,31 +155,34 @@ func (ht *heapTraversal) canReach(sink ssa.Instruction, sources []*source.Source
...
+	result[fn] = true
+	node := cg.Nodes[fn]
+	if node == nil {
+		return
+	}
+	// Visit the callees within "fn".
+	for _, out := range node.Out {
+		reachableFunctions(out.Callee.Func, result, cg, depth-1)
+	}
+	// Visit the callers of "fn".
```
mlevesquedion, line 177: What are the implications of also gathering the callers? Does this lead to more true/false positives?

internal/pkg/earpointer/taint.go
mlevesquedion, line 164: Consider keeping `result` in last position since it is an out parameter. Or returning it instead.

internal/pkg/earpointer/taint.go
```diff
@@ -206,33 +211,52 @@ type SourceSinkTrace struct {
 
 // Look for <source, sink> pairs by examining the heap alias information.
 func SourcesToSinks(funcSources source.ResultType, isTaintField func(named *types.Named, index int) bool,
-	heap *Partitions, conf *config.Config) []*SourceSinkTrace {
+	heap *Partitions, conf *config.Config) map[ssa.Instruction]*SourceSinkTrace {
 
-	var traces []*SourceSinkTrace
+	calleeMap := mapCallees(heap.cg) // map a callsite to the possible callees
```
mlevesquedion, line 216:
```suggestion
	calleeMap := mapCallees(heap.cg)  // map a callsite to the possible callees
```

internal/pkg/earpointer/taint.go
```diff
@@ -206,33 +211,52 @@ type SourceSinkTrace struct {
...
 		for _, s := range sources {
-			srcRefs[s] = srcAliasRefs(s, isTaintField, heap, callees)
+			srcRefs[s] = srcAliasRefs(s, isTaintField, heap, reachable)
 		}
-		// Traverse all the callee functions (not just the ones with sink sources)
-		ht := &heapTraversal{heap: heap, callees: callees, visited: make(ReferenceSet)}
-		for member := range callees {
+		// Traverse all the reachable functions (not just the ones with sink sources)
+		// in search for connected sinks.
+		ht := &heapTraversal{heap: heap, callees: reachable, visited: make(ReferenceSet)}
```
mlevesquedion, line 238: Consider renaming the field, as `callees` is no longer accurate.

internal/pkg/earpointer/taint.go
guodongli-google: It is to capture more inter-procedural taint propagation. The new added tests contain some example. 
For instance, in the following code,`createSource2()` return type `interface{}` rather than  `Source`, so Levee may not report a finding at `Sinkf(createSource2())` since it loses the source type information upon the exit of `createSource2()`.

func createSource2() interface{} {
	return &core.Source{}
}

Sinkf(createSource2())

In general, the goal is to connect the sources and sinks even there are no explicit embedding structure in the functions containing the sources and sinks.

I have seen some new findings in k8s by using this extension, but most of they are likely FPs, which can be partially reduced using context sensitiveness (in the next CL).

internal/pkg/earpointer/taint.go
guodongli-google: Without this one unit test fails:

../levee/testdata/src/levee_analysistest/example/tests/receivers/tests.go

```
type sourceBuilder struct {
  ...
}

func (b *sourceBuilder) buildP() {
  ...
}
```
Method `buildP` is not in the call graph. It is because it is not included in the method set used by the following code: 

https://github.com/golang/tools/blob/master/go/ssa/ssautil/visit.go#L44-L45

This seems to be a bug in SSA. A temporary work-around is to expose "sourceBuilder" by renaming it to "SourceBuilder" as done here.

timothy-king: Missing a function is concerning. If we are not under time pressure, I would like to understand the problem better before proceeding instead of hacking around it.

How does the unit test fail? Which package is being analyzed? Why does exposing the function fix this?

guodongli-google: An internal bug report with a reproducer has been created in Buganizer.

It seems that the ssa builder may ignore some unexposed methods:
 
https://github.com/golang/tools/blob/master/go/ssa/builder.go#L2278-L2279

internal/pkg/earpointer/taint.go
mlevesquedion: Sounds good, makes sense to me!

mlevesquedion (approved):

internal/pkg/earpointer/taint.go
timothy-king: ssautil was producing a list of functions that correctly concluded that sourceBuilder and these methods were unreachable. (There are other concerns for other cases, but this case was a correct conclusion.) I think we want this code to rely on the callgraph it is given, and the most recent version does this.

Changing to "SourceBuilder" is a good fix in this case. I think this has been sufficiently addressed.

timothy-king (approved):
LGTM

PR #325, opened 2022-05-06
Upgrade the Go version to v1.18, and x/tools to v0.1.11. Support generics better.

### LINKED ISSUE #323: Generics are not supported by analyzers
## Bug report

**Describe the bug**
Generics are not supported by analyzers or helper functions. Most notably the new `*types.TypeParam` type is not recognized.

**To Reproduce**

1. Upgrade x/tools to a recent version (which supports ssa + generics):

```
go get golang.org/x/tools@v0.1.11-0.20220504225841-45c8a7131235
go mod download github.com/yuin/goldmark 
go get golang.org/x/tools/internal/gocommand@v0.1.11-0.20220504225841-45c8a7131235
```

2. Add a generic type with a method to `internal/pkg/sourcetype/testdata/test_stackoverflow.go`

E.g.
```
type G[T any]struct{}

func (G[T]) M(_ T) {
}
```

3. Run the sourcetype test in verbose mode:

```
% go test -v ./internal/pkg/sourcetype
=== RUN   TestSourceTypeDoesNotStackOverflow
unexpected type received: *types.TypeParam T; please report this issue
--- PASS: TestSourceTypeDoesNotStackOverflow (0.02s)
PASS
ok  	github.com/google/go-flow-levee/internal/pkg/sourcetype	0.048s
```

Note the **`unexpected type received: *types.TypeParam T; please report this issue`** in the output.

**Additional context**
n/a

guodongli-google:
Address the differences related to the new SSA format.
Remove errors/warnings on generics types, and add unit tests.

Fixes #323


guodongli-google:
Have no idea why the linter "addlicense" fails since it succeeds locally.

.github/workflows/ci.yaml
```diff
@@ -29,7 +29,7 @@ jobs:
       - name: Check goimports
         if: ${{ always() }}
         run: |
-          go install golang.org/x/tools/cmd/goimports
+          go install golang.org/x/tools/cmd/goimports@latest
```
PurelyApplied, line 32: I'm always a little wary of `@latest`, since builds would not longer be strictly reproducible.  (Not that this _wasn't_ an issue with the bare `go get`s.)

Is there any significant downside to using the fixed sha referenced in the go.mod?

Ditto the other two.

go.mod
```diff
@@ -1,9 +1,19 @@
...
 
 require (
 	github.com/google/go-cmp v0.5.2
-	golang.org/x/tools v0.0.0-20200416214402-fc959738d646
+	golang.org/x/exp/typeparams v0.0.0-20220428152302-39d4317da171
+	golang.org/x/tools v0.1.11-0.20220504225841-45c8a7131235
 	sigs.k8s.io/yaml v1.2.0
 )
+
+// TODO: clean up indirect imports.
```
PurelyApplied, line 12: Is this actionable?  I thought `// indirect` imports were the result of a direct import not building with a `go.mod` that details explicit dependencies, so the consumer has to choose explicit versions.  But other than removing our direct dependencies, I don't think we get to clean up the indirect ones, do we?

I think having them grouped separately is cleaned up enough.

internal/pkg/sourcetype/sourcetype.go
```diff
@@ -47,6 +49,7 @@ func isSourceType(c *config.Config, tf fieldtags.ResultType, t types.Type, seen
 
 	switch tt := t.(type) {
 	case *types.Named:
+		// generics
```
PurelyApplied, line 52: Accidental comment?

PurelyApplied (commented):
I think we should signal to the user _somewhere_ that generic support is not enabled.  That could be in the README inclusive-or as a message when these skip blocks are hit.  But it seems like a gap in coverage that a user might be expecting otherwise.

.github/workflows/ci.yaml
guodongli-google: Agreed. It is safer to use a specific version rather than the lastest one.

go.mod
guodongli-google: These indirect imports are actually produced automatically by `go mod tidy`, and I don't have a better way to handle them now.

guodongli-google:
> I think we should signal to the user _somewhere_ that generic support is not enabled. That could be in the README inclusive-or as a message when these skip blocks are hit. But it seems like a gap in coverage that a user might be expecting otherwise.

Agreed. However warning each generics usage will be too much, and it is better to specify it somewhere else, e.g. in a README file.
There are follow-up CLs where I want to handle the generics better, e.g. through the "InstantiateGenerics" mode. Nevertheless, the full support won't be available soon. I'd let you recommend how to put in such a message.

PurelyApplied (commented):
I'm sorry to have let this get so stale.

I saw Tim's #327 and have started a thread there, but to summarize: it would be nice to fold this and that PR into each other.  I can go either way on either requiring 1.18+ vs Tim's `init()` approach to set the expectation between `interface{}` vs `any`, but like this PR's addition of tests (which I suppose require 1.18).

sjbarag:
Looks like most of this landed via #327. Are there still generics-support gaps?

PR #327, opened 2022-08-04
Update to dependency on golang.org/x/tools to v0.1.12.

### LINKED ISSUE #326: Crashes when analyzing Go 1.19 standard libraries
Go standard libraries starting in 1.19 now contain generics. [Example](https://cs.opensource.google/go/go/+/refs/tags/go1.19:src/sync/atomic/type.go;l=54). Levee's `go.mod` uses a version of x/tools/ssa that predates support for generics and crashes on this code. See https://github.com/kubernetes/kubernetes/issues/111452 for how this manifests.

timothy-king:
This prevents crashes on Go code containing generics. The standard library now contains generics at head and levee crashes for folks using newer toolchains.

See https://github.com/kubernetes/kubernetes/issues/111452.

Reflects running:
```
go get golang.org/x/tools@v0.1.12
go mod tidy
```

Fixes #326


internal/pkg/utils/utils_go118.go
```diff
@@ -0,0 +1,23 @@
...
+
+//go:build go1.18
+// +build go1.18
+
+package utils
+
+func init() {
+	// After 1.18 default rendering of interface{} is any.
+	DefaultEmptyInterface = "any"
+}
```
PurelyApplied, line 23: (a) This is a clever, clean way to do this sort of versioning that I haven't had the occasion to see before.  Neat!

(b) I can see it going either way, but would it be better to just keep it simple and bump the `go.mod` to require 1.18?  I don't see us maintaining distinct branches for backwards compatibility, and the old version is tagged if someone using the scanner is working at pre-1.18.

Then again, it's not _that_ much mental overhead here, so keeping it as is and more versatile could be good, unless / until there's any pressure to simplify it in the future.

PurelyApplied (approved):

internal/pkg/utils/utils_go118.go
PurelyApplied: Hmmm, related, I see that I've let #325 go stale without review, and that _does_ bump to go1.18.

@guodongli-google, if this PR is merged, do you see that throwing a huge wrench in #325?  Would that break without the go1.18 bump?

PurelyApplied (changes requested):
Also overlapping with #325, but a test that uses generics would be a good thing to include.

I kind of feel like this should be folded into #325, so that PR doesn't need all the `interface{}` to `any` edits but could keep everything else (with or without the base 1.18 requirement).

internal/pkg/utils/utils_go118.go
timothy-king: I'll try to summarize the delta between this and #325. This is doing the minimum to be backwards compatible, fixes tests, bumps only the versions it needs to (CI env testing requirements) and prevents crashes while not changing the solvers. #325 bumps the version to 1.18 for simplicity, and does minimal changes to the solvers to prevent flows through type parameters and suppresses complaining (false negatives). This does not suppress type parameter complaining. So there is a lot of overlap between these two.

Whether to be backwards compatible or keep it simple and version bump is a judgement call for you :-)

timothy-king:
> I kind of feel like this should be folded into https://github.com/google/go-flow-levee/pull/325, so that PR doesn't need all the interface{} to any edits but could keep everything else (with or without the base 1.18 requirement).

I am not sure which order you are suggesting. This then #325. Or #325 then this?

internal/pkg/utils/utils_go118.go
PurelyApplied: Gotcha.  This PR doesn't stop the `unexpected type received: *types.TypeParam T; please report this issue`, but it wasn't trying to.  That was my disconnect.

PurelyApplied (approved):
Your other comment provided some clarity I was missing.  We can figure out how to pick up the desired bits of #325 after this.

Thanks!

